PR #24, opened 2017-07-13
Instead of converting paths back and forth, just use them as-is. 

andreas23:
Fixes build on cygwin.


andreas23:
Mhh. I see now why relative paths are used everywhere: the OCaml is native, so there is a zillion places where the path names would need to be translated back and forth. Using relative paths everywhere neatly sidesteps this issue.

wintersteiger:
Exactly, there a loads of places that depend on that and make is not always called from the root directory. I'm happy to change that once all of this is merged into master, but until then I prefer to keep it as is. 

BTW, I can't reproduce your problem via my cygwin/python/etc. Do you have any symbolic links in your paths?

msprotz:
I think we should refrain from adding python as a dependency to the build system of HACL*. This is likely to make a lot of people at INRIA cringe. And it seems like this ought to be achievable without Python.

How about using cygpath -m and/or a combination of readlink -f? (Possibly with `READLINK=$(shell which greadlink >/dev/null 2>&1 && echo greadlink || echo readlink)`)? That's why we assume that everyone is running within a cygwin terminal: so that we have these tools at our disposal in the Makefiles and don't need to take yet another dependency.

wintersteiger:
@msprotz Sure, I'm happy with any tool that converts between absolute and relative path.

Originally I used the `realpath` command line utility, but the version that ships with Ubuntu 14.04 doesn't support the `--relative-to` option (thus breaking the Travis build, which can't be upgraded to 16.x easily). My main reasons for choosing Python instead are that it's a one-liner and that if/when the move to SCons happens, we would likely have to do exactly the same anyways. I've seen many references to `readlink -f`, but none that would actually convert to/from relative paths, but then I didn't invest a lot of time either.

wintersteiger:
Forgot to add: it probably wouldn't be too much of an effort to turn all the relevant paths into absolute ones, at purely aesthetic impact on the build logs.

msprotz:
I use `readlink -f` to convert from relative to absolute; on OSX, this has to be `greadlink` and the Everest script already makes sure you have the right utility.

```
jonathan@chartreuse:~/Code/hacl-star/code (migrating_sts_bis) $ greadlink -f ../../hacl-star/..
/Users/jonathan/Code
```

In the interest of fixing this soon, and minimizing the impact on other people, do you think:

```
HACL_HOME=$(shell $(READLINK) -f $(THIS_DIR)/..)
```

would work? I don't mind absolute paths in the build logs.

msprotz:
But what I don't understand is the semantics of Makefile's include directive.

If `code/ed25519/Makefile` does `include ../Makefile.include`, can't `Makefile.include` use relative paths to *itself*? Is this why we need to compute an absolute point of reference for the root of the build directory?

wintersteiger:
Yes, readlink (or realpath) does relative -> absolute, but the other direction isn't that easy. Makefile.include already uses it's own (absolute) path (THIS_DIR) to find the root (HACL_HOME_ABS), but I kept everything relative from there on. The Makefiles previously used relative paths everywhere (not just for ../Makefile.include, but even for ../bin/fstar.exe, etc), so it was easier for me to keep those parts relative while refactoring. `$(abspath ...)` and `$(realpath ...)` are native to gnu-make though, thus easily applicable to everything. (And I don't care about build log aesthetics either.)

The trickier situations are where make calls itself via $(MAKE) in a different directory, because from there on everything that was previously relative is not updated and thus not relative to the current directory anymore. So, I just keep `HACL_HOME_ABS` as a reference, but we could just adopt the convention to keep exclusively absolute paths in variables.

msprotz:
But we have committed to using GNU make everywhere, so you can freely make the assumption that   `$(abspath ...)` and `$(realpath ...)` are available, right?

msprotz:
The convention to use absolute paths everywhere in variables seems fine, right? This is what we would get anyhow if we were to use a `./configure`-style script?

PR #28, opened 2017-08-15
Add dependencies back for production nss

franziskuskiefer:
I think it's nicer to leave the dependencies in here. That makes the docker image way cleaner and makes it easier to run locally.
What do you think @beurdouche ?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ verify_all:
 	$(MAKE) verify -C specs
 	$(MAKE) verify -C code
 
-build:
+build: dependencies
 	@echo $(CYAN)"# Generating the HaCl* code (specialized for NSS)"$(NORMAL)
 	$(MAKE) nss-snapshot -C test
 	@touch build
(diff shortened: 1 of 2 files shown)


franziskuskiefer:
Is there a way to get the good revisions of F*/Z3/Hacl* from Everest?
The idea is to keep the Dockerfile for the NSS CI simple and have a single place where the good revisions are coming from. Right now this would be this branch (or whatever revision of HACL* we point the Dockerfile to), which "knows" which F* and Kremlin versions to use. You can have a look at the Dockerfile [here](https://nss-review.dev.mozaws.net/D411). It only checks out this branch and runs `make nss`.

prosecco:
I think having dependencies is good for now. It guarantees “known-good” versions.

msprotz:
You could also do:

```
git clone https://github.com/project-everest everest
(cd everest && git checkout <the-sha1-you-like>)
everest/everest kremlin pull make
everest/everest fstar pull make
```

if that suits your workflow better.

Makefile
```diff
@@ -30,12 +30,21 @@ nss: test
 	$(MAKE) nss-production -C test
 	@echo $(CYAN)"\nDone ! Generated code can be found in 'snapshots/nss-production'."$(NORMAL)
 
+dependencies:
+	@echo $(CYAN)"# Get and build F* and KreMLin"$(NORMAL)
+	opam switch 4.04.2
+	eval `opam config env`
+	git submodule update --init
+	opam config exec -- make -C dependencies/FStar/src/ocaml-output
```
beurdouche: You need to add the following line here.
`opam config exec -- make -C dependencies/FStar/ulib/ml`

PR #31, opened 2017-08-18
Review curve25519 (DO NOT MERGE !)

beurdouche:
DO NOT MERGE !

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@ HaCl-star
 HaCl* is a formally verified cryptographic library in [F\*],
 developed as part of [Project Everest].
 
-[![Build Status](https://travis-ci.org/mitls/hacl-star.svg?branch=master)](https://travis-ci.org/mitls/hacl-star)
-
 HACL stands for High-Assurance Cryptographic Library and its design is
 inspired by discussions at the HACS series of workshops.
 The goal of this library is to develop *reference* implementations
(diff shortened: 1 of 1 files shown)


beurdouche:
Per request from Ekr, multiple rounds of reviews on Curve25519 starting by a coding style...
Inlined suggestions in a PR as we don't have a proper code review tool ! We need one !

jkzinzindohoue:
How important is it to review the Low* code (@prosecco @beurdouche) ? The specs has received a lot of attention, so has the generated C code, and this is not a good time for me to rewrite the Low* code of curve25519 (although I admit that from a style standpoint it could be improved).
Also I am not in favor of the (**) notation as a long term solution. It is convenient at the moment for readability but there are thousands of lines of code, it is just too much to edit.
Instead I wish we could collapse the lines where the expressions return `unit` using the emacs fstar mode.
I will open an issue for that.

beurdouche:
You should not worry about it, there is absolutely no rush. The idea is more to make it a bit cleaner so that people looking at it will not judge the quality of the NSS code by the look of the HaCl code. I can do small coding style improvements by myself without touching the proofs anyway...
I am ok with not adding (**) and to find a better solution, but reading this code is currently very difficult. Also selecting expressions returning unit is not a solution as these could be computationally relevant if they are stateful.

jkzinzindohoue:
Sure, I meant `Tot unit` and `Ghost 'a pre post`

specs/Spec.Curve25519.fst
```diff
@@ -21,8 +22,11 @@ let zero : elem = 0
 let one  : elem = 1
 let ( +@ ) = fadd
 let ( *@ ) = fmul
+// BB: Are the operator definitions used ? Else remove.
+// BB: Add a blank line
 (** Exponentiation *)
 let rec ( ** ) (e:elem) (n:pos) : Tot elem (decreases n) =
+// BB: Is this an operator or a comment ?
```
beurdouche: Ohh that's the name :) Didn't look properly on first pass !

PR #44, opened 2017-09-26
Add targets and patches for NSS

franziskuskiefer:
These changes would allow NSS to move off the branch and use master.
Changes in hints don't have to be merged I can remove them.

I noticed that the `unit-tests` target doesn't work. I adapted the Makefile in kremlib to make it work. All other changes are NSS only and should be pretty minimal.

@beurdouche Can you have a look at this and let me know what you think?


Makefile
```diff
@@ -59,6 +59,14 @@ verify-secure_api: secure_api.dir-verify
 verify: .verify-banner verify-ct verify-specs verify-code verify-secure_api
 	@echo $(CYAN)"\nDone ! Please check the verification output"$(NORMAL)
 
+verify-nss:
```
beurdouche: I am not convinced, this should be part of our top-level Makefile, as it is NSS specific...
If I remember correctly the overall verification for the full HACL* (currently in NSS) is faster than the average NSS CI run, so it should not affect it. Am I correct ?

Makefile.build
```diff
@@ -83,6 +83,17 @@ snapshots/hacl-c:
...
 		./snapshots/hacl-c
 
+snapshots/nss: snapshots/snapshot-gcc-unrolled
+	# Compile everything but only copy out the things NSS is using.
+	mkdir -p snapshots/nss
+	cp snapshots/snapshot-gcc-unrolled/Curve25519.c snapshots/nss/hacl_curve25519_64.c
+	cp snapshots/snapshot-gcc-unrolled/Curve25519.h snapshots/nss/hacl_curve25519_64.h
+	cp snapshots/kremlib/FStar.h snapshots/nss/fstar_uint128.h
+	cp snapshots/kremlib/kremlib.h snapshots/nss
+	cp snapshots/kremlib/gcc_compat.h snapshots/nss
```
beurdouche: In the current version of `kremlib` in NSS, `gcc_compat.h` is inlined into `kremlib.h`.
This will require either a patch to NSS or an inlining.

code/nss-patches/license.txt
```diff
@@ -0,0 +1,13 @@
+// Copyright 2016-2017 Microsoft Corporation
```
beurdouche: This should be at the very least `INRIA and Microsoft Corporation`

snapshots/kremlib/Makefile
```diff
@@ -80,7 +82,7 @@ unit-tests: libhacl.so
 	-I . -I ../snapshots/kremlib -I ../../test/test-files/ -I $(TWEETNACL_HOME) \
 	$(TWEETNACL_HOME)/tweetnacl.c ../../test/test-files/hacl_test_utils.c \
 	../../test/test-files/unit_tests.c libhacl.so -o unit_tests.exe
-	./unit_tests.exe
+	LD_LIBRARY_PATH=. PATH=. DYLD_LIBRARY_PATH=. ./unit_tests.exe
```
beurdouche: Nice catch, I think this should be done with `unit_tests32.exe` too

Makefile
```diff
@@ -106,6 +114,10 @@ unit-tests:
 	@echo $(CYAN)"# Testing the HaCl* shared library"$(NORMAL)
 	$(MAKE) -C snapshots/hacl-c unit-tests
 
+unit-tests-unrolled:
```
beurdouche: This should definitely not be at top-level but handled in `test/Makefile`

code/nss-patches/patch.sh
```diff
@@ -0,0 +1,12 @@
...
+
+cwd=$(cd $(dirname $0); pwd -P)
+
+# Add license headers to all files.
+for file in `find "$cwd/../../snapshots/nss/" -type f`; do
+    echo -e "$(cat $cwd/license.txt)\n\n$(cat $file)" > "$file"
+done
+
+# Patch files where file names changed.
+patch "$cwd/../../snapshots/nss/hacl_curve25519_64.h" < "$cwd/hacl_curve25519_64.h.patch"
```
beurdouche: I tried your branch and it got rejected on my machine...
```
patch: **** rejecting file name with ".." component: /opt/hacl-star/code/nss-patches/../../snapshots/nss/hacl_curve25519_64.h
patch: **** rejecting file name with ".." component: /opt/hacl-star/code/nss-patches/../../snapshots/nss/hacl_curve25519_64.c
```

code/nss-patches/license.txt
franziskuskiefer: It's just copied from your branch ;) I'll change it.

snapshots/kremlib/Makefile
franziskuskiefer: Yep, this needs to go there as well. I didn't run them. But I'll add it.

code/nss-patches/patch.sh
franziskuskiefer: Hm, weird. How did you call the make file?

Makefile
franziskuskiefer: We never verified everything in the NSS CI. The runtime of verifying Curve25519 is fine. But verifying everything will take quite a bit longer.

Makefile
franziskuskiefer: Well, if we want to move NSS to master these targets have to go somewhere. What's your suggestion where to put them?

Makefile
beurdouche: Ok, I can take that as-is and move it later when I find a better place

Makefile
beurdouche: As mentioned before: `test/Makefile`

Makefile.build
msprotz: I've reverted this change on kremlin master -- it's useful for some situations to only have gcc_compat.h

code/nss-patches/patch.sh
msprotz: is it a gnu vs bsd utility issue?

Makefile.build
beurdouche: This means we would have multiple unverified files instead of just one. I'll inline it mechanically  as `kremlib.h` is our TCB in HACL*.

Makefile
franziskuskiefer: obviously... 😬

code/nss-patches/patch.sh
franziskuskiefer: Hm, I can't really test this right now because `make prepare` isn't working on Mac.

PR #46, opened 2017-10-02
Rewrite Buffer.Utils.xor_bytes_inplace

### LINKED ISSUE #43: rewrite Buffer.Utils.xor_inplace using a loop
Right now, it's a tail-recursive function that lives in `secure_api/utils/Buffer.Utils.fst`.

Now that we have support for loops, it'd be great to rewrite like it's done in other places in HACL*, so that we're not at the mercy of a fragile TCO optimization by the C compiler:

```
code/hmac/Hacl.HMAC.SHA2_256.fst:let xor_bytes_inplace a b len = C.Loops.in_place_map2 a b len (fun x y -> H8.logxor x y)
```

CC @BarryBo who initially flagged this, and @jroesch who might be interested in fixing it.

This would also be good for the WASM backend which doesn't have TCO.

jroesch:
This appeared to successfully re-extract on my machine, just waiting on CI to come back.

Fixes #43 

cc @protz @BarryBo


beurdouche:
This is now verifying, so if there are no objections, I'll merge it very soon...
https://msresearch-ext.visualstudio.com/Everest/_build/index?buildId=13141&_a=summary

PR #48, opened 2017-10-06
hacl-c compilation fixes for embedded targets

### LINKED ISSUE #47: Compilation of HACL* on bare metal targets
Hello,

the generated C code does not compile for bare-metal embedded targets (i.e. no OS). The problem is calling `printf()` and `exit()` functions (for example here https://github.com/mitls/hacl-star/blob/master/snapshots/hacl-c/kremlib.h#L76 ).

Another thing is missing defines for `htolen64()` and similar for non-linux embedded targets - there probably have to be defined manually (here is an example here https://github.com/jedisct1/libhydrogen/blob/v0/impl/common.h ).

I can manually edit the generated `kremlin.h` file to get it working, but I doubt that is the right way to go. I am happy to edit the template for autogenerated files if you can point me in the right direction.

How does removing `printf()` and `exit()` (or defining them as empty functions for bare-metal targets) affects the verification proof?

podhrmic:
Added byte swapping operations for embedded targets (no-std)

Removed `exit()` ad `printf()` calls for embedded targets

Fixes #47


beurdouche:
This should be fixed in the new version of Kremlin. I am closing this and keeping track in https://github.com/mitls/hacl-star/issues/47 until the next release of the C code (hopefully within 2 weeks).

PR #50, opened 2017-10-06
Add `attribute unused` to silence compiler warnings

podhrmic:
I am using the C snapshot as a part of a larger embedded project where I can't control the compiler flags (and using a precompiled library is not an option either). To avoid excessive warnings about unused functions, I added `__attribute__((unused))`

it has no effect on normal operation, and I am assuming that since the hacl-c is based on a verified library, the unused functions are not a problem,


beurdouche:
Closing as mentionned in https://github.com/mitls/hacl-star/issues/51.
We can't edit the generated code manually, so we'll have to find another solution

PR #73, opened 2017-10-25
Fix some make prepare paths

franziskuskiefer:
`HACL_HOME` is a relative path. To make the `make prepare` target work you have to either make `HACL_HOME` absolute or drop the paths (as I did here).


franziskuskiefer:
Also note that `make prepare` doesn't work on Mac at the moment because of https://github.com/FStarLang/FStar/pull/1320 (and some other issues I didn't figure out yet).

CC @beurdouche

PR #78, opened 2017-11-20
Update Makefile so even the static library is cleared

podhrmic:
Clears the static library generated in C snapshot directory

### DIFF
--- a/snapshots/hacl-c/Makefile
+++ b/snapshots/hacl-c/Makefile
@@ -145,4 +145,4 @@ unit-tests32: libhacl32.so
 	LD_LIBRARY_PATH=. DYLD_LIBRARY_PATH=. ./unit_tests32.exe
 
 clean:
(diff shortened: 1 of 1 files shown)


beurdouche:
Thanks !

podhrmic:
@beurdouche this change was recently overwritten by 7959da317504a6e1754d2fb1b1d72b499b3ccf8a

Perhaps changing the Makefile in kremlib folder would make more sense: https://github.com/mitls/hacl-star/blob/master/snapshots/kremlib/Makefile ?

beurdouche:
Indeed, sorry I didn't notice your patch was for `snapshots/hacl-c` that is overwritten when copied from `snapshots/kremlib`.
Fixed by https://github.com/mitls/hacl-star/commit/43b018b789e77fba10b800344758956a6fbec039

PR #98, opened 2017-12-07
No poly1305 verification in verify-nss

franziskuskiefer:
Let's disable this for now so we can proceed. This seems a bigger issue in the NSS CI that we can't resolve right now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ verify-nss:
 	$(MAKE) verify -C code/salsa-family
 	$(MAKE) Spec.Chacha20.fst-verify -C specs
 	$(MAKE) ct -C code/poly1305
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -72,7 +72,9 @@ verify-nss:
 	$(MAKE) verify -C code/salsa-family
 	$(MAKE) Spec.Chacha20.fst-verify -C specs
 	$(MAKE) ct -C code/poly1305
-	$(MAKE) verify -C code/poly1305
+	# Verification of poly1305 is disabled for now because of CI issues.
```
beurdouche: It might be good to slightly change the sentence to make sure that people understand that Poly is forally verified but that verification is disabled in that specific target which is only used by Mozilla's CI

Issue #100, opened 2017-12-11
vec128.h issues


franziskuskiefer:
There are a couple issues in vec128.h.

- [x] arm neon doesn't compile (first part of the fix is in #89)
`/vec128.h:198:57: error: ‘vec {aka struct <anonymous>}’ has no member named ‘vec_xor’ `
- [x] On intel the `_mm_shuffle` is broken, i.e `n%4` has to be constant
`vec128.h:87:20: error: the last argument must be an 8-bit immediate
     r.v = ((vec128)_mm_shuffle_epi32((__m128i)x.v, _MM_SHUFFLE((3 + n) % 4, (2 + n) % 4, (1 + n) % 4, n % 4)));`
- [ ] The code currently supports only SSSE3. It should be possible to make this work with SSE2 and without SSSE3 so that there's an option for people without SSSE3 to use the vectorised version.
- [ ] `__SSSE3__` is not defined for GCC.
- [x] `/usr/lib/gcc/aarch64-linux-gnu/5/include/arm_neon.h:22755:10: error: argument 3 must be a constant immediate`.  Argument 3 in `vsriq_n_u32` must be a constant immediate.
- [x] `__attribute__ ((vector_size (16)))` is a GCC extension and not valid C.

There might be more but this is as far as I got.

franziskuskiefer:
There are only two points open here.
* We work around the `__SSSE3__` in the NSS build system. Not sure if that's something you want to fix here as well.
* An SSE2 version would be nice, but involves writing new code. Not sure if it's worth the effort.

beurdouche:
Marina @polubelova and Karthik @karthikbhargavan have a lock on Lib.IntVector.fst so I am assigning this to you. If it is not relevant anymore, please close the issue.

PR #105, opened 2018-01-18
Fix ARM vec128.h code and inline ChaCha rounds

franziskuskiefer:
Here are two fixes for ARM for issue #100 .
* Inline the line function for ChaCha rounds. This doesn't work on ARM otherwise because argument 3 in `vsriq_n_u32` must be a constant immediate.
* `vec_rotate_left` reused variables.

This makes Chach20_Vec128 compile on ARM. However it is not correct as it's failing unit tests. (I didn't look into it any further yet.)

cc @beurdouche @karthikbhargavan


karthikbhargavan:
Hm. Let me try it again. It should pass tests so its concerning.

franziskuskiefer:
All the issues in here were addressed.

PR #106, opened 2018-01-26
vec128 windows fixes

franziskuskiefer:
This makes vec128.h usable with MSVC.
The main change is that `vec` is simply `_m128i`, which then requires a couple other small changes. Everything else is clean-up really.


franziskuskiefer:
all of this was addressed.

PR #107, opened 2018-02-07
sync with snaptshot/kremlib version

franziskuskiefer:


prosecco:
vec128 sync across directories

PR #116, opened 2018-03-08
gcm spec and tests

franziskuskiefer:
Not sure if this is exactly what we want. It's more to show what's missing :slightly_smiling_face:


specs/Spec.GF128.fst
```diff
@@ -44,7 +44,7 @@ let poly len text r =
   acc
 let finish (a:elem) (s:tag) : Tot tag = decode (a `fadd` (encode blocksize s))
 
-let gmac (len:size_nat) (text:lbytes len) (k:key) : Tot tag  =
```
ad-l: Not sure why you want this renaming - this function computes the full polynomial rather than a single multiplication, so `gmac` seems more appropriate than `gmul`

franziskuskiefer: The function is usually called something with `mul`. But I re-named again. I'm not fuzzy about the name here.

franziskuskiefer:
Abandoning this one in favour of #118

PR #119, opened 2018-04-06
aes gcm 128 and intel fmul

franziskuskiefer:
This is PR #118 with the addition of an `fmul` implementation for GF128 modelled after C code using intel intrinsics.


karthikbhargavan:
dev_specs is dead. 
long live dev.

PR #120, opened 2018-04-12
Update CMakeLists.txt to add install targets for the libraries and the header files

romen:
This PR modifies the `CMakeLists.txt` in the `/snapshots/hacl-c` and in the `/` folders, so that when the Makefile is generated from them a `make install` target is created.

After a successful `make build` in the root directory, installing the compiled libraries is then just  a matter of `cd build && make install`.

The prefix for installation can be tweaked when calling cmake by invoking it as
`cmake -DCMAKE_INSTALL_PREFIX:PATH=/path/to/prefix ..`.

The static and the shared libraries are installed to `$PREFIX/lib` while the header files to `$PREFIX/include/hacl`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,28 @@ set(SOURCE_FILES_EXPERIMENTAL
    snapshots/hacl-c-experimental/drng.c
    snapshots/hacl-c-experimental/Hacl_Random.c)
 
+# Public header files
+set(HEADER_FILES
+    snapshots/hacl-c/AEAD_Poly1305_64.h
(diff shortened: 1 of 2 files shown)


beurdouche:
@romen Thank you for the patch ! I cherry-picked the commit for the root CMakefile, I am currently working on the snapshot generation at the moment, so I will update the `hacl-c` snapshot with this change soon enough...
https://github.com/mitls/hacl-star/commit/72f40a4e500f50f54fd19ecb76517c281a1aa48d

romen:
@beurdouche  Thanks for the quick feedback!

PR #122, opened 2018-05-10
Crypto.AEAD.Main.fsti: cut dependency on FStar.Pointer

tahina-pro:
So that miTLS does not depend on FStar.Pointer or anything like that.
Also, later, we might switch the implementation to using FStar.Modifies, new modifies clauses which are not necessarily related to ArrayStruct.


tahina-pro:
Please disregard this pull request, I chose the wrong base.

PR #123, opened 2018-05-10
Crypto.AEAD.Main.fsti: cut dependency on FStar.Pointer

tahina-pro:
So that miTLS does not depend on FStar.Pointer or anything like that.
Also, later, we might switch the implementation to using FStar.Modifies, new modifies clauses which are not necessarily related to ArrayStruct.


secure_api/Crypto.AEAD.Main.fsti
```diff
@@ -121,7 +121,8 @@ let fresh_addresses (rid:HS.rid) (addrs:FStar.TSet.set address) (m0:HS.mem) (m1:
 (*                 | SomeRefs addrs -> TSet.complement addrs in *)
 (*               FStar.Heap.modifies_t mod_refs (Map.sel h0.h r) (Map.sel h1.h r))))) *)
 
-let fp = FStar.Ghost.erased FStar.Pointer.Base.loc
+noextract val fp' : Type u#0
```
msprotz: I think you can use both `noextract inline_for_extraction`

secure_api/Crypto.AEAD.Main.fsti
```diff
@@ -300,5 +301,5 @@ val decrypt
   (ensures (fun h0 verified h1 ->
                invariant st h1 /\
                enc_dec_liveness st aad plain cipher_tag h1 /\
-               // modifies_fp FStar.Pointer.Base.(loc_union (footprint st) (loc_buffer (Plain.as_buffer plain))) h0 h1 /\
+               // modifies_fp FPLOC.(loc_union (footprint st) (loc_buffer (Plain.as_buffer plain))) h0 h1 /\
```
msprotz: where is `FPLOC` defined?

msprotz: I guess it's just an abstract reference for later on?

msprotz:
Please see if things work out by leaving the `let` in the interface, marking it as `inline_for_extraction noextract` then you have my 👍 to commit. Thanks!

secure_api/Crypto.AEAD.Main.fsti
tahina-pro: yes, it could be replaced with FStar.Modifies.

tahina-pro:
Hmm, I'm not sure I understand your comment. I actually want to say that `loc` is an abstract ghost type, which is why I need to go through an auxiliary `loc'` type that will not be extracted.

msprotz:
Ok, I understand now. But this only fixes clients of the library, right? When extracting secure_api, we will still have FStar.Pointer.Base in our dependency graph.

I have the feeling that the uses of `fp` were only local and tentative. Is it possible to remove the uses of `fp` as a temporary band-aid to restore extraction and the everest build?

tahina-pro:
Yes, this is only to fix miTLS for now; secure_api/aead/Crypto.AEAD.Main.fst still depends on FStar.Pointer.
I can't judge it by myself, but if you are confident about the fact that nobody is using `fp` so far, then yes, I can remove it from the interface altogether.

msprotz:
Unless we fix it for HACL* too, this won't buy us much since everest upgrades will keep failing.

I would suggest you go ahead and comment out fp for the moment. Then, we see what happens at the next everest upgrade.

PR #127, opened 2018-07-26
Align convention for constants towards size_<name>

beurdouche:
Moving away from `blocklen`, `blocksize` or `<random names>` towards `size_<name>`...
Any objections ? @polubelova @karthikbhargavan (CC: @s-zanella)
This is likely to break extraction of the code, but we need to do a pass on it anyway...


beurdouche:
No objection from Marina.

beurdouche:
@s-zanella It is just that I haven't pushed changes in the code as we are gonna rewrite most of it anyways...

beurdouche:
We don't need that now...

PR #130, opened 2018-10-26
Updated AESNI detection

wintersteiger:
@protz I can't tell which `cpuid.h` EverCrypt is supposed to #include. I currently need this to work with clang-7 on Ubuntu 16.04 and 18.04. Below is the bit if AESNI detection from Amaury, which, I think, should work for more systems, but I don't know what systems you need this to work with.

### DIFF
--- a/providers/evercrypt/c/evercrypt_autoconfig.c
+++ b/providers/evercrypt/c/evercrypt_autoconfig.c
@@ -49,15 +49,16 @@ bool detect_aesni() {
 
 #else // _MSC_VER
 
-#include <cpuid.h>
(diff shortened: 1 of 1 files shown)


beurdouche:
Can we merge or close this PR please ?

parno:
I think this is all being handled by verified Vale code as part of EverCrypt, thanks to @protz .

msprotz:
Correct. This can be closed and this code will cease to exist as soon as I merge.

~ jonathan

beurdouche:
Thanks !

PR #132, opened 2018-11-27
Merge _dev_frodo

s-zanella:
This is mostly porting SHA3 and Frodo to the latest version of `lib`, but it also includes some changes to `lib` itself. I checked that SHA3, Frodo and lib verify. I still get errors verifying Blake2 but so I do in `_dev`.
I don't expect this to cause any regressions. I'll wait until the end of the day before merging to give you a chance to review the changes if interested (go to the end of the page to see the changes to `lib`).

Some notes I gathered that I would like to discuss:

- The main point of having a single parametric type of buffers for mutable and immutable buffers is to be able to write functions that can take as argument either a mutable or an immutable buffer as long as the immutability is respected. However, even the simplest example keeps the mutability tag as an explicit argument when extracted which is very off-putting:
```FStar
module Test

open FStar.HyperStack.ST
open Lib.Buffer

let uint8 = UInt8.t

val f: #t:buftype -> b:lbuffer_t t uint8 2ul -> Stack uint8
  (requires fun h -> live h b)
  (ensures  fun h0 _ h1 -> modifies (loc b) h0 h1)
let f #t b =
  b.(0ul)

let main () : St C.exit_code =
  push_frame();
  let b = create 2ul 0uy in
  let y = f b in
  pop_frame();
  C.EXIT_SUCCESS
```
extracts to
```
...
uint8_t f(Lib_Buffer_buftype t, void *b)
...
exit_code main()
{
  uint8_t b[2U] = { 0U };
  uint8_t y = f(Lib_Buffer_MUT, (void *)b);
  return EXIT_SUCCESS;
}
```
Ideally we would like the tags to be erased and enable things like this commit I had to revert:
https://github.com/project-everest/hacl-star/commit/7b3be54b0aaabeba06dda3d4f2fbf5a5622bc7ca

- I don't understand the need for the `modifies_includes` and `modifies_trans}` in `Lib.Buffer`. They don't have a pattern and they don't seem to be used anywhere. They are also directly provable from lemmas in `LowStar`.

- Why are `modifies{1,2,3}` restricted to `lbuffer`. They could be generalized to `buffer`.


s-zanella:
Thanks Benjamin. My criteria for inserting/removing unfolds is that it's generally bad to unfold a term defined as a `match` and is generally good to unfold mere abbreviations as long as they aren't used in SMT patterns.

karthikbhargavan:
Looks good to me. I am seeing quite a bit of fragility in "unfold"s so would love to have a formula like yours and see if it works for most of our use cases.

PR #134, opened 2018-12-14
Lib.Buffer: Make b1 and b2's types and lengths unrelated in modifies3

kevinmkane:
In modifies3, argument b2 refers to #a1 and #len1, which are also used for
the type and length of b1. Change to #a2 and #len2 which are declared but
not used, as there's no need for the types and lengths of these two buffers
to be the same.


beurdouche:
@kevinmkane Thanks Kevin ! I just caught and fixed a similar typo in newer versions of the library thanks to you : )

PR #146, opened 2019-03-02
one-way live sub only

tahina-pro:
With @aseemr  and @nikswamy , we intend to remove the precondition on `LowStar.Monotonic.Buffer.mgsub`. As a consequence, `live_sub` has to be weakened: one can only prove the liveness of a subbuffer from the liveness of its enclosing buffer, not the converse. (More precisely, see https://github.com/FStarLang/FStar/pull/1667#issuecomment-468800761 for the proposed change in F*)

It seems that I can already successfully weaken `live_gsub` in HACL* accordingly, per this pull request: with the current F* master, I have HACL* green CI; and with FStarLang/FStar@310b90cd3fd4bddc12a95ab31679e4572935c674, I have a local successful `everest make test verify`.

Do you foresee any problem in no longer being able to derive the liveness of a buffer from the liveness of one of its subbuffers?

Thanks,


s-zanella:
Thanks for the heads-up!
I tested this change in `_dev` and `_dev_combinators` (this includes the rest of `lib`, SHA3, Frodo, ChaCha20) and it didn't cause any regressions.

PR #150, opened 2019-04-08
Correct a typo

felixonmars:


msprotz:
Good catch! Thanks for fixing it.

PR #153, opened 2019-04-18
Low* tests: restore verification, improve platform coverage

tahina-pro:
This pull request tries to:
* restore verification for EverCrypt Low* tests, removing most admits.
  The remaining admits are:
  - disjointness of plaintext and aad buffers for AEAD. There is at least one solution to this problem, e.g. https://github.com/project-everest/hacl-star/tree/taramana_tests_disjoint_plaintext_aad_vectors, but stumbling upon extraction issues, so that solution is **not yet** included in this pull request.
  - false preconditions on legacy EverCrypt.fst functions
* ensure that all tests (hash, hmac, hkdf, aead, aes128gcm, cipher, curve25519, poly1305, chacha20, chacha20poly1305) are performed on all platforms (Vale, HACL*, HACL*+AVX, HACL*+AVX2, OpenSSL, BCrypt).
  At this point I am not sure that it is really the case, we should have a closer look.


msprotz:
Thanks Tahina, this is much cleaner and much better now.

PR #159, opened 2019-04-30
Taramana tests

msprotz:


providers/test/Test.fst
```diff
@@ -296,31 +296,46 @@ let rec test_aead_loop (LB len vs) =
 let test_aead () : St unit =
   test_aead_loop aead_vectors_low
 
-let rec test_chacha20poly1305_loop (i: U32.t): St unit =
-  let open Test.Vectors.Chacha20Poly1305 in
-  if i `U32.gte` vectors_len then
-    ()
+#push-options "--z3rlimit 32"
+
+let test_chacha20poly1305_one (v: Test.Vectors.Chacha20Poly1305.vector): Stack unit (fun _ -> True) (fun _ _ _ -> True) =
```
msprotz: can this one go in Test.NoHeap since it is now in `Stack`?

providers/test/Test.fst
```diff
@@ -359,126 +374,347 @@ let test_dh () : St unit =
...
+      match i with
+      | Hacl -> SC.hacl
+      | Vale -> SC.vale
+      | OpenSSL -> SC.openssl
+      | BCrypt -> SC.bcrypt
+
+#push-options "--z3rlimit 16"
+
+inline_for_extraction
+let set_config (c: config) : Stack unit (fun _ -> True) (fun _ _ _ -> True) =
```
msprotz: so there's an understanding here that the client always calls `check_config` before calling `set_config`... can there be a comment to that effect?

providers/test/Test.fst
```diff
@@ -359,126 +374,347 @@ let test_dh () : St unit =
...
+  (Hacl, f_none) `ts_cons`
+  external_test_set
+
+inline_for_extraction
+noextract
+let hash_test_set =
+  (Vale, f_none) `ts_cons` (
+  (Vale, f_shaext) `ts_cons` (
+  (Hacl, f_none) `ts_cons` (
+  external_test_set)))
```
msprotz: we no longer use openssl or bcrypt for hashes; is `external_test_set` still needed here?

providers/test/Test.fst
```diff
@@ -359,126 +374,347 @@ let test_dh () : St unit =
...
+inline_for_extraction
+let test_chacha20poly1305_body (print: C.String.t -> St unit) : St unit =
+    print !$"  >>>>>>>>> AEAD (ChachaPoly vectors)\n";
+    test_chacha20poly1305 ()
+
+inline_for_extraction
+noextract
+let test_hash_body (print: C.String.t -> St unit) : St unit =
+    print !$"  >>>>>>>>> Hash (Test.Hash)\n";
+    Test.Hash.main ();
```
msprotz: this test now has very little value; can we remove it and simplify?

providers/test/Test.fst
```diff
@@ -296,31 +296,46 @@ let rec test_aead_loop (LB len vs) =
...
+  let Test.Vectors.Chacha20Poly1305.Vector cipher_and_tag cipher_and_tag_len plain plain_len aad aad_len nonce nonce_len key key_len = v in
+  if not (
+    key_len = 32ul &&
+    nonce_len = 12ul &&
+    0ul `U32.lt` plain_len &&
+    (4294967295ul `U32.sub` 16ul) `U32.gte` plain_len &&
+    (plain_len `U32.div` 64ul) `U32.lte` (4294967295ul `U32.sub` aad_len) &&
+    cipher_and_tag_len = plain_len `U32.add` 16ul
+  ) then
+    C.String.print !$"Warning: skipping a chacha20poly1305 instance because bounds do not hold"
```
msprotz: should we make these fatal errors? if a test is poorly crafted then surely there's a big problem, isn't there?

providers/test/Test.fst
```diff
@@ -359,126 +374,347 @@ let test_dh () : St unit =
...
+  features_avx = false;
+  features_avx2 = false;
+  features_bmi2 = false;
+  features_adx = false;
+  features_aesni = false;
+  features_shaext = false;
+}
+
+inline_for_extraction
+let f_avx : features = // [@inline_let] ({ f_none with features_avx = true; })
```
msprotz: why didn't this work? should we fix it?

providers/test/Test.fst
```diff
@@ -359,126 +374,347 @@ let test_dh () : St unit =
...
+  features_avx = false;
+  features_avx2 = false;
+  features_bmi2 = false;
+  features_adx = false;
+  features_aesni = false;
+  features_shaext = true;
+}
+
+inline_for_extraction
+type impl = | Hacl | Vale | OpenSSL | BCrypt
```
msprotz: I think we should get rid of OpenSSL / BCrypt -- they're here only for legacy purposes and we're getting rid of those legacy functions very soon

providers/test/Test.fst
```diff
@@ -359,126 +374,347 @@ let test_dh () : St unit =
...
+    print !$"  >>>>>>>>> Hmac (Test.NoHeap)\n";
+    test_hmac hmac_vectors_low;
+    print !$"  >>>>>>>>> HKDF (Test.NoHeap)\n";
+    test_hkdf hkdf_vectors_low
+
+inline_for_extraction
+noextract
+let test_default_body (print: C.String.t -> St unit) : St unit =
+    print !$"  >>>>>>>>> FINITE-FIELD DIFFIE-HELLMAN\n";
+    test_dh ();
```
msprotz: this test is disabled and doesn't do anything; can we remove it?

providers/test/Test.fst
```diff
@@ -359,126 +374,347 @@ let test_dh () : St unit =
...
+    print !$"  >>>>>>>>> Chacha20\n";
+    test_chacha20 chacha20_vectors_low
```
msprotz: this leaves only the chacha20 test; can we make this a `test_chacha20_body` with, for the time being, only hacl/none? in the future, there will be hacl/avx, and possibly more

providers/test/Test.fst
tahina-pro: In fact, the `{x with field = value}` syntax makes F* add a `let` binding, which would have to be inlined for extraction. This is not the case currently. Ideally we should allow the user to ask for this introduced `let` binding to be inlined at extraction, through, say, `[@inline_let]`. Right now this is not done. So I think it is really on the F* syntax side.
(More generally, we should review all F* "syntactic sugar" constructs silently introducing `let` bindings.)

providers/test/Test.fst
tahina-pro: Thanks for confirming.
I initially thought that we should test OpenSSL and BCrypt for the purpose of testing the "bindings" between EverCrypt and existing libraries.

providers/test/Test.fst
tahina-pro: Thanks for confirming. Will do.

providers/test/Test.fst
tahina-pro: Thank you, this is exactly what I wanted to know: which test configurations we need for chacha20.

providers/test/Test.fst
tahina-pro: OK, will do. Same for poly1305 and curve25519 above. Then I let you decide whether I should also add it to "the Wasm tests" (`Test.NoHeap.main`)

msprotz: Yes please

PR #163, opened 2019-06-05
signed integers in Lib.IntTypes

karthikbhargavan:


lib/Lib.IntTypes.fst
```diff
@@ -322,7 +495,18 @@ let lognot #t #l a =
   | U32  -> UInt32.lognot a
   | U64  -> UInt64.lognot a
   | U128 -> UInt128.lognot a
-
+  | S8 -> Int8.lognot a
+  | S16 -> Int16.lognot a
+  | S32 -> Int32.lognot a
+  | S64 -> Int64.lognot a
+
+let shift_right_negative (#n:pos) (a:Int.int_t n{a < 0}) (s:nat) : Tot (Int.int_t n) =
```
kevinmkane: This looks like you're manually implementing the arithmetic right shift, unless there's some magic here I'm missing that uses this just for spec and this ends up being extracted as native operators. @s-zanella added support in `FStar.Int` and KreMLin for an arithmetic right shift that extracts to the `>>` operator on compilers that will emit a native arithmetic right shift instruction (`sar` on Intel). How does this extract?

lib/Lib.IntTypes.fsti
```diff
@@ -580,98 +751,101 @@ let ( ~. ) #t #l = lognot #t #l
 ///
 
 inline_for_extraction
-val div: #t:inttype{t <> U128}
-  -> a:uint_t t PUB
-  -> b:uint_t t PUB{uint_v b > 0}
-  -> uint_t t PUB
+val div: #t:inttype{t <> U128 (*/\ t <> S128*) /\ unsigned t}
```
s-zanella: Do we want to generalize this to signed types?

lib/Lib.IntTypes.fst
```diff
@@ -60,6 +84,24 @@ let size_to_uint64 x = FStar.Int.Cast.uint32_to_uint64 x
...
+  | U32 -> u32 x
+  | U64 -> u64 x
+  | U128 -> UInt128.uint_to_t x
+  | S8 -> i8 x
+  | S16 -> i16 x
+  | S32 -> i32 x
+  | S64 -> i64 x
+  | S128 -> Int128.int_to_t x
+  
+#reset-options "--z3rlimit 500"
```
s-zanella: This is excessive. Note that this is running with a quite high `fuel` and `ifuel` and it will succeed with lower fuel and less resources.

lib/Lib.IntTypes.fst
```diff
@@ -131,6 +268,11 @@ let add_mod #t #l a b =
   | U32  -> UInt32.add_mod a b
   | U64  -> UInt64.add_mod a b
   | U128 -> UInt128.add_mod a b
+  | S8 -> Int8.int_to_t (Int.add_mod (Int8.v a) (Int8.v b))
```
s-zanella: There are no `_mod` operators in `FStar.IntN` modules for good reason: there is no simple way of implementing them in C short of instructing the compiler to use wrapping semantics for signed integer under/overflow.
If such operators are generally useful, we could add them and make extraction fail unless passing some option to KreMLin (so as to not use wrapping semantics unless explicitly requested), with a visible warning that it will be unsound to compile the code without wrapping semantics.

For the time being, I think we should restrict `_mod` operators to `unsigned` types.

lib/Lib.IntTypes.fst
s-zanella: Why is this still here?

lib/Lib.IntTypes.fst
```diff
@@ -365,6 +729,10 @@ let ones t l =
     let y = (UInt128.shift_left x 64ul) `UInt128.add` x in
     assert_norm (UInt128.v y == pow2 128 - 1);
     y
+  | _  -> mk_int #t #l (ones_v t)
```
s-zanella: This can be implemented just as `mk_int (ones_v t)` without the case analysis

lib/Lib.IntTypes.fst
```diff
@@ -365,6 +729,10 @@ let ones t l =
...
+
+inline_for_extraction
```
s-zanella: See the comment above about wrapping semantics for signed integer arithmetic.

lib/Lib.IntTypes.fst
```diff
@@ -374,6 +742,13 @@ let eq_mask #t a b =
   | U32  -> UInt32.eq_mask a b
   | U64  -> UInt64.eq_mask a b
   | U128 -> UInt128.eq_mask a b
+  | S8   -> to_i8 #U8 #SEC (UInt8.eq_mask (to_u8 a) (to_u8 b))
```
s-zanella: Can we define a more efficient `eq_mask` operator for `signed` integers using `shift_arithmetic_right`?

lib/Lib.IntTypes.fst
```diff
@@ -424,41 +862,59 @@ let gt_mask #t a b = logand (gte_mask a b) (neq_mask a b)
...
   if uint_v m = 0 then
     UInt.logand_lemma_1 #(bits t) (uint_v a)
   else
-    UInt.logand_mask #(bits t) (uint_v a) (uint_v m)
-
+    UInt.logand_mask #(bits t) (uint_v a) (uint_v m) 
+  |_->
+  if uint_v m = 0 then
+    Int.logand_lemma_1 #(bits t) (sint_v a)
+  else admit()
```
s-zanella: Does it hold for `signed` types? Do we want to prove it?

lib/Lib.IntTypes.fsti
```diff
@@ -1,705 +1,817 @@
...
 
 inline_for_extraction
-val zeroes: t:inttype -> l:secrecy_level -> v:uint_t t l{uint_v v = 0}
-
-inline_for_extraction
-val ones: t:inttype -> l:secrecy_level -> v:uint_t t l{uint_v v = maxint t}
-
-inline_for_extraction
-val eq_mask: #t:inttype -> uint_t t SEC -> uint_t t SEC -> uint_t t SEC
+val eq_mask: #t:inttype{unsigned t} -> int_t t SEC -> int_t t SEC -> int_t t SEC
```
s-zanella: This requires `unsigned t`, which hides the fact that the implementation in `Lib.IntTypes.fst` wouldn't typecheck for signed types (the corresponding cases are dead code).

lib/Lib.IntTypes.fsti
```diff
@@ -1,705 +1,817 @@
...
 
-val eq_mask_lemma: #t:inttype -> a:uint_t t SEC -> b:uint_t t SEC -> Lemma
-  (requires True)
-  (ensures  (v a = v b ==> v (eq_mask a b) == maxint t) /\
-            (v a <> v b ==> v (eq_mask a b) == 0))
+val eq_mask_lemma: #t:inttype{unsigned t} -> a:int_t t SEC -> b:int_t t SEC -> Lemma
```
s-zanella: Same, this requires `unsigned t` so the proof given is actually not proving anything about signed types.

lib/Lib.IntTypes.fst
```diff
@@ -131,6 +249,11 @@ let add_mod #t #l a b =
   | U32  -> UInt32.add_mod a b
   | U64  -> UInt64.add_mod a b
   | U128 -> UInt128.add_mod a b
+  | S8   -> Int8.int_to_t (Int.add_mod (Int8.v a) (Int8.v b))
```
s-zanella: Removing these as discussed, but note that this wouldn't extract because it uses mathematical integers

lib/Lib.IntTypes.fst
```diff
@@ -1,127 +1,245 @@
...
-  | U8, U1 -> FStar.UInt8.rem u 2uy
+  | U1, U16 -> uint8_to_uint16 u
+  | U1, U32 -> uint8_to_uint32 u
+  | U1, U64 -> uint8_to_uint64 u
+  | U1, U128 -> UInt128.uint64_to_uint128 (uint8_to_uint64 u)
+  | U1, S8 -> uint8_to_int8 u
+  | U1, S16 -> uint8_to_int16 u
+  | U1, S32 -> uint8_to_int32 u
+  | U1, S64 -> uint8_to_int64 u
+  | U1, S128 -> pow2_le_compat (bits t' - 1) (bits t); mk_int #S128 #l' (v u)
```
s-zanella: Not extractable because it uses mathematical integers. Same for every other case using `v u`

lib/Lib.IntTypes.fsti
```diff
@@ -1,705 +1,817 @@
...
+val pow2_3: n:nat -> Lemma (pow2 3 = 8)  [SMTPat (pow2 n)]
+val pow2_4: n:nat -> Lemma (pow2 4 = 16) [SMTPat (pow2 n)]
 
 ///
 /// Definition of machine integer base types
 ///
 
 type inttype =
-  | U1 | U8 | U16 | U32 | U64 | U128
+  | U1 | U8 | U16 | U32 | U64 | U128 | S8 | S16 | S32 | S64 | S128
```
s-zanella: Supporting `S128` complicates is complicated.   
There is no code using it at the moment and any code using `S128` won't extract because there is no implementation of `int128_t` in KreMLin.
Shall we remove `S128` until we need it and it is supported in KreMLin?

s-zanella:
I'm happy with the current revision and would like to merge this into `_dev`.
I've ported the changes to the rest of `lib` and SHA3 and Frodo. I switched to Z3-4.8.5 at the same time and I got very few regressions due to these changes.

The rest of `specs` and `code` needs to be ported but this should be fairly easy and we can do it once we merge this PR. I added a `Lib.IntTypes.Compatibility` module that should be enough to `open` after `Lib.IntTypes` to get backwards compatibility. It also acts as a hint as to how to fix regressions.

lib/Lib.IntTypes.fst
```diff
@@ -374,25 +668,88 @@ let eq_mask #t a b =
...
-      logxor (u1 0) (u1 1) == u1 1 /\ logxor (u1 1) (u1 0) == u1 1 /\
+      logxor (u1 0) (u1 0) == u1 0 /\ logxor (u1 0) (u1 1) == u1 1 /\
+      logxor (u1 1) (u1 0) == u1 1 /\ logxor (u1 1) (u1 1) == u1 0 /\
       lognot (u1 1) == u1 0 /\ lognot (u1 0) == u1 1)
+  | U8 | U16 | U32 | U64 | U128 -> ()
+
+val eq_mask_lemma_signed: #t:inttype{signed t /\ ~(S128? t)} -> a:int_t t SEC -> b:int_t t SEC -> Lemma
+  (if v a = v b then v (eq_mask a b) == ones_v t
+                else v (eq_mask a b) == 0)
+let eq_mask_lemma_signed #t a b =
```
GuillaumeGETTE: This does not typecheck on my computer (Z3 4.8.5 says: unknown because unknown).
Someone has the same issue ?

s-zanella: It works for me, but it's close to the z3rlimit. I doubled it to be safe.

s-zanella:
Shall we merge this into `_dev`?

msprotz:
Could these land on fstar-master as well to get the benefits of CI, Everest regressions, etc ?

Thanks,

~ jonathan

s-zanella:
> Could these land on fstar-master as well to get the benefits of CI, Everest regressions, etc ? Thanks, ~ jonathan
Sure. I'll do it as soon as it's merged into `_dev`.

specs/Spec.SPARKLE.fst
```diff
@@ -355,7 +353,7 @@ let sparkle512 steps input =
       let b6 = arx rcon.[6] b6 in
       let b7 = arx rcon.[7] b7 in
       l8 (b0,b1,b2,b3,b4,b5,b6,b7)
-  ) (b0,b1,b2,b3,b4,b5,b7,b7)
```
s-zanella: Heads up @beurdouche  @karthikbhargavan  @polubelova if you're relying on this specification for anything.

beurdouche: Nice catch, Thanks Santiago! I have a plan to implement lightweight contestant and add tests later in the summer.

beurdouche:
Merged in 07e09613fead116da61cd0af37f56e71eaa36bbc

PR #164, opened 2019-06-08
To havoc or not to havoc, that is the question. Use [option bool], that is the answer.

jaybosamiya:
This PR changes the flags in the machine state into a new type, and makes all the requisite changes across the repository to handle this change.

Previously, flags were stored as a (bounded) integer and the flags we care about (cf and of) were got by using division/modulo (:fearful:). Additionally, havoc'ing of flags was done via an uninterpreted function. However, this meant that the flags actually would have _some_ readable value, even though it wouldn't make sense.

Now, flags are a map from flags to flagvalues. This allows us to remove the division and modulo operations. Additionally, it allows us to store more than just a 1 or 0 in the flag. Now, a flag can either be set or cleared or not-safely-readable. To store these three values, we use `option bool` as the `flag_val_t` type.

To be able to now use `cf(efl)` or `overflow(efl)` from the `.vaf` files, we now need to show that the flag is actually valid before we can read it. Otherwise, the code goes into a not-ok state.

The `updated_cf` and `updated_of` functions expose this nicely; and so do the `maintained_cf` and `maintained_of`. For places where these cannot be used directly for some reason, there is always `valid_cf` and `valid_of`.

I've gone ahead and updated all of the parts of the code base that are affected by this change, so hopefully this change should be minimally disruptive.

BTW, since this PR changes some of the trusted semantics, it will need a review before it is merged in.


parno:
@jaybosamiya Can you tell us a bit more about the motivation for this change?  I was also previously leery about specifying the flags via mod and div, but thus far, it seems to have mostly worked out, largely, I suspect, because the effects are hidden by the Hoare-logic layer.  

Regarding the havoc-based approach, is there a specific case where you're concerned that we're unsound?

jaybosamiya:
Well, the motivation is not really the mod and div, and I am not really worried about unsoundness via the havoc (although with the changes with this PR, I believe it helps clean things up for both). The havoc _might_ become unsound in the future if we update things and forget to correctly update the part the reads the state to get the havoc'd `nat64`, but that is more of a code-review problem (and a totally tangential discussion), so I won't talk about that further.

There are two parts to the motivation for this PR's change:
1. For the instruction reordering stuff that I've been working on, we want to maintain a notion of havoc / not-havoc'd for each flag separately (explanation below)
2. For the generic architecture stuff, as far as I'm aware, there was a plan to switch over to a map anyways, so a change from `nat64` to `flag -> bool` and _then_ making the change to `flag -> option bool` has been rolled into one change in this PR.

The reason we need a havoc/not-havoc'd for each flag separately is because we want to be able to say that "given two equivalent starting states, if we apply the same instruction, we get equivalent final states", and this behavior is impossible to show with the previous model. The notion of equivalent states (see https://github.com/project-everest/hacl-star/blob/13450b8b9c480bf4b45548880699947a3302e966/vale/code/lib/transformers/Vale.Transformers.InstructionReorder.fst#L93-L102) expects only the carry and overflow flags to be the same (amongst other non-flag things). Since the havoc depends on other flags too (rightly so), we have non-equivalent states after running a havocing instruction, despite equivalent evaluation. The fix for this was to either try to achieve full flag equivalence (which would disallow practically all movement of instructions, since the rest of the flags are not modelled) or introduce a notion of havoc within the state per flag. After a discussion with Chris, the latter seemed to be a better option, and thus this PR.

parno:
Thanks for the explanation -- that does make sense.  It does seem preferable to try to continue under-specifying the other flags as long as we can.

Chris-Hawblitzel:
Before we merge, though, I want to investigate the performance impact

Chris-Hawblitzel:
The build on the CI server was slower with this change, but I don't know why yet

parno:
Good point; definitely worth checking in on.  I would have hoped that the Hoare-logic layer would mostly hide the performance effects.

PR #165, opened 2019-06-20
Simplify defs and lemmas about memory locations

tahina-pro:
Shortcuts such as `loc_in`, `loc_not_in`, `fresh_loc` are now in F* master (`LowStar.Monotonic.Buffer`)

Instances of `loc_includes_union_l` are still necessary for Z3 to pick the right path.


msprotz:
Thanks Tahina. This is exactly as we discussed yesterday and looks good to me. Please merge as soon as you have a green!

PR #171, opened 2019-07-01
Ensure repo-specific config is read (if it exists)

jaybosamiya:
The `Makefile.config` is used by the main `Makefile` but it is not used in subdirectories. Instead `KREMLIN_HOME` etc are picked up from the environment. This causes issues (for example when the emacs interactive process tries to find the includes using `make --quiet asd.fst-in`, which then give erroneous results). This PR causes the config to be picked up from the `HACL_HOME` which means that one always gets the correct includes even when working interactively.

PS: I noticed this issue when I started working on `hacl-star` from a non-everest directory. If one was working from an everest directory, this issue wouldn't ever come up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,9 +40,7 @@
 # - min-test: staged, runs only a subset of verification for the purposes of
 #   F*'s extended CI
 
-# Put your local configuration (e.g. HACL_HOME, KREMLIN_HOME, etc.) in
-# Makefile.config
--include Makefile.config
+include Makefile.common
 
 #########################
 # Catching setup errors #
@@ -163,8 +161,6 @@ clean:
 #################
 # Configuration #
 #################
-
-include Makefile.common
 
 IMPORT_FSTAR_TYPES := $(VALE_HOME)/bin/importFStarTypes.exe
 PYTHON3 ?= $(shell tools/findpython3.sh)
--- a/Makefile.common
+++ b/Makefile.common
@@ -1,7 +1,7 @@
 # This Makefile can be safely included from sub-directories for the purposes of
 # defining the .fst-in targets, as long as the sub-directory defines HACL_HOME.
 
-HACL_HOME 	?= .
+HACL_HOME 	?= $(shell git rev-parse --show-toplevel 2>/dev/null || echo .)
 
 # Put your local configuration (e.g. HACL_HOME, KREMLIN_HOME, etc.) in
 # Makefile.config
--- a/Makefile.common
+++ b/Makefile.common
@@ -2,6 +2,7 @@
 # defining the .fst-in targets, as long as the sub-directory defines HACL_HOME.
 
 HACL_HOME 	?= .
+-include $(HACL_HOME)/Makefile.config
 KREMLIN_HOME 	?= $(HACL_HOME)/../kremlin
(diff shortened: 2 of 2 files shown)


msprotz:
```
diff --git a/Makefile b/Makefile
index 097414c0b..2b1ee88db 100644
--- a/Makefile
+++ b/Makefile
@@ -40,9 +40,7 @@
 # - min-test: staged, runs only a subset of verification for the purposes of
 #   F*'s extended CI
 
-# Put your local configuration (e.g. HACL_HOME, KREMLIN_HOME, etc.) in
-# Makefile.config
--include Makefile.config
+include Makefile.common
 
 #########################
 # Catching setup errors #
@@ -164,8 +162,6 @@ clean:
 # Configuration #
 #################
 
-include Makefile.common
-
 IMPORT_FSTAR_TYPES := $(VALE_HOME)/bin/importFStarTypes.exe
 PYTHON3 ?= $(shell tools/findpython3.sh)
 ifeq ($(OS),Windows_NT)
diff --git a/Makefile.common b/Makefile.common
index 8cffdd42e..ad9bcd8e1 100644
--- a/Makefile.common
+++ b/Makefile.common
@@ -2,6 +2,11 @@
 # defining the .fst-in targets, as long as the sub-directory defines HACL_HOME.
 
 HACL_HOME      ?= .
+
+# Put your local configuration (e.g. HACL_HOME, KREMLIN_HOME, etc.) in
+# Makefile.config
+-include $(HACL_HOME)/Makefile.config
+
 KREMLIN_HOME   ?= $(HACL_HOME)/../kremlin
 FSTAR_HOME     ?= $(HACL_HOME)/../FStar
 VALE_HOME      ?= $(HACL_HOME)/../vale
```

how about this diff instead?

jaybosamiya:
The changes look reasonable to me. I've made them above. Additionally, I made a small change to `HACL_HOME`'s default. Instead of picking `.` by default, it will try to pick the root of the git repository if it can, and default to `.` only if it is unable to.

This should actually reduce the necessity to manually set up `HACL_HOME` to the root of the repo always, and should have it work automatically. Do let me know if you think this change makes sense or we should change/revert it. Thanks!

msprotz:
Why do you need to set HACL_HOME locally? Is the default setup not working for you? Just trying to understand why we need to resort to git for files that may now be in the presence of a .git (e.g. downloaded archive for a release tag off of github).

jaybosamiya:
The fix that I've suggested (in  https://github.com/project-everest/hacl-star/pull/171/commits/d9fd61c5f8e4cf7ae2774e5d1134b3b9ab5a3903) goes via the `.git` if it can get it, otherwise it defaults to the `.` behavior (which means it should work even when not in the presence of a place with `.git`).

The issue that happened however, was that I had a stray configuration from some previous tests around, which messed with things and caused a `does not seem to point to the current directory.  Stop.` message, which I fixed via the above change but reverting the change still seems to work fine now that I've cleaned things up from previous tests, so looks like it was some local config issue.

I'm reverting the change, since the previous behavior is easier to reason about, as well as the fact that I can't seem to reproduce the original failing case which got fixed by the change. Sorry for the noise.

jaybosamiya:
Closing this PR since you've already made the core changes in the Makefile on `fstar-master`.

PR #172, opened 2019-07-03
Revamping CTR and AEAD specifications; a new EverCrypt.CTR

msprotz:
This pull request contains many changes.
- AEAD specifications are now more abstract and the notion of expanded key does not appear in the specification's interface (@ad-l and @fournet, please review spec changes)
- Key expansion is now entirely an implementation matter, from the perspective of both EverCrypt.AEAD and EverCrypt.CTR
- Specifications now distinguish between a notion of expanded key, as mandated by NIST (now in a separate spec module) and a notion of *concrete* expanded key, that may contain more precomputations, as mandated by a given implementation.
- As such, a shared `impl` type (between CTR and AEAD) now determines the *concrete* expanded key, while the usual cipher algorithm type (also shared between CTR and AEAD) determines the expanded key.
- Several helpers, both in the spec (iv, key, expanded key, etc. etc. etc.) and in the implementation (concrete key expansion for vale, machine lengths, etc.) are thus now shared between CTR and AEAD
  - See EverCrypt.CTR.Keys for a key-expansion routine shared between CTR and AEAD
- A more functional specification for the block function (no state in spec) and an idealization-friendly stream-cipher that `xor`s after the pseudo-random stream has been generated.
- Added a couple lemmas in Vale.Arch.BufferFriends and proof sketches

Some things to note.
- All interfaces use Lib.IntTypes.uint8 for secrets; I limited the use of Lib.Buffer types, as the indices were causing difficulties (`bytes` and `lbytes` don't play well together) and were causing further pain due to erased (ghost) algorithms at the agility level.

Looking forward.
- I have a couple "plumbing"-related things to do for extraction, but no blocker in sight.
- There are several proof equivalence for specs left, please see EverCrypt.CTR. In particular:
  - equivalence between Vale-AES and Hacl-AES (@Chris-Hawblitzel can you check my provisional lemma seems alright?)
  - connecting various libraries, e.g. blit vs repeati
- I'm moving EverCrypt interfaces to use Lib.IntTypes.uint8 as the secret type, but right now the interop wrappers are in terms of UInt8.t -- should we switch all wrappers to use secret integers? Right now, this accounts for the majority of instances of `friend Lib.IntTypes` (@R1kM for the wrapping layer between Vale and Low*)
- I need tests (I don't have anything right now). I could use some help for that, the test vectors are already present, it's only a matter of extending EverCrypt.Test (@tahina-pro as the test expert)
- For tests, it would probably be good to add an allocation function on the stack, i.e. EverCrypt.CTR.alloca (just like for AEAD)
- Missing `free` function
- We continue to suffer from discrepancy between endianness libraries; @s-zanella any update on the plan to reconcile FStar.Endianness and Lib.ByteSequence?

I'm pushing this before heading offline for the weekend... please comment here or on Slack. Thanks!


msprotz:
Getting the thing to compile was of course more work than expected because I realized that LowStar.Endianness wasn't working at all. This is now fixed, but you'll need a fresh kremlin to compile this. CC @ad-l who might be trying this out on this branch.

msprotz:
I've added tests, and the computed result is incorrect; unsurprisingly so, since there's `assume`s all over the place. I'll attempt to debug as best as I can.

msprotz:
Thanks to @R1kM my endianness mistake was spotted and tests are now passing for AES-CTR.

msprotz:
I intend to merge this today. CC @karthikbhargavan if you haven't seen this already.

PR #174, opened 2019-07-12
Randomness: use syscall when available, don't exit on error

victor-dumitrescu:


beurdouche:
Thanks !

lib/c/Lib_RandomBuffer_System.c
```diff
@@ -77,6 +89,5 @@ void hacl_aligned_free(void * ptr) {
 #endif // HACL_IS_WINDOWS
 
 void randombytes(uint8_t * x,uint32_t len) {
-  if (! (read_random_bytes(len, x)))
-    exit(1);
+  return (read_random_bytes(len, x));
```
msprotz: hey Victor, how come this is compiling? you're returning a value for a function marked `void`

victor-dumitrescu: Ah, you're right, thanks! I had a version of this file in the Tezos codebase, where I ran the tests, seems I somehow omitted to copy this change over, I'll push a fix.

PR #175, opened 2019-07-15
Provide tighter arguments for extraction to OCaml

jaybosamiya:
Previously, `--extract` was run with a `*` and subtracting out things that we wanted to eliminate from the graph. Now instead, we run it with a `-*` and explicitly opt-in for anything that is required for OCaml extraction. This should even further eliminate large chunks of the dependency graph, speeding up builds. This should hopefully be even more useful in partial builds.

Thanks @msprotz for suggesting this for speeding things up :)


msprotz:
This looks good to me. Please merge as soon as you get a green on linux.

PR #180, opened 2019-08-08
Refresh in-tree versions of curve25519

msprotz:


msprotz:
@wintersteiger can you take a quick look at the renaming of the old ecdh into the scalarmult and see if we need to change these three occurrences as well?

```
tests/benchmark/bench_curve25519.cpp:      { EverCrypt_Curve25519_ecdh(shared_secret, our_secret, their_public); }
tests/benchmark/bench_curve25519.cpp:      { Hacl_Curve25519_51_ecdh(shared_secret, our_secret, their_public); }
tests/benchmark/bench_curve25519.cpp:      { Hacl_Curve25519_64_ecdh(shared_secret, our_secret, their_public); }
```

msprotz:
the rest of it looks good to me. thanks Marina!

msprotz:
the only remark that I have is that some comments that explain the difference between ecdh and scalarmult would be good (one line for each would be enough)

wintersteiger:
Looks fine to me, the ecdh functions still have the same signatures, but I'd appreciate a comment to explain the differences too.

polubelova:
https://tools.ietf.org/html/draft-irtf-cfrg-curves-11#section-6.1

It says that we must check whether the result of the x25519 function is the all-zero value and abort if so.

wintersteiger:
Good point, thanks for letting us know!

PR #181, opened 2019-08-09
Stronger postconditions for AEAD: freeable, as_kv, g_is_null...

tahina-pro:
This pull request strengthens some postconditions for `EverCrypt.AEAD`. It allows `encrypt` and `decrypt` to:
* preserve `freeable` and `as_kv`
* return `InvalidKey` only if the state is `null`

This allows significant simplifications in the miTLS wrapper for AEAD, see project-everest/mitls-fstar@09b5b1af6370d00b77796159d68fad8fc27100cd and before.


tahina-pro:
Unfortunately, I have not been able to similarly strengthen the postcondition of `create_in`: when it returns `UnsupportedAlgorithm`, it is not enough to conclude that the algorithm is not a `supported_alg`. Indeed, creation can fail, for instance, if trying to use Vale with a wrong configuration:
https://github.com/project-everest/hacl-star/blob/542c702e4ba81c67b0bbde1f42dc6a3f549f080f/providers/evercrypt/fst/EverCrypt.AEAD.fst#L129-L144)

msprotz:
Once we have a fallback version of AES-GCM in C then we can reveal that unsupported algorithm implies alg <> { AES_GCM, CHACHAPOLY }

msprotz:
This looks good to me please merge as soon as you get a green @tahina-pro

PR #184, opened 2019-08-16
Inttypes for SHA. Do not merge

R1kM:


vale/code/arch/x64/interop/Vale.Wrapper.X64.Sha.fsti
```diff
@@ -6,17 +6,19 @@ module B = LowStar.Buffer
...
+let uint32_p = B.buffer uint32
 unfold
-let uint32_i = IB.ibuffer UInt32.t
+let uint32_i = IB.ibuffer uint32
 unfold
-let uint8_p = B.buffer UInt8.t
+let uint8_p = B.buffer uint8
 unfold
-let uint64 = UInt64.t
+let uint64 = uint_t U64 PUB
```
s-zanella: This is `PUB` while the types above (`uint32, uint8`) are `SEC`. Why the difference?

vale/code/crypto/sha/Vale.SHA.SHA_helpers.fst
```diff
@@ -174,12 +180,22 @@ let lemma_add_mod32_associates (x y z:int) :
...
 *)
 
-#reset-options "--using_facts_from ''"
 let lemma_add_mod_associates_U32 (x y z:UInt32.t) :
   Lemma (add_mod x (add_mod y z) == add_mod (add_mod x y) z)
   =
-  assert_norm (FStar.UInt.add_mod (v x) (FStar.UInt.add_mod (v y) (v z)) == FStar.UInt.add_mod (FStar.UInt.add_mod (v x) (v y)) (v z));
-  v_inj (add_mod x (add_mod y z)) (add_mod (add_mod x y) z)
+  let open Lib.IntTypes in
+  calc (==) {
```
s-zanella: This can be solved by `canon_semiring` given an appropriate ring instance (see https://github.com/FStarLang/FStar/blob/master/examples/semiring/CanonCommSemiring.Test.fst#L57)
It may be worth if this kind of reasoning is common.

vale/code/crypto/sha/Vale.SHA.SHA_helpers.fst
```diff
@@ -433,15 +453,23 @@ let lemma_sha256_msg1 (dst src:quad32) (t:counter) (block:block_w) : Lemma
...
+    a +. b +. c +. d == d +. c +. b +. a)
   =
-  lemma_add_mod_commutes (add_mod (add_mod d c) b) a;
-  assert (add_mod (add_mod (add_mod d c) b) a == add_mod a (add_mod (add_mod d c) b));
-  lemma_add_mod_commutes (add_mod d c) b;
-  assert (add_mod a (add_mod (add_mod d c) b) == add_mod a (add_mod b (add_mod d c)));
-  lemma_add_mod_commutes d c;
-  assert (add_mod a (add_mod b (add_mod d c)) == add_mod a (add_mod b (add_mod c d)));
-  ()
+  let open Lib.IntTypes in
```
s-zanella: Idem

providers/evercrypt/fst/EverCrypt.HKDF.fst
```diff
@@ -20,18 +20,21 @@ open Spec.Hash.Definitions
 open Spec.HKDF
 friend Spec.HKDF
 
+// TODO: Is this safe here? Can we safely break abstraction at this level?
+friend Lib.IntTypes
```
s-zanella: I believe this is safe. This is no different than untrusted C code calling into verified code using `Lib.IntTypes` and enforcing secrecy internally.

secure_api/merkle_tree/MerkleTree.Spec.fst
```diff
@@ -45,7 +44,7 @@ let hash2_raw src1 src2 =
 /// requires extending the definitions of tags and hashes. Our
 /// extended definition of hash justifies skipping any concrete
 /// computation on dummies.
-
+noeq
```
s-zanella: Because this is client code of `EverCrypt.Hash` it's fine to break the `Lib.IntTypes` abstraction so that this is an `eqtype`, avoiding the use of `prop` and `StrongExcludedMiddle`.

specs/Spec.HKDF.fsti
```diff
@@ -7,12 +7,12 @@ open Spec.Hash.Definitions
 val extract:
   a: hash_alg ->
   key: bytes{ Spec.HMAC.keysized a (Seq.length key) } ->
-  data: bytes{ Seq.length data + block_length a < max_input_length a } ->
-  Tot (lbytes (hash_length a))
+  data: bytes{ Seq.length data + block_length a <= Lib.IntTypes.max_size_t } ->
```
s-zanella: This is a stricter bound than before, because of the length restriction on `Lib.Buffer`

specs/Spec.HMAC.fsti
```diff
@@ -16,5 +17,5 @@ let keysized (a:hash_alg) (l:nat) =
 val hmac:
   a: hash_alg -> //18-07-09 can't mix refinements and erasure??
   key: bytes{ keysized a (Seq.length key) } ->
-  data: bytes{ Seq.length data + block_length a < max_input_length a } ->
+  data: bytes{ Seq.length data + block_length a <= max_size_t } ->
```
s-zanella: This is a stricter bound than before, because of the length restriction on `Lib.Buffer`

code/hash/Hacl.Hash.PadFinish.fst
```diff
@@ -238,46 +237,6 @@ let hash_word_len (a: hash_alg): n:U32.t { U32.v n = hash_word_length a } =
...
-        let dsti = B.sub dst U32.(8ul *^ i) 8ul in
-        C.Endianness.store64_be dsti s.(i);
-        let h2 = ST.get () in
-        Endianness.be_of_seq_uint64_base (S.slice (B.as_seq h2 s) (U32.v i) (U32.v i + 1)) (B.as_seq h2 dsti);
-        Endianness.be_of_seq_uint64_append (S.slice (B.as_seq h2 s) 0 (U32.v i))
-          (S.slice (B.as_seq h2 s) (U32.v i) (U32.v i + 1))
-  in
-  C.Loops.for 0ul (hash_word_len a) inv f
-
+  match a with
```
s-zanella: Well spotted that this is just `uints_to_bytes_{le,be}`.

code/ed25519/Hacl.Spec.BignumQ.Lemmas.fst
```diff
@@ -328,7 +330,7 @@ let lemma_div264_aux x =
     (wide_as_nat5 x / pow2 224) / pow2 40;
   (==) { lemma_div224 x }
     (v x4 + v x5 * pow2 56 + v x6 * pow2 112 + v x7 * pow2 168 + v x8 * pow2 224 + v x9 * pow2 280) / pow2 40;
-  (==) { }
+  (==) { _ by (canon_semiring int_cr) }
```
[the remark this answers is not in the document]
s-zanella: The fix for `canon_semiring` is in `FStar@master`: https://github.com/FStarLang/FStar/commit/e7af7de464bafb35389f344e43301c1028689dc2

code/hash/Hacl.Hash.Core.MD5.fst
```diff
@@ -104,24 +100,26 @@ val round_op_gen
...
     B.live h' abcd /\
     B.live h' x /\ // better to add this here also to ease chaining
-    B.as_seq h' abcd == Spec.round_op_gen f (B.as_seq h abcd) (E.seq_uint32_of_le 16 (B.as_seq h x)) (U32.v a) (U32.v b) (U32.v c) (U32.v d) (U32.v k) s (U32.v i)
+    B.as_seq h' abcd == Spec.round_op_gen f (B.as_seq h abcd) 
+			(Lib.ByteSequence.uints_from_bytes_le #_ #_ #16 (B.as_seq h x)) 
+			(U32.v a) (U32.v b) (U32.v c) (U32.v d) (U32.v k) s (U32.v i)
   ))
 
-#set-options "--max_fuel 0 --max_ifuel 0"
+#reset-options "--z3rlimit 200 --max_fuel 1 --max_ifuel 1"
```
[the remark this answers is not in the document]
R1kM: Good catch, I copied that from _dev, but the increased fuel and ifuel do not seem needed. I removed them.

code/hash/Hacl.Hash.Core.SHA2.fst
```diff
@@ -340,10 +338,10 @@ let shuffle a block hash ws =
 inline_for_extraction
 let zero (a: sha2_alg): word a =
   match a with
-  | SHA2_224 | SHA2_256 -> 0ul
-  | SHA2_384 | SHA2_512 -> 0UL
+  | SHA2_224 | SHA2_256 -> u32 0
+  | SHA2_384 | SHA2_512 -> u64 0
 
-#set-options "--max_fuel 0 --max_ifuel 0"
+#set-options "--z3rlimit 200 --max_fuel 2 --max_ifuel 2"
```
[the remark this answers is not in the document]
R1kM: Same as above, copied from _dev but the increase is unneeded here. I set fuel and ifuel to 0. It might become necessary when switching to _dev's precomputed ws though.

code/hash/Hacl.Hash.PadFinish.fst
```diff
@@ -190,9 +185,13 @@ let pad_3 (a: hash_alg) (len: len_t a) (dst: B.buffer U8.t):
       (**) assert FStar.Mul.(U128.v len * 8 < pow2 128);
       (**) assert FStar.Mul.(FStar.UInt.shift_left #128 (len_v a len) 3 < pow2 128);
       (**) assert FStar.Mul.(U128.(v (shift_left len 3ul)) = U128.v len * 8);
-      store_len a U128.(shift_left len 3ul) dst
+      (**) assert FStar.Mul.(U128.(v (shift_left len 3ul)) = U128.v len * 8);
```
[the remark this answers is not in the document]
R1kM: Duplication removed

code/hmac/Hacl.HMAC.fsti
```diff
@@ -27,17 +28,18 @@ let key_and_data_fits (a: hash_alg): Lemma
 
 inline_for_extraction
 let compute_st (a: hash_alg) =
-  tag: uint8_pl (hash_length a) ->
-  key: uint8_p{ keysized a (B.length key) /\ B.disjoint key tag } ->
+  tag: B.buffer uint8 {B.length tag == hash_length a} ->
+  key: B.buffer uint8{ keysized a (B.length key) /\ B.disjoint key tag } ->
   keylen: UInt32.t{ UInt32.v keylen = B.length key } ->
-  data: uint8_p{ B.length data + block_length a < pow2 32 } ->
+  data: B.buffer uint8{ B.length data + block_length a < pow2 32 - 1 } ->
```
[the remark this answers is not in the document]
R1kM: Unneeded typo, I removed the -1

providers/evercrypt/fst/EverCrypt.HKDF.fst
R1kM: Comment removed.

vale/code/arch/x64/interop/Vale.Wrapper.X64.Sha.fsti
R1kM: This uint64 will only be used to represent lengths of buffers, which will be passed as public arguments to Vale. I'd be fine with renaming it uint64_pub to avoid confusion in this fsti

secure_api/merkle_tree/MerkleTree.Spec.fst
msprotz: I suggested to Aymeric not trying to break the abstraction too much; since we're trying to settle on Lib.IntTypes everywhere, the fewer `friend`s we have, the better -- also, it would require adding an interface to this file

What solution would you envision? Something like Vale.Arch.BufferFriend to centralize the uses of friend? Or just friend in this module?

@R1kM needs to tell us whether the complexity of StrongExcludedMiddle is manageable or not

specs/Spec.HKDF.fsti
msprotz: but conceivably we don't need to restrict the length in spec, we can just demand at run-time an `uint32` for the length and then show that any `uint32` satisfies `< max_input_length a`, right?

secure_api/merkle_tree/MerkleTree.Spec.fst
R1kM: In the current status, I managed to successfully use Lib.IntTypes almost everywhere, and I think StrongExcludedMiddle is ok (it's only for a ghost security proof). The only problem left is MerkleTree.New.Low: the file is pretty big, and currently has no interface, which prevents us from using friend.
As an experiment, I tried to reveal IntTypes (through an assumed squash type equality), but verification failed line 680 in create_empty_mt, and I'm not completely sure why yet, it's probably fixable. 
If somebody knows exactly what needs to be exposed for clients to use Merkle Trees, a solution might be to have a small MerkleTree.New.Low.fsti, and have most of the existing code in a fst with friend Lib.IntTypes

specs/Spec.HKDF.fsti
R1kM: The reason for this bound can be found in Spec.HMAC.fst (which Spec.HKDF is relying on).
Spec.HMAC uses Lib.Sequence.concat, which requires that the combined length of the two sequences is smaller than max_size_t (i.e. 2^32). 
An easy fix for this is to use FStar.Seq instead. The current limitation is to agree with _dev which uses Lib.Sequence.

specs/Spec.MD5.fst
```diff
@@ -2,142 +2,136 @@ module Spec.MD5
...
-let h (x y z: U32.t) : Tot U32.t =
-  x `U32.logxor` y `U32.logxor` z
+let h (x y z: uint32) : Tot uint32 =
+  x ^. y ^. z
 
 inline_for_extraction
-let i (x y z: U32.t) : Tot U32.t =
-  y `U32.logxor` (x `U32.logor` U32.lognot z)
+let i (x y z: uint32) : Tot uint32 =
+  y ^. (x ^. ~. z)
```
s-zanella: It took me a while to spot this 1-symbol typo!

secure_api/merkle_tree/MerkleTree.Spec.fst
s-zanella: There are three places in `MerkleTree` where the secret integers abstraction in `Spec.Hash.Definitions.bytes_hash` needs to be broken:
1. When serializing hashes
2. When deserializing hashes
3. When comparing the computed root of a tree against a purported root

I ported `MerkleTree.New.Low` to use `Lib.IntTypes.uint8` for hashes, breaking the abstraction in the first two cases and declassifying the result of the comparison in the third:
1. This requires calling `Lib.RawIntTypes.u8_to_UInt8` to declassify individual bytes of hashes. The extracted C code remains almost the same (I just had to introduce a temporary variable).
2. Rather than calling `Lib.RawIntTypes.u8_from_UInt8` to endorse individual bytes of hashes, which could have a significant runtime cost if done with a loop, I wrote a separate `Lib.RawBuffer` module that befriends `Lib.IntTypes` and does this calling `LowStar.Buffer.blit` with no change in the generated C code.
3. Since comparing hashes in constant-time is cheap enough, I used `Lib.ByteBuffer.lbytes_eq` to do the comparison in secret hashes, declassifying the result.

This first two changes are completely transparent to clients. The third could have some minor impact, but I don't think this matters for e.g. CCF. 
As an alternative, I implemented a non constant-time comparison in `Lib.RawBuffer.lbytes_eq` that we could use instead to generate exactly the same C as before. To switch to it, uncomment [this line](https://github.com/project-everest/hacl-star/blob/414118ebc486949fb9e08b30077afa5e57159d00/secure_api/merkle_tree/MerkleTree.New.Low.fst#L3054) and comment out the one above.

Issue #179, opened 2019-08-06
Subtyping check fails for creating Sequence with length 8 and above


jamesbarne:
Hi, I am confused by why the following code fails when declaring a sequence of length 8 and above. 

```FStar

open Lib.IntTypes
open Lib.Sequence

let a = assert_norm(8 <= max_size_t) //this assertion goes through

let (list1:list nat{List.Tot.Base.length list1 ≤ max_size_t}) =  [0;0;0;0;0;0;0]
let seq1 : lseq nat 7 = of_list list1 //succeeds 

let (list2:list nat{List.Tot.Base.length list2 ≤ max_size_t}) =  [0;0;0;0;0;0;0;0]
let seq2 : lseq nat 8 = of_list list2 //fails here
```

s-zanella:
See https://github.com/FStarLang/FStar/issues/879

Rather than increasing `--max_fuel`, use `assert_norm` to compute the length of the list literal:

```FStar
#set-options "--max_fuel 0"

let (list2:list nat{List.Tot.Base.length list2 = 8}) = 
  let l = [0;0;0;0;0;0;0;0] in
  assert_norm (List.Tot.length l = 8);
  l
```

PR #196, opened 2019-10-17
Merging fstar-master into _dev

R1kM:
Cleanup left to do in _dev before merging back into fstar-master:
  - code/experimental should move to a branch. Old, non-experimental code should be removed
  - code/old vs code/code-with-previous-libraries
  - specs/old vs specs/experimental
  - specs/tests should verify, not just be lax-checked ( see for instance Spec.AES256.Test.fst for the style needed for verification)
  - Guard Lib.IntVector functions with Vale CPU features preconditions.
  - Clean obsolete Makefiles (Makefile.ocaml for instance). Local Makefiles should be written for different folders when needed (they will be similar to the one in code/chacha20)
  - Randomness should be audited, and use Vale CPUID and RANDR instructions for Hardware randomness
  - Tests/Benchmarks in code/* should move to tests/ (for instance code/ed25519/Hacl.Benchmark.Ed25519.c) and be under CI
  - Clean (remove?) snapshots/ now that we have dist/
  - Figure out what to do with Hacl.Utils


.docker/build/build.sh
```diff
@@ -185,8 +185,16 @@ function refresh_hints() {
...
+    # If build hints branch exists on remote, remove it
+    exists=$(git branch -r -l "origin/BuildHints-$CI_BRANCH")
+    if [ ! -z $exists ]; then
+        git push $remote :BuildHints-$CI_BRANCH
+    fi
+
     # Push.
-    git push $remote $CI_BRANCH
+    git checkout -b BuildHints-$CI_BRANCH
+    git push $remote BuildHints-$CI_BRANCH
```
s-zanella: This change will eventually make it into `dev` and will trigger a hints refresh for every commit.
I think this is ok, but it changes the status quo of refreshing hints automatically only in `fstar-master`.

prosecco:
I dislike the hint commits because they tend to be different on different machines (e.g. OS X vs linux).
And I still keep getting stuck with stale hints and needing to git clean, which is not ideal.
Still, for now this seems ok, I will yell if it keeps breaking my workflow.

Is there a standard #set-options prefix we can agree on for new/existing HACL* files based on our experience so far.
In particular, if we could find a set of “—using_facts_from” commands that work well for almost all our files, this is useful documentation for new contributors,
or even old contributors like me.

-Karthik

.docker/build/build.sh
s-zanella: @protz reminded that this only happens for scheduled builds (the `hacl-nightly`) target. 
So, unless we schedule builds of `dev` hints won't be refreshed automatically. If this is still a problem, we could easily modify the build logic to exclude `dev`.
This also means that we'll be able to trigger an automatic hint refresh in `dev` by manually queueing a `hacl-nightly` build from VSTS, which I think it's a nice feature.

I think this change should stand; sorry for the noise.

s-zanella:
> Is there a standard #set-options prefix we can agree on for new/existing HACL* files based on our experience so far.

I'd suggest `--max_fuel 0 --max_ifuel 0 --z3rlimit 20`. 
These options make people be conscious of fuel usage and learn when it's appropriate to increase each type of fuel using `#push-options`/`pop-options` in a minimal scope. It also slightly increases the default `--z3rlimit 5` which is too little for complex proofs.  Add `--query_stats` in interactive mode` to let people see all Z3 proof attempts and the fuel settings they use. The only _drawback_ is that it forces people to learn about `fuel`/`ifuel` and proofs by normalization early in their learning curve, but then we can avoid wasteful proofs (https://github.com/FStarLang/FStar/issues/879, https://github.com/project-everest/hacl-star/issues/179).

> In particular, if we could find a set of “—using_facts_from” commands that work well for almost all our files, this is useful documentation for new contributors, or even old contributors like me.

I don't think we can find a good global setting for `--using_facts_from`. After all, if there was one, we could modify the module we're excluding facts from or add an interface to it that hides these facts.

msprotz:
😱  it's merged!

PR #198, opened 2019-10-23
Add qTESLA implementation to global Makefile and under CI

kevinmkane:
This adds the qTESLA implementation to the top-level Makefile and therefore includes its build in CI. This only covers the qTESLA-I parameter set using 64-bit signed digit arithmetic. Other parameter sets' code is present but not currently covered by the build.


Makefile.common
```diff
@@ -60,6 +60,7 @@ BASE_FLAGS = \
   -minimal \
   -add-include '"kremlin/internal/types.h"' \
   -add-include '"kremlin/lowstar_endianness.h"' \
+  -add-include '"kremlin/fstar_int.h"' \
```
msprotz: can you use the new syntax (see kremlin/Changes.md) to insert this only for the QTesla file?

kevinmkane:
Closing; code merged into `_dev` instead under PR #201.

PR #201, opened 2019-10-31
Add qTESLA to experimental directory and under CI

kevinmkane:


s-zanella:
I queued a CI build to confirm we get a green, but this looks good to me.
Once merged into `_dev`, we should merge it into `dev`, which is the new staging branch (sorry for the moving target).

PR #204, opened 2019-11-08
Implementation of HPKE

R1kM:
This PR implements HPKE in the current version of the specs. It also uses @protz's tactic to automatically multiplex between different implementations of Curve, ChachaPoly, and SHA256 and SHA512.


Makefile.common
```diff
@@ -93,6 +93,10 @@ POLY_BUNDLE=-bundle 'Hacl.Poly1305_32=Hacl.Impl.Poly1305.Field32xN_32' \
   -bundle 'Hacl.Poly1305_128=Hacl.Impl.Poly1305.Field32xN_128' \
   -bundle 'Hacl.Poly1305_256=Hacl.Impl.Poly1305.Field32xN_256'
 NACLBOX_BUNDLE=-bundle Hacl.NaCl=Hacl.Impl.SecretBox,Hacl.Impl.Box
+# The only functions not marked as noextract should be in each of the Hacl.HPKE.{variants}
+# Each of these module should be extracted to a different file. Therefore, this variable
+# should remain empty, and overriden only when we do not want extraction of variants
+HPKE_BUNDLE=
```
msprotz: good comment, thanks

code/hpke/Hacl.Impl.HPKE.fst
```diff
@@ -0,0 +1,492 @@
...
+  -> label_key:lbuffer uint8 8ul
+  -> label_nonce:lbuffer uint8 10ul
+  -> tmp:lbuffer uint8 (10ul +. context_len)
+  -> ST unit
+       (requires fun h0 ->
+         live h0 pkR /\ live h0 zz /\ live h0 pkE /\
+         live h0 info /\ live h0 o_key /\ live h0 o_nonce /\
+         live h0 context /\ live h0 secret /\ live h0 pkI /\
+         live h0 psk /\ live h0 tmp /\ live h0 label_key /\ live h0 label_nonce /\
+         disjoint o_key o_nonce /\ disjoint secret psk /\ disjoint secret zz /\
```
msprotz: any chance you could use an `all_disjoint` predicate here for readability?

code/hpke/Hacl.Impl.HPKE.fst
```diff
@@ -0,0 +1,492 @@
...
+  -> context:lbuffer uint8 context_len
+  -> secret:lbuffer uint8 (nhash_length cs)
+  -> pkI:lbuffer uint8 (nsize_dh_public cs)
+  -> psk:lbuffer uint8 (nhash_length cs)
+  -> label_key:lbuffer uint8 8ul
+  -> label_nonce:lbuffer uint8 10ul
+  -> tmp:lbuffer uint8 (10ul +. context_len)
+  -> ST unit
+       (requires fun h0 ->
+         live h0 pkR /\ live h0 zz /\ live h0 pkE /\
```
msprotz: same here, `all_live`

code/hpke/Hacl.Impl.HPKE.fst
```diff
@@ -0,0 +1,492 @@
...
+         as_seq h0 label_nonce `Seq.equal` S.label_nonce /\
+         as_seq h0 psk `Seq.equal` S.default_psk cs /\
+         as_seq h0 pkI `Seq.equal` S.default_pkI cs)
+       (ensures fun h0 _ h1 -> modifies (loc o_nonce |+| loc o_key |+| loc tmp |+| loc context |+| loc secret) h0 h1 /\
+         (let keyIR, nonceIR = S.ks_derive cs S.Base (as_seq h0 pkR) (as_seq h0 zz) (as_seq h0 pkE) (as_seq h0 info) None None in
+         as_seq h1 o_key == keyIR /\ as_seq h1 o_nonce == nonceIR))
+
+#set-options "--z3rlimit 100"
+
+noextract inline_for_extraction
```
msprotz: you should not need to write `inline_for_extraction` yourself here -- the tactic will automatically add it, and this may speed up extraction time for this file (since the normalizer won't have as much work to do)

R1kM: Good catch, thanks. I removed the `inline_for_extraction` annotation from all functions with the inline_ tactic attribute

code/hpke/Hacl.Impl.HPKE.fst
R1kM: Unfortunately, since I'm using hacl's buffers and not F*'s buffers, the two interact badly together, there is a type mismatch between lists. I'd be happy to use bigops predicate if there is an easy way to solve this, but I think the root cause is https://github.com/FStarLang/FStar/issues/65

code/hpke/Hacl.Impl.HPKE.fst
R1kM: Same as above

code/hpke/Hacl.Impl.HPKE.fst
msprotz: can you use FStar.BigOps to define all_disjoint and all_live in Lib.Buffer? might as well?

R1kM: I could try, but my guess would be that we would have the same problem with buffer vs lbuffer (I already had type issues in this code when I had a lbuffer while expecting a buffer for instance, and that's without using it inside a list constructor)

R1kM:
Thanks for the comments. Some replies below:

- Regarding assume vals: I agree it would be nicer, but we unfortunately don't have (yet?) implementations for all algorithms we are multiplexing on (for instance Curve448). I'd be in favour of an assume val that is well-understood (happy to add a comment to it if needed) than a match with admits because we don't have full implementations yet.

- Regarding sequence associativity: It seems that the assertions indeed trigger the lemmas in the correct way. This proof does not seem to be flaky. The issues are more with functions that perform a lot of allocations on the stack (and hence need to be split into push/pop + allocations and actual body of the function).

- Regarding merging soon: I'd be in favour as well. @beurdouche reported that the spec would change one last time in the next few days before being stable for a while. I'm happy to wait until it changes, propagate changes to this branch and then merge into `dev`

code/hpke/Hacl.Impl.HPKE.fst
R1kM: Following discussion, I used `LowStar.Monotonic.Buffer.all_disjoint` for readability here. `all_live` seems trickier, but also less important since liveness predicates are not pairwise.

specs/Spec.Agile.HPKE.fst
```diff
@@ -16,7 +16,7 @@ let pow2_61 : _:unit{pow2 61 == 2305843009213693952} = assert_norm(pow2 61 == 23
 let pow2_35_less_than_pow2_61 : _:unit{pow2 32 * pow2 3 <= pow2 61 - 1} = assert_norm(pow2 32 * pow2 3 <= pow2 61 - 1)
 let pow2_35_less_than_pow2_125 : _:unit{pow2 32 * pow2 3 <= pow2 125 - 1} = assert_norm(pow2 32 * pow2 3 <= pow2 125 - 1)
 
-#set-options "--z3rlimit 200"
+#set-options "--z3rlimit 20 --max_fuel 0 --max_ifuel 1 --initial_ifuel 1"
```
s-zanella: You could use the new `--fuel` and `--ifuel` options here.

specs/Spec.Agile.HPKE.fst
```diff
@@ -104,11 +106,14 @@ let id_of_mode m =
 /// Constants sizes
 
 inline_for_extraction
-let size_aead_nonce (cs:ciphersuite): size_nat = AEAD.size_nonce (aead_of_cs cs)
+let size_aead_nonce (cs:ciphersuite): size_nat = 12
```
s-zanella: It'd be nice to document where this value come from.
This could be in the form of a lemma that proves that this is the nonce length for all supported AEAD algorithms.

code/hpke/Hacl.Impl.Instantiate.AEAD.fst
```diff
@@ -0,0 +1,42 @@
...
+open FStar.HyperStack
+open FStar.HyperStack.All
+
+open Lib.Buffer
+
+module SAE = Spec.Agile.AEAD
+
+friend Spec.Agile.AEAD
+
+#set-options "--z3rlimit 60 --max_fuel 0 --max_ifuel 0"
```
s-zanella: Use `--fuel 0`?

code/hpke/Hacl.Impl.HPKE.fst
```diff
@@ -0,0 +1,488 @@
...
+      upd output 3ul (u8 4); upd output 4ul (u8 4); upd output 5ul (u8 4)
+  | DH_P256,       AES128_GCM,        SHA2_256 ->
+      upd output 0ul (u8 5); upd output 1ul (u8 5); upd output 2ul (u8 5);
+      upd output 3ul (u8 5); upd output 4ul (u8 5); upd output 5ul (u8 5)
+  | DH_P256,       CHACHA20_POLY1305, SHA2_256 ->
+      upd output 0ul (u8 6); upd output 1ul (u8 6); upd output 2ul (u8 6);
+      upd output 3ul (u8 6); upd output 4ul (u8 6); upd output 5ul (u8 6)
+  | DH_Curve25519, CHACHA20_POLY1305, SHA2_512 ->
+      upd output 0ul (u8 7); upd output 1ul (u8 7); upd output 2ul (u8 7);
+      upd output 3ul (u8 7); upd output 4ul (u8 7); upd output 5ul (u8 7)
```
s-zanella: Move this above `val build_context_default` so that it doesn't split the declaration and the definition?

R1kM:
Thanks Santiago for the comments. I added an interface for the spec and addressed the smaller items as well.
Since the spec has not moved in the passed week on dev, I'm merging this now

PR #205, opened 2019-11-11
CPU preconditions on SSE and MOVBE

R1kM:


Chris-Hawblitzel:
Small comment: I'd prefer that the bodies of `eval_Paddd`, `eval_Palignr`, `eval_Shufpd`, and `eval_Pshufb` not be copied and pasted into the bodies of `eval_VPaddd`, etc.  This could be avoided be factoring the common code into separate functions or by further parameterizing `eval_Paddd`, etc.  Otherwise, the changes look good to me.

R1kM:
Good point, I just pushed an update factorizing the semantics. The base functions are now named eval_{instr}_raw

parno:
I read through the specs, and they look sane to me.  Thanks for putting this together!

PR #208, opened 2019-12-03
Add C snapshot to nightly builds when generating hints

franziskuskiefer:


.docker/build/build.sh
```diff
@@ -169,6 +169,11 @@ function refresh_hints() {
     # Add all the hints, even those not under version control
     find $hints_dir -iname '*.hints' -and -not -path '*/.*' -and -not -path '*/dependencies/*' | xargs git add
 
+    # Add new files from the C snapshot in dist.
+    # Remove files first that we don't want in the snapshot.
+    find dist \( -name "*.d" -o -name "*.cmi" -o -name "*.cmx" -o -name "*.ocaml" -o -name "*.cmxa" \) | xargs rm
+    git add dist
```
s-zanella: In FStar, the OCaml snapshot is refreshed by passing `"git ls-files src/ocaml-output/ | xargs git add"` as the second argument to `refresh_hints` (i.e. `$extra`). That is, we only refresh snapshot files that are already under version control. I suggest we do the same for uniformity and to be aware of newly generated files--the person whose change causes the new file to be generated will need to check it in explicitly.

On the other hand, snapshots are less stable, so we may want to automatically add untracked files as well. In this case, we need to add `*.dll` the toplevel `.gitignore`. The other patterns (e.g. `*.d`, `*.cmi`, etc.) are already ignored.

franziskuskiefer: I'm fine with both. Really depends on how stable the snapshots are supposed to be I think. 

Maybe we can start with only refreshing existing files as that's less controversial and change it later if people want to automate adding new files?

s-zanella: Yes, that's a good idea. We can discuss this on the Hacl meeting today.

To be clear, that would mean removing the lines above and passing
`"git ls-files dist | xargs git add"` instead of `true` to `refresh_hints_and_dist`

msprotz:
We chatted about this during the weekly meeting (which I forwarded to you).

We are happy to have everything auto-regenerated as follows:
- first, build the snapshot
- then, `rm -rf dist/*/*`
- then, run a full build
- then, `git add dist`

This requires updating the .gitignore to ignore dll files

It would be nice to warn if someone is pushing an out-of-date snapshot, in the way that is currently done for F* https://github.com/FStarLang/FStar/blob/master/.docker/build/build.sh#L336 -- we'd have to ignore the diff on INFO.txt since this is only mildly relevant.

.gitignore
```diff
@@ -24,3 +24,4 @@ code/old/experimental/aesgcm/aesgcm-c89
 Makefile.config
 .#*
 tests/benchmark/build
+*.dll
```
s-zanella: `*.dll` is already above.

```suggestion
```

PR #210, opened 2019-12-12
Improve error messages in Makefile concerning Vale

blipp:
- Removing lines 60+61 in `Makefile` because `VALE_HOME` is set inside `Makefile.common` and thus always exists.
- Introduce check if the directory `$VALE_HOME` exists because if it doesn't, an error message “$VALE_HOME/bin/vale.exe does not exist” could be confusing. I don't have a strong opinion on this one, if you want we can keep only the one checking for `vale.exe`, but I'd keep the hint to `get_vale.sh` in any case.
- Introduce newline to these error messages to improve readability.


msprotz:
Thanks! This looks good. Please merge as soon as you get a green.

PR #219, opened 2020-01-14
Nicer inline assembly generation for Curve procedures

R1kM:
This PR proposes changes to Vale to enable a nicer, manual formatting of generated inline assembly.

It should not be merged before finishing formatting all Vale fadd, fmul, … procedures.


Chris-Hawblitzel:
The inline printer has gotten pretty sophisticated, and the interaction with the inline assembler is subtle, so it would be good to add a bunch of (unverified) regression tests to the build that create some `code` values directly, print them, assemble them, run the code, and check the output.  This way, we can test all the corner cases and make sure nothing ever breaks them.  I think these tests could be done from a single `.fst` test file that creates a bunch of code values and prints them to a `.h` file, and a single `.c` file that calls the generated functions in the `.h` file.

Chris-Hawblitzel:
Minor style comment: `open FStar.Mul` was removed from a couple of files and `*` used for tuple types.  I'd prefer to keep `open FStar.Mul` and consistently use `&` for tuple types.

PR #221, opened 2020-01-15
Replace doc source with a link to GitHub

parno:
I propose putting a GitHub banner in place of the link to the rst files; the former seems more generally useful than the latter.


msprotz:
tested it and it looks good. thanks Bryan!

PR #224, opened 2020-01-22
Fix specs Makefile

msprotz:
Santiago, Karthik recently reported that the specs local Makefile no longer works. I tried to patch it up, but I thought I'd run the changes by you since you touched it recently.

Some relative paths seemed to need a fix, and somehow that shortest stem rule didn't seem to kick in for the .exe rule.

We should also enforce that this local Makefile uses a recent GNU make.

### DIFF
--- a/specs/Makefile
+++ b/specs/Makefile
@@ -11,14 +11,16 @@ FSTAR_ROOTS += $(wildcard tests/*.fst) $(wildcard alternative/*.fst) $(wildcard 
 .PHONY: all-tests
 all-tests: $(subst .,_,$(patsubst %.fst,test-ml-%,$(notdir $(wildcard tests/*.fst))))
 
-.PRECIOUS: specs/tests/%_AutoTest.ml
-specs/tests/%_AutoTest.ml:
+.PRECIOUS: tests/%_AutoTest.ml
+tests/%_AutoTest.ml:
 	echo "if not ($*.test ()) then (print_endline \"$* failed\"; exit 1)" > $@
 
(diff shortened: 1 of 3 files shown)


specs/Makefile
```diff
@@ -11,14 +11,16 @@ FSTAR_ROOTS += $(wildcard tests/*.fst) $(wildcard alternative/*.fst) $(wildcard
...
 
-.PRECIOUS: specs/tests/%_AutoTest.ml
-specs/tests/%_AutoTest.ml:
+.PRECIOUS: tests/%_AutoTest.ml
+tests/%_AutoTest.ml:
 	echo "if not ($*.test ()) then (print_endline \"$* failed\"; exit 1)" > $@
 
-.PRECIOUS: specs/tests/%.exe
-specs/tests/%.exe: $(ALL_CMX_FILES) specs/tests/%_AutoTest.ml
+%.exe:
```
msprotz: note: this cancels the previous pattern rule now defined in Makefile.local

s-zanella:
This looks good to me, but I'm curious how it was working for me before. 
Let me try locally before merging.

msprotz:
Thanks Santiago. I wonder if we should do something like this in the nightly build

```
for f in code/*/Makefile specs/Makefile lib/Makefile; do
  git clean -fdx && (cd $(dirname $f) && make -j)
done
```

thoughts?

s-zanella:
It looks expensive to do a `git clean -fdx` after each build. Leaving out the `git clean -fdx &&` part would be a cheap sanity test.

PR #227, opened 2020-01-29
Blake2 vectorized and verified

msprotz:
This Karthik and Benjamin's pull request. I'm opening it up since I fixed a bunch of build issues on this branch.

I'll let Karthik comment. The diff should be clean, except for the following files:

```
hints/Hacl.Poly1305.Field32xN.Lemmas2.fst.hints
hints/Hacl.Poly1305.Field32xN.Lemmas1.fst.hints
hints/Hacl.Spec.Poly1305.Field32xN.Lemmas.fst.hints
hints/Spec.Agile.AEAD.fst.hints 
```

these four files do not verify on this branch unless fresh hints are generated for them, on a linux machine, with z3rlimit_factor=4. What's up? Are they inherently fragile? Did something change in Lib.IntVector or something that caused some breakage?

I have added some notes for future users to describe how to submit a clean pull request, in `PULL.md`. @s-zanella I am particularly interested in your comments for this file -- I hope it helps future students and contributors minimize the amount of noise on pull requests.


lib/c/libintvector.h
```diff
@@ -71,13 +78,13 @@ typedef __m128i Lib_IntVector_Intrinsics_vec128;
   (_mm_shuffle_epi8(x0, _mm_set_epi8(13,12,15,14,9,8,11,10,5,4,7,6,1,0,3,2)))
 
 #define Lib_IntVector_Intrinsics_vec128_rotate_left32(x0,x1)	\
-  ((x1 == 8? Lib_IntVector_Intrinsics_vec128_rotate_left32_8(x0) : (x1 == 16? Lib_IntVector_Intrinsics_vec128_rotate_left32_16(x0) : _mm_xor_si128(_mm_slli_epi32(x0,x1),_mm_srli_epi32(x0,32-x1)))))
+  ((x1 == 8? Lib_IntVector_Intrinsics_vec128_rotate_left32_8(x0) : (x1 == 16? Lib_IntVector_Intrinsics_vec128_rotate_left32_16(x0) : _mm_xor_si128(_mm_slli_epi32(x0,x1),_mm_srli_epi32(x0,32-(x1))))))
```
msprotz: this would have been avoided with a `static inline` function

lib/c/libintvector.h
```diff
@@ -2,9 +2,16 @@
 #define __Vec_Intrin_H
 
 #include <sys/types.h>
-#include <wmmintrin.h>
+
+#define Lib_IntVector_Intrinsics_bit_mask64(x) -((x) & 1)
+
```
msprotz: at this stage we should contemplate having a `./configure` script that checks for things such as:
- presence of OCaml and whether to build bindings
- recent enough toolchain that has the avx2 headers, and bail otherwise

lib/c/libintvector.h
```diff
@@ -341,7 +365,181 @@ typedef __m256i Lib_IntVector_Intrinsics_vec256;
 #define Lib_IntVector_Intrinsics_vec256_interleave_high128(x1, x2) \
   (_mm256_permute2x128_si256(x1, x2, 0x31)) 
 
-#define Lib_IntVector_Intrinsics_bit_mask64(x) -((x) & 1)
+#elif defined(__aarch64__) || defined(_M_ARM64) || defined(__arm__) || defined(_M_ARM)
```
msprotz: this is not tested under CI; were you able to confirm that it builds on ARM?

prosecco:
> How feasible is it to switch the intvector macros to static inline functions? This would i) avoid silly mistakes like the ones you discovered with missing parentheses, and ii) provide some minimal amount of type-checking, as kremlin would emit static inline declarations for what the F* signatures claim, and then the C compiler would verify that you are providing functions with at least the same signature
“static inline” does not work for intrinsics, since some of them expect “immediate” arguments.
> is there any hope of testing the libintvector functions against some executable semantics of any kind for those vector instructions?
We should test them, yes. I was hoping we could test libintvector.h in C, and independently prove Lib.IntVector correct against a low-level semantics of the intrinsics defined in Lib.IntVector.Intrinsics
> can you remove the definition of BLAKE2_BUNDLE in code/blake2/Makefile? the one from Makefile.common should be enough
I didn’t know which ones were needed. Can do.

msprotz:
More comments:

- build fails on OSX: unlike what I wrote in my email, this is an actual error (and not some user error of mine):

```
Hacl_Blake2b_256.c:265:11: error: argument to '__builtin_ia32_permdi256' must be a constant integer
    v12 = Lib_IntVector_Intrinsics_vec256_shuffle64(v03, n0, n13, n23, n33);
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./libintvector.h:274:4: note: expanded from macro 'Lib_IntVector_Intrinsics_vec256_shuffle64'
  (_mm256_permute4x64_epi64(x0, _MM_SHUFFLE(x4,x3,x2,x1)))
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/11.0.0/include/avx2intrin.h:838:12: note: expanded from macro '_mm256_permute4x64_epi64'
  (__m256i)__builtin_ia32_permdi256((__v4di)(__m256i)(V), (int)(M))
           ^                                              ~~~~~~~~
```

- CI fails on linux because Spec.Agile.AEAD still refuses to go through -- I'll try to push more hints or tweak rlimits

I'll push a commit with the regenerated C code as soon as I have a successful build.

msprotz:
This also does not compile on my Raspberry Pi:
```
pi@pitagyros:/tmp/hacl-star/dist/portable-gcc-compatible $ make Hacl_Poly1305_32.o
cc -fPIC -I /include -I. -I ../kremlin/include -I ../kremlin/kremlib/dist/minimal -Wall -Wextra -Werror -std=c11 -Wno-unused-variable -Wno-unknown-warning-option -Wno-unused-but-set-variable -Wno-unused-p
arameter -Wno-infinite-recursion -g -fwrapv -D_BSD_SOURCE -D_DEFAULT_SOURCE -fPIC -Wno-unused -Wno-parentheses -Wno-deprecated-declarations -g -std=gnu11 -O3   -c -o Hacl_Poly1305_32.o Hacl_Poly1305_32.c
In file included from Hacl_Poly1305_32.h:25,
                 from Hacl_Poly1305_32.c:25:
libintvector.h: In function ‘Lib_IntVector_Intrinsics_vec128_load64s’:
libintvector.h:512:3: note: use -flax-vector-conversions to permit conversions between vectors with differing element types or numbers of subparts
   return vld1q_u64(a);
   ^~~~~~
libintvector.h:512:10: error: incompatible types when returning type ‘uint64x2_t’ but ‘Lib_IntVector_Intrinsics_vec128’ was expected
   return vld1q_u64(a);
          ^~~~~~~~~~~~
libintvector.h: In function ‘Lib_IntVector_Intrinsics_vec128_load32s’:
libintvector.h:517:20: error: passing argument 1 of ‘vld1q_u32’ from incompatible pointer type [-Werror=incompatible-pointer-types]
   return vld1q_u32(a);
                    ^
In file included from libintvector.h:369,
                 from Hacl_Poly1305_32.h:25,
                 from Hacl_Poly1305_32.c:25:
/usr/lib/gcc/arm-linux-gnueabihf/8/include/arm_neon.h:10413:29: note: expected ‘const uint32_t *’ {aka ‘const unsigned int *’} but argument is of type ‘uint64_t *’ {aka ‘long long unsigned int *’}
 vld1q_u32 (const uint32_t * __a)
            ~~~~~~~~~~~~~~~~~^~~
In file included from Hacl_Poly1305_32.h:25,
                 from Hacl_Poly1305_32.c:25:
libintvector.h: In function ‘Lib_IntVector_Intrinsics_vec128_interleave_low64’:
libintvector.h:527:35: error: incompatible type for argument 1 of ‘vgetq_lane_u64’
   uint64_t a[2] = {vgetq_lane_u64(x1,0),vgetq_lane_u64(x2,0)};
                                   ^~
```
note that this is a file which does not use vectorized stuff.

lib/Lib.Buffer.fsti
```diff
@@ -601,6 +601,36 @@ val loop:
...
+  -> a_spec:Type
+  -> refl:(mem -> GTot a_spec)
+  -> footprint:Ghost.erased B.loc
+  -> spec:(mem -> GTot (i:size_nat{i < v n} -> a_spec -> a_spec))
+  -> impl:(i:size_t{v i < v n} -> Stack unit
+      (requires loop_refl_inv h0 n a_spec refl footprint spec (v i))
+      (ensures  fun _ _ h -> loop_refl_inv h0 n a_spec refl footprint spec (v i + 1) h)) ->
+  Stack unit
+    (requires fun h -> h0 == h)
+    (ensures  fun h0 _ h1 -> loop_refl_inv h0 n a_spec refl footprint spec (v n) h1)
```
s-zanella: This is a straightforward special case of `Lib.Buffer.loop`. 
Does using it makes type checking easier in practice, or is it to be able to write e.g. `let refl h = as_seq h b` instead of `let refl h i = as_seq h  b`?

prosecco:
> This is a straightforward special case of Lib.Buffer.loop.
> Does using it makes type checking easier in practice, or is it to be able to write e.g. let refl h = as_seq h b instead of let refl h i = as_seq h b
> 
It is just a little bit simpler, yes, when passing in functions, and it happens to be perhaps the most common case of loop.

-Karthik

prosecco:
I fixed these errors, but of course we need a completely new set of Makefile rules for arm.
Maybe we should just define a new Makefile.arm rather than put conditionals everywhere?

msprotz:
Until we have CI on ARM, let's just rely on cursory testing, for instance:
```
make Hacl_Blake2s32.o
```
ok?

I don't think it's that much of a deal to fix the Makefile (mostly filter out a few 256-bit variants, along with no .S files since they're X64 only), but more importantly, the Everest multiplexing currently assumes that 128-bit is equivalent to X64 && avx, which is not true.

So since it's still a little bit of work to do, let's wait and do everything at once, unless of course ARM is a priority and you have someone who can set up ARM CI for us (i.e. just compiling the C code).

karthikbhargavan:
I will make local ARM makefiles for testing chacha/poly/blake.
We can then try to see how to generalize the global makefiles.

msprotz:
```
/usr/lib/gcc/arm-linux-gnueabihf/8/include/arm_neon.h:6324:28: note: expected ‘uint64x2_t’ but argument is of type ‘Lib_IntVector_Intrinsics_vec128’
 vgetq_lane_u64 (uint64x2_t __a, const int __b)
                 ~~~~~~~~~~~^~~
In file included from Hacl_Blake2s_32.h:25,
                 from Hacl_Blake2s_32.c:25:
libintvector.h:528:10: error: incompatible types when returning type ‘uint64x2_t’ but ‘Lib_IntVector_Intrinsics_vec128’ was expected
   return vld1q_u64(a);
          ^~~~~~~~~~~~
libintvector.h: In function ‘Lib_IntVector_Intrinsics_vec128_interleave_high64’:
libintvector.h:532:35: error: incompatible type for argument 1 of ‘vgetq_lane_u64’
   uint64_t a[2] = {vgetq_lane_u64(x1,1),vgetq_lane_u64(x2,1)};
                                   ^~
In file included from libintvector.h:369,
                 from Hacl_Blake2s_32.h:25,
                 from Hacl_Blake2s_32.c:25:
/usr/lib/gcc/arm-linux-gnueabihf/8/include/arm_neon.h:6324:28: note: expected ‘uint64x2_t’ but argument is of type ‘Lib_IntVector_Intrinsics_vec128’
 vgetq_lane_u64 (uint64x2_t __a, const int __b)
                 ~~~~~~~~~~~^~~
In file included from Hacl_Blake2s_32.h:25,
                 from Hacl_Blake2s_32.c:25:
libintvector.h:532:56: error: incompatible type for argument 1 of ‘vgetq_lane_u64’
   uint64_t a[2] = {vgetq_lane_u64(x1,1),vgetq_lane_u64(x2,1)};
                                                        ^~
In file included from libintvector.h:369,
                 from Hacl_Blake2s_32.h:25,
                 from Hacl_Blake2s_32.c:25:
/usr/lib/gcc/arm-linux-gnueabihf/8/include/arm_neon.h:6324:28: note: expected ‘uint64x2_t’ but argument is of type ‘Lib_IntVector_Intrinsics_vec128’
 vgetq_lane_u64 (uint64x2_t __a, const int __b)
                 ~~~~~~~~~~~^~~
In file included from Hacl_Blake2s_32.h:25,
                 from Hacl_Blake2s_32.c:25:
libintvector.h:533:10: error: incompatible types when returning type ‘uint64x2_t’ but ‘Lib_IntVector_Intrinsics_vec128’ was expected
   return vld1q_u64(a);
```
still getting this after git pull in the dist/ directory

msprotz:
(that's on ARM)

prosecco:
The dist is not refreshed, looks like. (see https://github.com/project-everest/hacl-star/blob/karthik_blake2/lib/c/libintvector.h <https://github.com/project-everest/hacl-star/blob/karthik_blake2/lib/c/libintvector.h>_

msprotz:
I updated this branch, and cherry-picked the fixes that had been pushed on `karthik_blake2`. This now compiles on my Raspberry Pi.

msprotz:
@karthikbhargavan can you:
- remove the definition of BLAKE2_BUNDLE in code/blake2/Makefile
- move the test to tests/ so that it's picked up by CI
- remove the `extern` declarations in the test file? this is very unsafe and might cause ABI mismatches or linking errors down the line, possibly without being caught; please do `#include "Hacl_Blake2XXX.h"` -- since your bundle is the same for local & global Makefile, the test file will work in a local and global setting
- fix your test: `main` needs to return a value, and the return value must correspond to the test outcome so that any failure can be caught on CI
- I don't see any use of --max_fuel --max_ifuel; have you considered using them in your code? I would suspect significant proof improvements, especially with all those global lists in your context

prosecco:
I think I did all these. Still some bugfixes needed for ARM.

msprotz:
merging this per discussion with @karthikbhargavan ... we will fix the following later:
- ARM build (@karthikbhargavan)
- blake2 added to Spec.Agile and multiplexed with EverCrypt (@R1kM)
- possibly CI for ARM

PR #229, opened 2020-01-30
Lib.Sequence.Lemmas

polubelova:
- Simplify preconditions of the lemma_map_blocks_vec and lemma_repeat_blocks_multi_vec lemmas (it takes a block as an input, not an entire input sequence)
- Remove restrictions on `w` in the lemma_repeati_vec and lemma_repeat_blocks_multi_vec lemmas
- Add lemma_map_blocks_ctr_vec which is used in the CTR mode
- Update spec equivalence proofs for chacha20 and poly1305


msprotz:
@polubelova would you still like to merge this?

polubelova:
I'll close this PR and will do another one when `poly1305_avx512` and `chacha20_avx512` are done. Thanks!

PR #231, opened 2020-02-04
Kyber (work in progress)

beurdouche:
This is work from Guillaume that needs to be updated, so that we don't loose it.


prosecco:
Is this based on Guillaume’s work?

beurdouche:
Yes, it is. It was sitting in a dead branch, so the intent is to preserve his commit history
and merge everything in _dev in experimental/kyber…

If you feel like doing it, don’t hesitate : )

PR #239, opened 2020-02-19
Remove custom modifies lemmas; saner rlimits in a P256 proof

msprotz:
FYI... @Frosne @beurdouche @s-zanella


msprotz:
The whole file goes through with rlimit 150 on my linux box. I didn't try to reduce the rlimit further but it might be that an even lower value works.

PR #240, opened 2020-02-19
Add helpers in Lib.Buffer to ease the transition to const buffers

msprotz:
This also removes internal code duplication... FYI @s-zanella you might be interested


msprotz:
On branch `protz_const` for now, there were some regressions in extraction... will reopen once ready.

PR #243, opened 2020-02-24
Make the Curve25519 call-graph indexed over a predicate

msprotz:
Instead of being parameterized over a field representation, the generic
code is now parameterized over a field representation and a global
invariant.

```
type field_spec = M51 | M64
let index = field_spec & Type
```

Therefore, at instantiation-time, we now do:

```
let scalarmult_vale64 = mk_scalarmult (M64, Vale.CPU_Features.(adx /\ bmi2))
let scalarmult_hacl64 = mk_scalarmult (M64, True)
let scalarmult_hacl51 = mk_scalarmult (M51, True)
```

where `mk_scalarmult` is the higher-order, tactic-rewritten function
that generates a specialized instance of `scalarmult` for:
- a given field representation
- a global invariant that is carried throughout the whole call-graph
- a given set of functions that require the global invariant and
  implement the core operations for that field representation.

M64 has two implementations; M51 only has one. This is modular: further
implementations possibly requiring new fancy invariants may be added at
any time without having to touch the curve implementation.

This fixes a previous limitation of the curve code, wherein the hacl64
version would still inherit the Vale precondition because the code had
no way of referring to an abstract precondition that would be
instantiated depending on what the core operations need.

The implementation works and has the nice property that it didn't
require modifying the tactic. However, the result is somewhat verbose,
and achieving proper reduction behavior was a struggle. Indeed, having
the index be a pair now requires us to be very careful when destructing
and/or constructing.

1) In types, intermediary let-bindings where the left-hand side is a
pattern block reduction and prevent calls from type-checking. Therefore,
we do:

```
inline_for_extraction
let f_t (i: index) =
  let s = fst i in
  let p = snd i in
  ... ->
  felem s ->
  ... ->
  Stack (fun _ -> requires p /\ ...) (...)
```

instead of `let s, p = i`.

2) The default `fst` projector is not marked as `inline_for_extraction`,
so we shadow it and only use our version marked as
inline_for_extraction. In code, we use `[@inline_let]` to define `s`.

3) The presence of a pair seems to block reduction. This does not work:

```
inline_for_extraction
let i = M51, True

val f: #i:index -> f_t i
```

but this does:

```
val f: #i:index -> f_t (M51, True)
```

4) The index being a pair seems to interact poorly with implicit
argument inference and many call-sites had to be annotated explicitly.

5) Unless the predicate is annotated at Type0, universe binding problems
arise, as the tactic assumes that the code being rewritten is not
universe-polymorphic.

--------------------------------------------------------------------------------

There may be other solutions; for instance, one might imagine a second
index that is not touched by the tactic, e.g:

```
val g_t (s: field_spec) =
  #p:Type ->
  ... ->
  felem s ->
  ... ->
  Stack (fun _ -> requires p /\ ...) ...

val g: #s:field_spec -> g_t s
```

This is only marginally more verbose than the solution in this commit;
it might also reduce struggles when it comes to dealing with pairs.
However, the implicit argument will have to be explicitly provided,
unless...

... unless we write a tactic to infer `p` by just doing a lookup in the
context and finding a `p` that is an argument to the current function,
and has type `Type`. But this is also high-tech and we may have blown
the complexity budget for that.


msprotz:
I'm leaving this out for discussion with @s-zanella and @R1kM ... I'm not super satisfied, but this may be the only reasonable solution for now. The alternative I'm outlining might be marginally better, I don't know... thoughts welcome.

s-zanella:
Did you consider internalizing the invariant in the definition of the index type?
```
type implementation = | Hacl | Vale
type index = M51 of implementation | M64 of implementation | M64 of implementation
```
Or even
```
type index = M51_Hacl | M64_Hacl | M64_Vale
```
This only requires changing the definition of `limb`, `nlimb`, `wide`, `nwide` to pattern match on the new type.

That would be less verbose and avoid the problems with pairs. Am I missing something obvious?

msprotz:
Yes. This was ruled out early on, because:
- it would introduce conversions when calling spec, e.g. `== Spec.Curve25519.ecdh (spec_alg_of_impl_alg a) ...`
- it would introduce implementation details -- so far, Hacl.Impl.Curve25519 is oblivious to whether there is one or many implementations for a given representation
- adding an implementation would require touching Hacl.Impl, i.e. we would lose modularity

I really like the property that Hacl.Impl.Curve25519 is completely core-field-implementation-agnostic and this is what I'm hoping to preserve. But the main pain stems from the use of a pair in the index.

Maybe doing: `#s:field_spec -> #p:Type -> ... requires p` (like I suggest near the end) would be much simpler and the same amount of complexity as we have *now*. I just haven't had a chance to experiment with it -- we'd also need a little bit of tactics to make sure the implicit argument is instantiated sensibly.

s-zanella:
Got it. Thanks for the detailed explanation.

Then I don't see other solution than modifying the tactic in some way.
The solution you described in a comment of parametrizing the call graph by a global invariant looked promising, but it requires the tactic to rewrite and fiddle with types.

If this pattern will be common enough, it may be worth to modify the tactic to take an invariant as a second parameter (or add a new one, to avoid modifying the existing algorithms that don't use this feature) .

msprotz:
I can try that. Thanks for the in-depth look and helpful advice!

msprotz:
commit 6d73fabf280012a86ff5c3a1626643179ffd76f9 (master)
Author: Jonathan Protzenko <jonathan.protzenko@gmail.com>
Date:   Mon Mar 2 13:15:29 2020 -0800

    Rewrite higher-order tactic to insert preconditions.
    
    This follows Santiago's suggestion in https://github.com/project-everest/hacl-star/pull/243
    
    Essentially, the tactic, upon encountering a node in the call-graph,
    unfolds its type (because most declarations are of the form `let f: #i:index -> f_t i`),
    then pushes a precondition `p` in its requires clause, which then gets
    threaded all throughout the call-graph.
    
    This generates code that type-checks slower than before, owing to the
    need to unfold and refold types; furthermore, situations where a type
    could be left as-is because no extra parameters were needed now gets
    rewritten systematically.

commit 447128b0dfc60fad30bedf3373e37ebaeef63acc (HEAD -> protz_pre, origin/protz_pre)
Author: Jonathan Protzenko <jonathan.protzenko@gmail.com>
Date:   Mon Mar 2 14:03:11 2020 -0800

    Adapt the code to the updated tactic.
    
    There are some unpleasant side-effects; for instance, type abbreviations
    normally wouldn't need to mention the extra precondition that gets
    inserted by the tactic; however, in the interest of sharing types
    between precondition-unaware, generic code and specialized instances of
    base combinators that *do* require a precondition, I added the parameter
    to a few types.
    
    Other than that, works fine.

msprotz:
@protz
Update other algorithms too
2e40c92

This updated tactic seems to yield code that type-checks slower,
possibly owing to type abbrevations being less shared between various
variants of the function.

msprotz:
So this passes CI, but the code seems to type-check less efficiently. (I don't have hard evidence, just based on what I saw on my laptop.)

This is also quite a fair amount of extra complexity, but better than the previous iteration of this patch (which I left in the commit history via `git revert`).

If Curve25519 was the only algorithm for which we need this, I'd be inclined to do it manually; however, we might need it for other algorithms in which case it might turn out to be convenient. But for the purposes of sharing type abbreviations, the user is somehow required to anticipate on the insertion of the precondition by the tactic, which is weird.

s-zanella:
This looks great and much cleaner than before!

My original suggestion was to keep both versions of the tactic and use the simpler version when the invariant is trivial. That way we pay for the extra typechecking cost only when it's needed. 
I realize maintaining two versions of the tactic is a burden (unless modifying the tactic so that it doesn't rewrite types to add trivial invariants is easy). If typechecking is not significantly more expensive, I think that always carrying an invariant is reasonable.

msprotz:
It would've been some extra complexity to add a mode where no predicate is inserted, and in a sense there should always be a predicate (even though this is not enforced for the HACL vectorized instructions). So, I left it in there, even though it forces a rewriting of some leaf nodes that would not be rewritten otherwise.

The slowdown is not noticeable on CI so I'm inclined to merge this, pending @R1kM's approval once the ICFP deadline is over.

R1kM:
The PR looks good to me, feel free to merge.

How hard would it be to extend the predicate p to operate on HS.mem instead of being a pure predicate? I don't think any of the existing code requires it, but HPKE would: https://github.com/project-everest/hacl-star/blob/afromher_hpke/code/hpke/Hacl.Impl.HPKE.fsti#L35 . At the moment, we only have one implementation of P256, so adding this precondition on any instantiation with P256 is not restrictive, but we might have similar cases in the future.

msprotz:
if hpke needs it then of course we should make it happen; I think it's pretty trivial to do

I don't understand, however, from your example, where `h` is needed

R1kM:
On lines 36 and 37, when the curve used is P256, we require bounds on `nat_from_bytes_le (as_seq h0 skE)`
At the moment, we only have one implementation of P256, so it is not really restrictive. Nevertheless,
in the future, I could imagine a different implementation that does not require this precondition, or that internalizes it.

msprotz:
This is trickier because now you need to receive the heap and possibly the arguments, too -- which makes the precondition really not generic at all and means that you can no longer have a single precondition threaded for all the call-graph.

s-zanella:
These preconditions aren't necessary. I removed them (see https://github.com/project-everest/hacl-star/commit/afe04c947f55a81f48af6c4e99d4b0bd5abee488). @R1kM please update HPKE when you can.

I discussed this with @Frosne who had concerns about eliminating the requirement that the scalar is in the [1..p-1] interval. I argued that the scalar is a trusted input and doesn't need to be validated, We discussed the possibility of validating it dynamically. She also suggested adding alternative functions with the pre-conditions, but the C code would be identical. I'm curious to hear your preferences.

Frosne:
@beurdouche @blipp

prosecco:
> I discussed this with @Frosne who had concerns about eliminating the requirement that the scalar is in the [1..p-1] interval. I argued that the scalar is a trusted input and doesn't need to be validated, We discussed the possibility of validating it dynamically. She also suggested adding alternative functions with the pre-conditions, but the C code would be identical. I'm curious to hear your preferences.

I think it is important for is to implement the dynamic check for P256 before removing the precondition; we have no idea what will happen if a scalar outside the prime field is provided to the code. Can we add a PR to fix this please? Unilaterally removing the precondition from master seems dangerous :(

s-zanella:
We know exactly what happens because it's specified here: https://github.com/project-everest/hacl-star/blob/master/specs/ecdsap256sha2/Spec.DH.fst.
I don't believe there is anything wrong with the specification, but if we keep the pre-conditions in the spec, they also would need to be added in the specification.

beurdouche:
> We know exactly what happens because it's specified here: https://github.com/project-everest/hacl-star/blob/master/specs/ecdsap256sha2/Spec.DH.fst.
> I don't believe there is anything wrong with the specification, but if we keep the pre-conditions in the code, they also would need to be added in the specification.

Ok, makes sense.. : ) We'll discuss more in Paris as we wanted a defensive API for a while and we'll get back...

s-zanella:
> Unilaterally removing the precondition from master seems dangerous :(
I don't see how this could be dangerous for C clients. The extracted code is identical.

msprotz:
What I'm getting from this discussion is:
- threading a restrictive precondition throughout spec & implementation is redundant, because the outermost entry point that is exposed to F* clients already properly specifies what happens if the point is out of range
- @beurdouche would like a defensive C API; maybe you would be interested in sketching Hacl.Defensive.ECDSA?
- this means that my pull request is good in its current and I can merge it.

Let's chat more about this this week.

msprotz:
> I will double check the precondition story with @Frosne, until please don't merge until then.

@beurdouche this is exclusively about the tactic, which changes nothing to the current state of the code -- why block this pull request?

Frosne:
Hacl.Defensive.ECDSA, I think it's more about me.

beurdouche:
> @beurdouche <https://github.com/beurdouche> this is exclusively about the tactic, which changes nothing to the current state of the code -- why block this pull request?
> 
I didn’t go back to the top of the discussion, that’s why I didn’t realize, but  - please - I am
not the one that started to fork the discussion on something else within this thread…

This discussion should move to its own issue...

R1kM:
Thanks @s-zanella . I merged from `master` and I'm currently removing this precondition from the HPKE code on my branch. Changes to add predicates operating on memory do not seem needed at the moment, nor in the near future.
I'd be in favour of merging this PR: It is clean and self-contained to the tactic changes (and propagating it to actual code), and it'd be nice to have this code under CI.

PR #245, opened 2020-02-25
Son blake2s 256

Kachoc:
Add blake2s 256


msprotz:
Thanks Son, this looks good to me. Please merge as soon as you get a green!

karthikbhargavan:
Actually I am not sure this version is needed. It is identical to the 128
version.

PR #247, opened 2020-03-02
Add the prefetch-elimination transform

jaybosamiya:
Due to some experiments that demonstrate some pretty nice improvements due to the addition of some prefetches, it seemed useful to introduce a prefetch-elimination transform, that demonstrates that code with and without prefetches is exactly the same.


msprotz:
requesting another review from Aymeric since we need two reviews to merge

jaybosamiya:
Good question. The pattern could indeed _technically_ be removed. However, notice that it is set to be a _very_ restrictive pattern, which only shows up in the internals of how the transformers work. In particular, its pattern exactly the same as its post condition, meaning it would otherwise be required manually whenever you need that particular postcondition. Since this particular file is only an internal file, and never meant to be `open`ed in any file other than the implementation of the highest-level API for the transformers, that too puts an even stronger restriction of when the pattern can get triggered. Thus overall, I believe the pattern is safe to keep, and behaves only as a user convenience that does not have any negative effects to proof performance elsewhere due to unexpected triggering.

PR #251, opened 2020-03-11
Preconditions for DH on P-256

s-zanella:
Commit afe04c947f55a81f48af6c4e99d4b0bd5abee488 weakens the preconditions of Diffie-Hellman functions on P-256. 

I didn't think that change would spark such a heated debate, because:
1. the extracted code is identical;
2. it's backwards compatible: no changes to F* or C clients needed.

The intention of the PR is not to revert the commit, but to open a discussion about what we think are required preconditions and checks in the DH API. More broadly, we might use this space to discuss how defensive we want the API to be. Some points for discussion:

1. Should we add checks and preconditions that aren't required but make the API harder to misuse?
2. Should these checks and pre-conditions be included in the specification as well?

For reference, the original discussion started here: https://github.com/project-everest/hacl-star/pull/243#issuecomment-597048137


s-zanella:
My rationale for the original change besides the two points above was:

- The specification allows the change.
- After the change, the implementation behavior remains fully specified and identical as before.
- With a few exceptions (e.g. defensive F*-level APIs) redundant preconditions in implementations should be avoided.
- A DH private key is a trusted input and does not need to be validated. It can be validated to make the API harder to misuse and avoid pitfalls, but this is not required.

code/ecdsap256sha2/Hacl.Impl.P256.DH.fsti
```diff
@@ -34,7 +36,11 @@ val ecp256dh_r:
   -> Stack uint64
     (requires fun h ->
       live h result /\ live h pubKey /\ live h scalar /\
-      disjoint result pubKey /\ disjoint result scalar)
+      disjoint result pubKey /\ disjoint result scalar /\
+      nat_from_bytes_le (as_seq h (gsub result (size 0) (size 32))) == 0 /\
+      nat_from_bytes_le (as_seq h (gsub result (size 32) (size 32))) == 0 /\
```
s-zanella: These two preconditions are clearly redundant and do not add any security.

Frosne:
IMHO, it's important to have at least a check for point at infinity at the end (that makes the API failing). Not sure that it would be nice to give it as the result otherwise

s-zanella:
I agree with @Frosne that the check for point at infinity is required (the original commit didn't remove this check).

R1kM:
I'm no expert in cryptography, so I'm not sure I understand perfectly what happens if the scalar is out of bounds (even if the mathematical behaviour is specified in Spec.DH). 
Is it something that should never be allowed because it weakens the cryptographic guarantees? If so, shouldn't it be reflected in the spec? (by returning an error when the scalar is out of bounds)

Frosne:
Just to mention: 
https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf 
5.6.1.2 ECC Key-Pair Generation - Given valid domain parameters, each valid private key d is an integer that is randomly selected in the interval [1, n−1]

That's where the precondition was coming from :)

s-zanella:
I don't believe that passing a scalar out of range to `ecp256dh_i` or `ecp256dh_r` has correctness or security implications. One may argue that passing `0` is undesired, but so is passing `1`, or any other predictable input. Ideally, the scalar reduced modulo the prime must be (close to) uniformly chosen from the `[1..p-1]` interval, but that's not something that can be enforced in the F* interface.

Ultimately, the decision of whether we restrict the range of the scalar or not depends on our interpretation of the mathematical definition of Diffie-Hellman functions. If we believe the domain has to be restricted to scalars in `[1..p-1]`, then both the specification and the implementation must enforce this restriction.

s-zanella:
> Just to mention:
> https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf
> 5.6.1.2 ECC Key-Pair Generation - Given valid domain parameters, each valid private key d is an integer that is randomly selected in the interval [1, n−1]
> 
> That's where the precondition was coming from :)

I understand where it comes from. But why the specification didn't have the same precondition?

Frosne:
So it is extremely helpful that you opened the pull request! We will solve the issue and will have both the code and the spec organized :)

s-zanella:
What people think of adding the precondition in the specification and the implementation?
Then, it would be meaningful to also expose an alternative function that enforces it via a runtime check.

beurdouche:
I much prefer to have two versions, one with the precondition, and one without it but with a defensive check.

Frosne:
We could have even three, 
a) without any precondition, that will make it easier to have hpke, 
b) with precondition (yes, the code will be the same)
c) defensive

s-zanella:
Having a) would require a specification without a precondition.
Note that c) requires any runtime checks to be constant-time.

msprotz:
We're going through pull requests with the other hacl maintainers and this one still doesn't seem to be settled. We don't see any dynamic checks in the P256 code that enforce that the scalar `s` is in the right range.

Do we understand this correctly?

PR #252, opened 2020-03-12
Fix HACL code following issue #1940 on F*

R1kM:
Small fixes were required following F* issue 1940, removing preconditions that should have been equivalent to False, but not detected.

### DIFF
--- a/code/curve25519/Hacl.Impl.Curve25519.Field64.Vale.fst
+++ b/code/curve25519/Hacl.Impl.Curve25519.Field64.Vale.fst
@@ -144,6 +144,7 @@ let fmul out f1 f2 tmp =
   ) else
     Vale.Wrapper.X64.Fmul.fmul_e (sub tmp 0ul 8ul) f1 out f2
 
+#push-options "--z3rlimit 500"
 [@ CInline]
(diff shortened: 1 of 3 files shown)


R1kM:
I'm not sure they can be described as "morally" wrong. The disjoint_or_equal predicate is needed because of the Vale interop, but it's mostly because we lack a more generic aliasing reasoning in the interop itself: We should not care about whether these two arrays overlap, even partially.
That being said, they were wrong in the type theory. They implied that types `lbuffer uint8 n1` and `lbuffer uint8 n2` were equal for different `n1` and `n2`.
Hope this answers your question.

code/curve25519/Hacl.Impl.Curve25519.Field64.Vale.fst
```diff
@@ -144,6 +144,7 @@ let fmul out f1 f2 tmp =
   ) else
     Vale.Wrapper.X64.Fmul.fmul_e (sub tmp 0ul 8ul) f1 out f2
 
+#push-options "--z3rlimit 500"
```
s-zanella: This increases the rlimit from 300 to 500. I guess that's reasonable.

PR #253, opened 2020-03-15
Fixed Blake2 spec and implementation.

prosecco:
The previous version of the Blake2 spec had a bug where it was producing the wrong results on some input lengths. This PR fixes the spec, adds a number of new tests, and modifies the implementation to match the new spec.


prosecco:
I closed this because I had not committed the new .c files in the dist/
I will do so and reup.

specs/Spec.Blake2.fst
```diff
@@ -419,7 +421,7 @@ val blake2_update_blocks:
   -> s:state a ->
   Tot (state a)
 
-let split a len =
+let split (a:alg) (len:nat) : nb_rem:(nat & nat){let (nb,rem) = nb_rem in nb * size_block a + rem == len} =
```
msprotz: should this move to a spec helper module or something? surely other algorithms must need this

specs/Spec.Blake2.fst
```diff
@@ -397,20 +397,22 @@ let blake2_update1 a prev m i s =
...
   -> prev:nat
-  -> m:bytes{prev + length m <= max_limb a}
+  -> rem:nat
+  -> m:bytes{prev + length m <= max_limb a /\ rem <= length m /\ rem <= size_block a}
   -> s:state a ->
   Tot (state a)
 
-let get_last (a:alg) (m:bytes) : block_s a =
-  let rem = length m % size_block a in
+let get_last_block (a:alg) (m:bytes) (rem:nat{rem <= length m /\ rem <= size_block a}) : block_s a =
```
msprotz: is there a way to rename this to make it apparent that the block is zero-padded?

specs/tests/Spec.Blake2.Test.fst
```diff
@@ -276,108 +778,67 @@ let test6_expected : lbytes 64 =
...
+  let result12 = blake2b_test 12 test12_plaintext test12_key test12_expected in
+  let result13 = blake2b_test 13 test13_plaintext test13_key test13_expected in
+  let result14 = blake2b_test 14 test14_plaintext test14_key test14_expected in
+  let result15 = blake2b_test 15 test15_plaintext test15_key test15_expected in
+  let result16 = blake2b_test 16 test16_plaintext test16_key test16_expected in
   //
   //
   // RESULT
   //
   if result1 && result2 && result3 && result4 && result5 && result6 then
```
msprotz: you are missing some test failure here

I would suggest chaining, e.g. 
```
let result = true in
let result = result && blake2s_test 1 ... in
let result = result && blake2s_test 2 ... in

if result then
  print "all tests successful"
else
 print "some test failed"
result

Comment (msprotz) [reply specs/tests/Spec.Blake2.Test.fst]: note that you should be able to throw exceptions since test () is likely in the ML effect -- if that's a style you prefer, of course.

File: specs/Spec.Blake2.fst [hunk 4]
```diff
@@ -397,20 +397,22 @@ let blake2_update1 a prev m i s =
 val blake2_update_last:
     a:alg
   -> prev:nat
-  -> m:bytes{prev + length m <= max_limb a}
+  -> rem:nat
```
msprotz: I think it's conceptually a little annoying that the remainder cannot be recovered from `prev` and the length of the input, and instead has to be passed down from the caller, but perhaps that's the best way to make sure that all the functions are in agreement as to what the remainder is exactly.

specs/Spec.Blake2.fst
prosecco: This particular split function is different from the one used in other algorithms, because Blake2 treats a *full* final block differently. Using a generic split here led to the previous bug, so trying to be very explicit here.

specs/Spec.Blake2.fst
prosecco: Good point: get_last_padded_block maybe?

specs/tests/Spec.Blake2.Test.fst
prosecco: I was trying not to short-circuit tests, since sometimes having failures in some tests but not others gives you more information. I will fix it so it captures all results.

specs/Spec.Blake2.fst
prosecco: Agreed, but this will require the (fairly complicated) logic of "split" to be repeated within this function. We can revisit later if need be.

specs/Spec.Blake2.fst
msprotz: 👍

specs/Spec.Blake2.fst
msprotz: sounds good

prosecco:
@protz Do you understand the build error? It appears to be in Poly.Lemmas or something?

msprotz:
I think it's just a fragile proof that broke. It's everywhere, including on nightly -- looks like you got unlucky.

code/blake2/Hacl.Impl.Blake2.Generic.fst
```diff
@@ -622,16 +622,23 @@ inline_for_extraction noextract
...
+  let row1 = rowi hash 1ul in
+  store_row first row0;
+  store_row second row1;
+  let h1 = ST.get() in
+  Lib.Sequence.eq_intro (as_seq h1 full)
+	(Lib.Sequence.(as_seq h1 (gsub full 0ul (size_row al)) @|
+		       as_seq h1 (gsub full (size_row al) (size_row al))));
+  let final = sub full (size 0) nn in
+  copy output final;
+  pop_frame()
```
s-zanella: Is this change needed? 
Keeping the application of the `salloc1` combinator would be more idiomatic.

prosecco: Not needed, but the code generated from the salloc1 was a bit weird with many uu__ variables. We may be able to go back to salloc while keeping the nice C code; I was experimenting with different options and this one generated decent code, so I kept it. Indeed, using salloc for all allocations would be nice.

s-zanella: I believe we could get close to the current C code with `salloc1` by introducing the same intermediate variables. The only difference would be that one cannot control the name of the allocated buffer, so `full` would become a generic `b`.
Another difference, compared to the previous code is that `salloc1` would zero the contents of the buffer in the stack when done. This still uses `memset`, so might not be honored by the compiler, but we will replace it with a more robust version eventually.

s-zanella: I don't see any `uu_` variables in the previous code, only common subexpressions that could be assigned to intermediate variables.

prosecco: Let me try to do another pass with salloc and see what happens.
Regarding salloc, I'd like to make some documentation on when and how to use it.
If we think this is a good style, we may as well do it more universally around hacl*.

prosecco: fixed, and green, it seems. let's merge?

s-zanella: There is some documentation in the tutorial: https://github.com/project-everest/hacl-star/blob/dev/lib/tutorial/Tutorial.fst#L260

PR #254, opened 2020-03-16
Squaring for MM

Frosne:
Montgomery multiplication consists of 2 parts: big multiplication followed by reduction. In several cases the multiplication could be replaced by more efficient squaring algorithm, that preserves some already computed values.


Frosne:
Will do.

msprotz:
As more code gets added to P256, it would really be good to see more `#set-options "--fuel 0 --ifuel 0"` make their way into your codebase, along with adequate usage of `#push-options` and `#pop-options` (as opposed to `#reset-options`). This is something that was requested by Santiago and I on the previous pull request.

For instance, the first file that appears in the diff uses rlimit 1000: https://github.com/project-everest/hacl-star/pull/254/files#diff-d9fbb714d22989cbd15bd967a6a84396R597

Here are the reasons why I'm nervous about this:
- rlimit 1000 means that z3 is having a genuinely hard time discharging the proof obligation; this proof is likely fragile or brittle and might break in the future, and will be very hard to fix for someone not familiar with the proof
- without fuel and ifuel zero, proofs are slower because z3 explores many inversions / unrollings -- as more code arrives in HACL*, this means slower builds
- there is no documentation on what is the expected level of fuel or ifuel for any given proof, and this means that if a proof breaks and someone needs to fix it, they can't tell if the proof was meant to work with fuel/ifuel 0, or with some other value -- this means more struggle waiting to happen in the future.

(Side-note: my preferred style is to *never* use fuel or ifuel, relying on local calls to `allow_inversion` for ifuel and normalization-based unrollings for fuel, a trick I learned from Santiago... I wrote a bunch about this topic in the upcoming refresh of the Low* tutorial, see for instance https://github.com/FStarLang/kremlin/blob/protz_tutorial/book/tutorial/specs/Spec.Bignum.fst#L42 or https://github.com/FStarLang/kremlin/blob/protz_tutorial/book/tutorial/code/Impl.Bignum.fst#L99 or https://github.com/FStarLang/kremlin/blob/protz_tutorial/book/tutorial/code/Impl.Bignum.fst#L149 for a preview)

Thanks,

Jonathan

s-zanella:
I strongly support @protz suggestion.
This is a larger issue with the P-256 codebase, not specific to this PR. 
I've been setting "--fuel 0 --ifuel 0" and reasonable `--z3rlimit` for files I had to update, but there are many others that still use the default or even greater fuel settings, mostly for the sake of retrying fragile proofs. This can now be better achieved using `--quake n`, although this is still suboptimal,

prosecco:
I agree in principle, and I am trying to be more disciplined about this in my own code.
@protz: the new Low* tutorial is certainly going to help, since may of the current HACL* developers
evolved their programming style without this guidance.

More generally, though, it sounds like we’re saying that the F* defaults are just plain wrong for Low* code.
- Maybe we would like all files to have fuel/ifuel set to 0 and change this for specific functions?
- Could we offer generic lemmas that should be called for unrolling lemmas/inverting types, instead
  of relying on obscure-sounding F* options?

——

The rest of this comment is a more philosophical point that is unrelated to the current issue:

At an even higher level, I am worried that the programming style we are promoting for Low*,
even with the new tutorial, is focused too much on pragmatism and less on user experience.
We are telling peoplt that if you have no idea why F*/Z3 is doing something, do a step by step
proof using calc, often for trivial equalities. I really dislike this direction and think that asking
people to use super-detailed proofs is a little bit of a failure in our methodology. At that stage,
many people would rather switch to Coq.

parno:
FWIW, I tend to agree with @prosecco's philosophical point.  One of the things that attracted me to software verification (in the form of Dafny at the time) and made it seem usable was that function definitions "just worked", unlike in, say TLA+ or Coq, where everything has to be unfolded explicitly.  Having lived in the automatic world for quite a while (in both Dafny and F*), when I had to move to fuel 0, ifuel 0 settings to interact with the HACL* specs, it led to some rather unintuitive proof failures, since my mental model didn't account for the lack of fuel.

That said, using rlimits of 1000+ is also a bad habit, as @msprotz notes, since having fragile proofs like that are just mines waiting to be hit later.  It would be nice to be able to produce stable proofs without giving up on automation.

Frosne:
Thanks for the comments! 

I will close the pull request for now and open it when the suggested changes will be done :)

s-zanella:
I don't believe @msprotz is suggesting giving up on automation. That isn't my suggestion either.
A basic knowledge of the concepts of `fuel` and `ifuel` and how F* interacts with Z3 goes a long way, and probably can be summarized in a couple of paragraphs.

Insisting on always using `--fuel 0 --ifuel 0` would be a bad idea, but asking F* programmers to develop an intuition of why they need to use a higher setting, and asking them to use it sparingly and knowingly in my opinion improves the programmer experience with F* besides improving proof stability and maintainability.

s-zanella:
Reopened the PR to keep the discussion alive.
Also this is not an issue with this PR, so I think that it could be merged now.

msprotz:
There's several points being brought up in this discussion, let me try to summarize.

- Yes, we should adopt better defaults and discipline ourselves to always use fuel/ifuel 0 by default, if only for pragmatic reasons. Let's hope we can convey that to current and future users of Low*. I agree that it's not ideal but that's the reality of working with a large-scale proof development in the presence of a large SMT context. (I was entirely pragmatic in my comment, and I don't argue for giving up on automation; rather, I'm arguing that a precise understanding is likely to give you *more* automation rather than a hard-to-diagnose failure followed by a very manual proof.)

- I agree that having nicer lemmas to selectively allow inversion and unrolling of functions would go a long way. Documentation is also important -- I don't know where allow_inversion is documented, for instance. Having an equivalent of allow_inversion (e.g. allow_unrolling) would also be excellent, and has been proposed on Slack by @Chris-Hawblitzel (https://everestexpedition.slack.com/archives/C0N8CV6NA/p1583372040038400)

- There are larger philosophical questions, i.e. why do we have to be so explicit with unrolling/inversion, or with calc proofs, when one could reasonably hope to rely on automation, since it's the promise of F* somewhat. This is a long-running concern, and has multiple explanations. Some, stemming from underlying limitations of the SMT solver (z3 just doesn't work for non-linear arithmetic, period) -- maybe we should forbid non-linear arithmetic (like Vale does) in HACL* code and demand that anything be proven with the (now much-improved) ring tactic? Some other limitations are open problems, i.e. is there any way to automatically figure out which definitions should be unrolled for a given proof (I doubt it?), or can something be done to improve the SMT context, or the performance of z3...?

PR #256, opened 2020-03-17
lib: PrintSequence: remove dec_pad printing functions

mtzguido:
These will no longer be exported by F*, and were not behaving well in any case (they were not fixed-width). They do not seem to be used anywhere besides defining these wrappers.

### DIFF
--- a/lib/Lib.PrintSequence.fst
+++ b/lib/Lib.PrintSequence.fst
@@ -18,9 +18,6 @@ let print_nat8_hex_pad x =
 let print_nat8_dec x =
   IO.print_uint8_dec (u8_to_UInt8 (u8 x))
 
-let print_nat8_dec_pad x =
-  IO.print_uint8_dec_pad (u8_to_UInt8 (u8 x))
-
 
 let print_nat32_hex x =
   IO.print_uint32 (u32_to_UInt32 (u32 x))
@@ -30,9 +27,6 @@ let print_nat32_hex_pad x =
 
 let print_nat32_dec x =
   IO.print_uint32_dec (u32_to_UInt32 (u32 x))
-
-let print_nat32_dec_pad x =
-  IO.print_uint32_dec_pad (u32_to_UInt32 (u32 x))
 
 
 let print_nat64_hex x =
@@ -44,9 +38,6 @@ let print_nat64_hex_pad x =
 let print_nat64_dec x =
   IO.print_uint64_dec (u64_to_UInt64 (u64 x))
 
-let print_nat64_dec_pad x =
-  IO.print_uint64_dec_pad (u64_to_UInt64 (u64 x))
-
 
 let print_uint8_hex x =
   IO.print_uint8 (u8_to_UInt8 x)
@@ -56,9 +47,6 @@ let print_uint8_hex_pad x =
 
 let print_uint8_dec x =
   IO.print_uint8_dec (u8_to_UInt8 x)
-
-let print_uint8_dec_pad x =
-  IO.print_uint8_dec_pad (u8_to_UInt8 x)
 
 
 let print_uint32_hex x =
@@ -70,9 +58,6 @@ let print_uint32_hex_pad x =
 let print_uint32_dec x =
   IO.print_uint32_dec (u32_to_UInt32 x)
 
-let print_uint32_dec_pad x =
-  IO.print_uint32_dec_pad (u32_to_UInt32 x)
-
 
 let print_uint64_hex x =
   IO.print_uint64 (u64_to_UInt64 x)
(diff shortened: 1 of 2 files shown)


mtzguido:
Alright, if they were indeed use I have no issue in keeping them! How about instead of taking that extra argument I just pad to the maximum possible string length of the decimal number?

s-zanella:
@beurdouche It would be nice to have some visibility of these projects. If they track `FStar@master`, they should be routinely updated to keep up with breaking changes upstream (or routinely broken by them).
A link to a public repository, where available, would already help. A link to their CI logs would be even better.
Perhaps rather than extending F* ulib with functions that don't have a clear semantics, these functions could added and maintained downstream?

msprotz:
Hi Benjamin,

I looked at the respective commit messages (FStarLang/FStar@87f717022263ab8f60eb097ece2fc5cf854be4b8) and in HACL* (3bea7de03f8eb616bb59c2a9f16037db0ef00f1f) and I could not find any indication of the intent of these functions, the purpose that they serve for projects outside of Everest, or tests that would indicate what the intended behavior is. I also don't remember any prior discussion, pull request or review that would've led to a discussion of the intended semantics and usage of these functions, and a cursory grep in the Slack archives doesn't seem to yield anything.

In this situation, it's very difficult for F* maintainers like Guido to know what to do when there's no CI or tests that would cover these functions. Asking them to be aware of things that only *you* know of is, in my opinion, unfair, especially when Everest CI is currently blocked on this. Furthermore, I'm sure you didn't mean that, but it's very easy to interpret your mention of "more critical" projects as a derogatory comment towards Everest.

What would be helpful going forward is:
- being concrete about what projects are relying on these functions; it's much more motivating to know that they're used by, e.g. MLS, rather than by some unspecified critical project
- writing tests; if you care enough about the usage of these functions, there should be at least basic tests in F* under CI -- this is the reality of most software projects: if you code isn't covered by CI, then any would-be contributor has a license to change it
- owning the code; since you authored it, perhaps *you* could submit a pull request to F* to fix the behavior and give these functions clearer semantics?

Everest is a fast-moving project; however, we commit to not breaking whatever is under Everest CI (F*, KreMLin, HACL*, miTLS, quackyducky), and F* hackers regularly go out of their way to fix projects that they don't normally contribute to, like HACL*. What we can't guarantee, however, is not breaking things that are invisible to us or not covered by our CI.

If your project is critical enough that it can't be talked about and that we can't be made aware of it, you might find it easier to maintain, as Santiago suggests, this kind of code on your side, so that you can keep the existing behavior of the code without any danger of someone trying to clean it up.

Thanks,

Jonathan

beurdouche:
> @beurdouche It would be nice to have some visibility of these projects. If they track `FStar@master`, they should be routinely updated to keep up with breaking changes upstream (or routinely broken by them).

Most of the external projects rely on HACL* Lib to avoid constant breakages from F*, and yes, that's also why we built Lib...

> A link to a public repository, where available, would already help. A link to their CI logs would be even better.

Ideally we could indeed, but some of the important ones are not opensource.

> Perhaps rather than extending F* ulib with functions that don't have a clear semantics

I guess you certainly mean you would have done better, I will just ignore your rudeness...

> these functions could added and maintained downstream?

They probably could have yes. But I wonder what would be the goal of putting those in Lib when people might need them in F*.

Frosne:
Guys! Let's stay peaceful and help each other...

prosecco:
Guys, I think we don’t need to get so confrontational so quickly on such things.

There are two questions at play:

- The FStar.Uint* changes that are needed by the switch to stdint.
  These seem fine and justified to me.
  Of course, there may be many other projects relying on F*/ulib,
  and we should find a way of communicating changes to these groups.

- The HACL Lib.Print* functions that are proposed for deletion.
  This is a bit more risky, since HACL* is used by many people that are invisible to the F* team.
  In this case, asking @beurdouche and others at Inria for a review is the right move.
  We’d probably not want to delete these functions, but reimplement them with the new F* libraries.
  Again @beurdouche or I are best placed to help with this, and I apologize if we were not quick enough to respond.

More generally, Lib should be treated as our documentation of what we need from the F* ulib within HACL* and in projects that depend on HACL*. These other projects cannot all be put under Everest CI, but they promise to use on Lib* files.
So, Lib is indeed our CI protection; if it breaks, then F* has broken some project we are building based on HACL*.

-Karthik

mtzguido:
Yes, let's not get worked up over a silly thing. Some documentation would have been nice but I should have realized that things in `lib/` are externally facing and been more careful. Good thing we have PRs :).

This PR is anyway moot now since I'm restoring the `dec_pad` functions in F*. Closing it.

beurdouche:
@mtzguido Again, Thanks a lot for the discussion and the fix, this is highly appreciated as usual :)

PR #259, opened 2020-03-18
tidy up lemma_mult_{le,lt}_right calls

mtzguido:
There is a mistake in FStar.Math.Lemmas where lemma_mult_lt_right is actually a subcase of lemma_mult_le_right (since it ensures `b * a <= c * a` instead of `b * a < c * a`). This will be fixed soon, so call lemma_mult_le_right instead.

One file, vale/code/arch/x64/interop/Vale.AsLowStar.MemoryHelpers.fst, even defines and uses the correct version. I'm keeping that one for now so we can commit these changes before fixing F*, but once that's done this should just call the F* lemma.

---

After merging this, I'm fixing `lemma_mult_lt_right` in F* and making another PR for HACL to use it in the Vale.AsLowStar.MemoryHelpers module, removing the custom one. As soon as this PR is merged I'm updating Math.Lemmas in F* (I have an orange from CI, due to this).


mtzguido:
> There are many other similar issues with `FStar.Math.Lemmas`, which has grown in a not very organized manner (e.g. duplicates lemmas, lemmas that are special case of others, lemmas whose proofs don't even require Z3 queries).
> The main problem fixing them is keeping backwards compatibility.

Indeed. I doubt we can really fix it in a backwards-compatible way, but we should do it anyway. We should probably slap an fsti on it-- @R1kM had one written down (though it might be a bit stale now?). It's anyway a separate issue, I'm mostly doing this to merge FStarLang/FStar#1949 so I don't have to debug these proofs anymore.

s-zanella:
It was more of a passing comment. Nothing bad with this PR and fixing `lemma_mult_lt_right`, which should be backwards compatible.

mtzguido:
>It was more of a passing comment. 

I was also just thinking out loud about `FStar.Math.Lemmas` :). I'd like to tidy it up sometime soon. Sorry if it seemed like complaining. And thanks for reviewing!

s-zanella:
> I was also just thinking out loud about FStar.Math.Lemmas :). I'd like to tidy it up sometime soon. Sorry if it seemed like complaining. And thanks for reviewing!

No worries, I didn't take it as a complaint. Keep up the good work!

PR #262, opened 2020-03-23
Javascript bindings for HACL-Wasm

denismerigoux:
The bindings are described only using `bindings/js/api.json`, and are exported in this npm package : https://www.npmjs.com/package/hacl-wasm.

New rules have been added in the `Makefile`
* `all-unstaged` now generates the npm package material in `dist/wasm`
* `doc-wasm` produces the documentation of the npm package in `dist/wasm/doc/out`
* `test-wasm` runs the test vectors on the exposed Javascript function.


denismerigoux:
The slow curve is removed!

denismerigoux:
Indeed, I forgot how to `git rm`.

bindings/js/.npmignore
```diff
@@ -0,0 +1,7 @@
+*.wast
+doc/
+main.html
+main.js
+browser.js
+README
+Hacl_Curve25519_64_Slow.wasm
```
msprotz: remove this one too?

Makefile
```diff
@@ -704,32 +694,81 @@ DEFAULT_FLAGS = \
...
+
+dist/wasm/doc/readable_api.js: dist/wasm/package.json
+	cd dist/wasm && \
+	mkdir -p doc && \
+	node api_doc.js
+
+dist/wasm/doc/out/index.html: dist/wasm/doc/readable_api.js
+	jsdoc $< -d $(dir $@)
+
+wasm-doc: dist/wasm/doc/out/index.html
```
msprotz: can the CI target depend on wasm-doc so that we can upload the JS documentation on github, too?

bindings/js/api.js
```diff
@@ -0,0 +1,285 @@
...
+      !(candidate.length === length) ||
+      !(candidate.constructor === Uint8Array)
+    ) {
+      throw new Error(
+        "name: Please ensure the argument " + candidate + " is a " + length + "-bytes Uint8Array."
+      );
+    }
+  };
+
+  const malloc_array = function(array) {
```
msprotz: seems more like a stack allocation to me? maybe `alloca_array`?

msprotz: or even `copy_array_to_stack`

bindings/js/api.js
```diff
@@ -0,0 +1,285 @@
...
+      memory[pointer + i] = array[i];
+    }
+    return pointer;
+  };
+
+  const read_memory = function(ptr, len) {
+    var result = new ArrayBuffer(len);
+    (new Uint8Array(result).set(new Uint8Array(Module.Kremlin.mem.buffer)
+      .subarray(ptr, ptr + len)));
+    return new Uint8Array(result);
```
msprotz: why is this done with a `set` while the `malloc_array` above is done with a `for`-loop?

bindings/js/.npmignore
denismerigoux: Hum no we need this one. Indeed in order to publish a new version of the package I `make` for Kremlin to generate the wasm files in `/dist/wasm`, and then files are copied from `bindings/js` to `dist/wasm`. At this point, `dist/wasm/.npmignore` effectively controls what goes into the npm package. But `Hacl_Curve25519_64_Slow.wasm` is generated by Kremlin and is still in `dist/wasm`, so we have to explicitly ignore it in `.npmignore`. Indeed, the behaviour of `.npmignore` is that is ignores whatever is written in `.gitignore` (sic)...

msprotz: aah, ok, thanks for the explanation!

polubelova: But you can change the Makefile target so that Kremlin won't generate `Hacl_Curve25519_64_Slow.wasm`

bindings/js/api.js
```diff
@@ -0,0 +1,282 @@
...
+        size = protoRet.size;
+      }
+      let retBuf = new ArrayBuffer(size);
+      (new Uint8Array(retBuf)).set(read_memory(pointer.value, size));
+      return retBuf;
+    });
+    if (return_buffers.length == 1) {
+      return_buffers = return_buffers[0]
+    }
+    memory[0] = sp;
```
msprotz: comment here? just to indicate you're resetting the stack pointer

code/curve25519/Makefile
```diff
@@ -27,9 +27,8 @@ $(HACL_HOME)/dist/vale/curve25519-inline.h:
 # CUSTOMIZE HERE: if necessary, move the bundle from $(HACL_HOME)/Makefile to
 # $(HACL_HOME)/Makefile.common; then, provide suitable compile flags.
 dist/Makefile.basic: $(filter-out %/prims.krml,$(ALL_KRML_FILES)) | dist
-	$(KRML) $^ -o libcurve.a $(BASE_FLAGS) $(VALE_BUNDLES) $(CURVE_BUNDLE_LOCAL) \
-	  -no-prefix 'Hacl.Impl.Curve25519.Field64.Local' \
-	  -drop 'Hacl.Impl.Curve25519.Field64.Local' \
+	$(KRML) $^ -o libcurve.a $(BASE_FLAGS) $(VALE_BUNDLES) $(CURVE_BUNDLE_BASE) \
```
msprotz: you might want to revert this

Makefile
denismerigoux: I just did that but then `jsdoc` is not installed (hence the failing build), so I can't generate doc with CI.

bindings/js/.npmignore
denismerigoux: Thanks @polubelova, I added a `bundle` option that should do that :)

bindings/js/api.js
denismerigoux: Done

code/hash/Hacl.Hash.Core.SHA2.fst
```diff
@@ -60,7 +60,8 @@ let index_h (a: sha2_alg) (i: U32.t): ST.Stack (word a)
     | SHA2_384 -> B.recall h384; IB.recall_contents h384 Constants.h384; h384.(i)
     | SHA2_512 -> B.recall h512; IB.recall_contents h512 Constants.h512; h512.(i)
 
-open Hacl.Hash.Core.SHA2.Constants
+let k224_256 = IB.igcmalloc_of_list HS.root Constants.k224_256_l
+let k384_512 = IB.igcmalloc_of_list HS.root Constants.k384_512_l
```
msprotz: why this change? as far as I remember it was important to have these constants exposed as a separate module per the comment regarding Vale... maybe it was to avoid vale taking a dependency on the whole code?

bindings/js/api.js
denismerigoux: Because I was not aware that `set` had an additional argument for an offset, changed

dist/wasm/browser.js
```diff
@@ -0,0 +1,42 @@
+// This module expects that all custom JS files have already been brought into
```
msprotz: do we want this file under version control? if not, there should be a Makefile rule to delete it

providers/evercrypt/fst/EverCrypt.Hash.fst
```diff
@@ -134,13 +134,16 @@ let init #a s =
 
 friend Vale.SHA.SHA_helpers
 
+// Avoid a cross-compilation unit symbol visibility... duplicate locally.
```
msprotz: ah, excellent, I see the fix now -- this is great, and will allow marking the constants as static 👍

dist/wasm/browser.js
msprotz: same question about the next few files, e.G. main.js, browser.html

code/hash/Hacl.Hash.Core.SHA2.fst
denismerigoux: This I don't know, it was in the `protz_wasm` branch when I took it over

dist/wasm/browser.js
denismerigoux: Good catch ! I only need `loader.js` and `shell.js` to make things work

denismerigoux: However I left all the files produced by Kremlin (`main.js`, `main.html`, etc) under version control, do you also want me to delete them from git ?

providers/evercrypt/fst/EverCrypt.Hash.fst
denismerigoux: Ok, I trust you on this :)

bindings/js/api.js
denismerigoux: Done

dist/wasm/browser.js
denismerigoux: Ok, deleting them with a Makefile rule

denismerigoux:
Thanks @msprotz  for the review ! For `jsdoc` you need to install it with 
```
npm install -g jsdoc
```

bindings/js/api.js
```diff
@@ -0,0 +1,283 @@
...
+    if (args.length != expected_args_number) {
+      throw Error("wrong number of arguments to call the F*-wasm function " + loc_name + ": expected " + expected_args_number + ", got " + args.length)
+    }
+    let memory = new Uint32Array(Module.Kremlin.mem.buffer);
+    let sp = memory[0];
+    var var_lengths = {};
+    // Populating the variable length arguments by retrieving buffer lengths
+    proto.args.map((arg) => {
+      if (arg.type === "buffer") {
+        if ((typeof arg.size === "string") && (arg.interface_index !== undefined)) {
```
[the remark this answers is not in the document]
msprotz: not sure there's much to do about it, but wondering whether the checks for well-formedness can be shared between the validator and this?

bindings/js/api.js
```diff
@@ -0,0 +1,283 @@
...
+      if (arg.type === "buffer") {
+        var size;
+        if (typeof arg.size === "string") {
+          size = var_lengths[arg.size]
+        } else {
+          size = arg.size;
+        }
+        var argByteBuffer;
+        if (arg.kind === "input") {
+          let func_arg = args[arg.interface_index];
```
msprotz: what was your conclusion regarding whether the `index` was really necessary? did you end up having non-sequential numbering anywhere?

bindings/js/api.js
```diff
@@ -0,0 +1,283 @@
...
+      proto.args[pointer.index].kind === "output"
+    ).map(pointer => {
+      let protoRet = proto.args[pointer.index];
+      var size;
+      if (typeof protoRet.size === "string") {
+        size = var_lengths[protoRet.size];
+      } else {
+        size = protoRet.size;
+      }
+      let retBuf = new ArrayBuffer(size);
```
msprotz: why is this not just `return read_memory(pointer.value, size)`? seems like you're doing an extra copy

bindings/js/api_test.js
```diff
@@ -0,0 +1,118 @@
+var path = require('path');
+var HaclWasm = require(path.resolve(__dirname,'./api.js'))
+var test_vectors = require(path.resolve(__dirname, './api.json'));
+var eq_array = function(a1, a2) {
+  if (a1.length !== a2.length) {
+    return false
```
msprotz: semicolons at the end of the line would be great for consistency with the rest of your code... have you thought about running `jslint` somewhere?

Makefile
```diff
@@ -704,32 +694,86 @@ DEFAULT_FLAGS = \
...
+# Doesn't work in WASM because un-materialized externals for AES128
+dist/wasm/Makefile.basic: FRODO_BUNDLE = -bundle Hacl.Frodo.KEM,Frodo.Params,Hacl.Impl.Frodo.*,Hacl.Impl.Matrix,Hacl.Frodo.*,Hacl.Keccak,Hacl.AES128
+
+# Doesn't work in Wasm because it uses assembler intrinsics
+ECDSA_BUNDLE = -bundle Hacl.Impl.ECDSA,Hacl.Impl.ECDSA,Hacl.Impl.ECDSA.*,Hacl.Impl.P256.*,Hacl.Impl.P256,Hacl.Spec.P256.*,Hacl.Impl.SolinasReduction,Hacl.Impl.LowLevel
+
+# No Vale Curve64; no "Local" Curve64 (local Makefile hack); only the slow 64
+# one
+dist/wasm/Makefile.basic: CURVE_BUNDLE = $(CURVE_BUNDLE_BASE) \
+	-bundle 'Hacl.Curve25519_64_Slow' \
```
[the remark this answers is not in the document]
denismerigoux: Ok, but actually this line 738 I added is not preventing the slow wasm curve from being generated...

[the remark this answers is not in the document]
msprotz: I guess the comment should read: no vale curve64, no local curve64, no slow 64 curve, only the 51 one -- line 738 *does* disable curve64/slow, correct? (curve64/slow = curve64 with intrinsics implement in Low*)

the only curve we're distributing for wasm is curve51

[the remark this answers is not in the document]
denismerigoux: You're right about the comment, I modified it so that it's clearer. I don't know what line 738 is doing, but `make dist/wasm/Makefile.basic` does produces a file named `Hacl_Curve25519_64_Slow.wasm`, even with line 738 enabled. I'll investigate with `-d reachability`

bindings/js/.npmignore
```diff
@@ -0,0 +1,6 @@
+*.wast
+doc/
+main.html
```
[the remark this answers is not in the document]
denismerigoux: You're right !

bindings/js/.npmignore
```diff
@@ -0,0 +1,6 @@
+*.wast
```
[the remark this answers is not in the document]
denismerigoux: Hum, it is still useful because when you run `make dist/wasm/package.json` it copies all the npm package material from `bindings/js` into `dist/wasm` ; if you don't keep the `.gitignore` then your git tree is going to be dirty with things in `dist/wasm` after your `make` call.

Makefile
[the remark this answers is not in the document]
msprotz: ah good catch, Curve25519_64_Slow should indeed be eliminated... thanks for investigating

bindings/js/.npmignore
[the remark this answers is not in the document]
denismerigoux: Aaah you're right about the `bindings/js/.gitignore`, I'll delete it. I thought you were talking about `dist/wasm/.gitignore`.

bindings/js/api.js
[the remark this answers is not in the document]
denismerigoux: So the lines of code that you're highlighting here are not checks _per se_. Here we have to case on the type of the arguments because buffers are allocated on the wasm memory whereas base types can be passed directly. Generally all the code in `api.js` assumes that the data it is fed has passed the well-formedness check.

bindings/js/api.js
denismerigoux: After exposing all the functions, my conclusion is the following : what we need for sure is a `is_visible_in_exposed_js_api` boolean flag. For the order in which they appear in the interface, it happens that for all the things I exposed it follows the order of the Wasm arguments, but this is a matter of aesthetics. For instance for the Blake2b/s interface, the `output length` argument is currently first in the JS API, but you could as well put it last. The `interface_index` number gives us a little more flexibility in designing the API at the expense of numbering a few arguments explicitely, I don't think it's a horrible tradeoff so no strong feelings for either keeping it or removing it.

bindings/js/api.js
denismerigoux: Indeed, `read_memory` already allocates a new buffer (not just a view into the wasm memory), so this here was unnecessary belt and straps.

Makefile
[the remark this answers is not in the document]
denismerigoux: `-d reachability` output : 
```
REACHABILITY: Hacl.Curve25519_64_Slow.ecdh is a root because it isn't private
```
I tried all `-bundle` combinations, impossible to prevent `Hacl_Curve25519_64_Slow.wasm`. However, putting `-drop Hacl.Curve25519_64.Slow` achieves the exact result I want in a single command. I know this option is deprecated but, here I don't know how else to do.

bindings/js/api_test.js
denismerigoux: Indeed, I used `jshint` to get consistency now

Makefile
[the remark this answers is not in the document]
denismerigoux: Ok, this worked :
```
dist/wasm/Makefile.basic: CURVE_BUNDLE_SLOW =
```
Thanks @protz  for the debugging help !

denismerigoux:
That's it @protz, all of your comments should be addressed, including the `doc/` additions :)

msprotz:
I fixed various bundles for curve, and also disabled most HPKE variants since many of them need either curve64, p256 or vectorized chacha-poly.

F* builds seem somehow broken so I don't know if CI will come back green. I tried to regenerate the base F* container.

msprotz:
This is merged, congratulations Denis on a great piece of work!

PR #263, opened 2020-03-23
Fix executable mode

msprotz:
Plenty of files having executable mode when they shouldn't.


s-zanella:
Thanks!
This is annoying. Could it be fixed once and for all on the F* side, which should never create files with executable flags set?

msprotz:
I think this mostly comes from people using cygwin setups. I'm not sure how to avoid that...

s-zanella:
Yes, my Cygwin setup does it. But I guess there must a way for F* to create the files without the executable flag Flagging @nik @aseemr @guido to see if they have some clue how to do it.

PR #264, opened 2020-03-24
Expose AutoConfig2_init() to libquiccrypto

ad-l:


ad-l:
Changed my mind - I want to switch the TLS tests to to libeverquic-crypto and deprecate libquiccrypto instead

PR #266, opened 2020-03-27
adding a --retry 3 to EverCrypt.Hash.Incremental.init, on consulation…

nikswamy:
… with @protz

This proof was breaking after some innocuous changes in F* ulib


msprotz:
I think this is related to an inversion that has to be performed to show that maximum_hash_length is > 0. This is a just a hunch, but I was looking at this proof closely just a few minutes ago.

In any case, I'm hoping to deprecate this piece of code soon, so I'm 💯 fine with a retry.

PR #269, opened 2020-03-30
Follow up for JS bindings

denismerigoux:
Follow-up of #262, fixes a doc generation bug for hacl-star.github.io/


denismerigoux:
@prosecco @msprotz  you've already reviewed the previous PR

PR #271, opened 2020-03-31
Disable OpenSSL and BCrypt

msprotz:
This means that a user can compile any dist/ without relying on system libraries or depending on MLCrypto, a huge sigh of relief.

We leave the legacy interface depending on OpenSSL/BCrypt enabled in dist/mitls, which is msvc-compatible and will be used for miTLS build onwards.


msprotz:
This should be particularly useful for @victor-dumitrescu who would like to distribute an OPAM package with all APIs in dist/ but without requiring OpenSSL.

msprotz:
seems to be some bug in kremlin reachability analysis... will look today

msprotz:
reachability analysis was sent off the rails because of a bundle name collision -- I've made bundle collisions fatal on master to avoid future errors... I also fixed some makefile bugs that surfaced following this fatal warning

there's still a Makefile bug left in which dist/wasm is constantly rebuilt because kremlin doesn't generate a Makefile.basic in case we go to wasm... probably worth fixing by `touch`ing Makefile.basic

PR #273, opened 2020-04-01
Compatibility with ARM Neon

karthikbhargavan:
This PR includes a bunch of changes needed to make HACL* code run on ARM Neon.
The main changes are in libintvector.h, Lib.IntVector.fst, and Lib.IntVector.Intrinsics.fst.
One notable difference with NEON is the absence of arbitrary vector permutes, so this PR
also disables vec_permute and instead introduces vec_rotate_lanes functions that allow vectors to be rotated. This causes some changes in Chacha20 and Poly1305 which have now been re-verified and re-benchmarked. Finally, I added C tests that run on ARM for various primitives in tests/ and these pass on an AWS ARM64 instance. More testing may be needed on other Arm platforms.


code/ecdsap256sha2/Hacl.Impl.LowLevel.fst
```diff
@@ -21,7 +21,7 @@ open FStar.Tactics.Canon
 open Spec.P256.Lemmas
 open Lib.IntTypes.Intrinsics
 
-#set-options "--fuel 0 --ifuel 0 --z3rlimit 500"
+#set-options "--fuel 0 --ifuel 0 --z3rlimit 1000"
```
s-zanella, line 24:
```suggestion
#set-options "--fuel 0 --ifuel 0 --z3rlimit 500"
```
This was fixed by another PR. Can you revert the change?

code/ecdsap256sha2/Hacl.Impl.LowLevel.fst
```diff
@@ -735,8 +735,7 @@ val sq0: f:  lbuffer uint64 (size 4) -> result: lbuffer uint64 (size 4) -> memor
  )
 
 
-#push-options "--z3rlimit 1000"
-
+#push-options "--z3rlimit 1500 --retry 3"
```
s-zanella, line 738:
```suggestion
#push-options "--z3rlimit 1000"
```
Idem.

dist/c89-compatible/libintvector.h
```diff
@@ -1,533 +1,554 @@
...
-
-#define Lib_IntVector_Intrinsics_vec128_interleave_low64(x1,x2) \
-  (vreinterpretq_u32_u64(vzip1q_u64(vreinterpretq_u64_u32(x1),vreinterpretq_u64_u32(x2))))
-
-#define Lib_IntVector_Intrinsics_vec128_interleave_high64(x1,x2) \
-  (vreinterpretq_u32_u64(vzip2q_u64(vreinterpretq_u64_u32(x1),vreinterpretq_u64_u32(x2))))
-
-#endif
-#endif
+#ifndef __Vec_Intrin_H
```
s-zanella, line 1: I think all lines in this file show as a diff because of a line-ending discrepancy. Can you convert line-endings to the right format (I think dos2unix will do)?

code/ecdsap256sha2/Hacl.Impl.LowLevel.fst
prosecco: Joy. Of course if I revert this, the thing no longer verifies on my laptop.
And if I remove hints, Poly1305.Lemmas no longer verifies on the CI machine.
When we are working with fragile proofs, let's not be too hasty to insist on dropping z3rlimits or removing hints. The workflow for getting a working PR through is already quite painful and time-consuming.

karthikbhargavan: Ok, reverted this, and got CI green, yay.

code/ecdsap256sha2/Hacl.Impl.LowLevel.fst
karthikbhargavan: reverted this too.

dist/c89-compatible/libintvector.h
karthikbhargavan: I ran dos2unix and checked in.

karthikbhargavan:
> Can you remove `.hints` files from the PR? This makes it easier to review, avoids conflicts with `master` and also ensures proofs still verify without hints. If you really need to add `.hints` files, please select the ones that are necessary. There's some advice for creating clean PRs in https://github.com/project-everest/hacl-star/blob/master/PULL.md.

I did not know how to remove .hints and make it still pass CI. Removing all hints made it fail CI. Probably I could do a careful selection, but this version passes and is up to date with master so I'd be happy if it were merged.

s-zanella:
Thanks!
Happy to merge this.

PR #274, opened 2020-04-03
Interfaces in F* ulib for IntN and UIntN modules

aseemr:
I have added interfaces for F* UIntN and IntN modules. It invalidated all the hints that used these modules. But surprisingly only 6 proofs in hacl broke, and I could just regenerate hints for them. I had an Everest green with new hints for these 6 files.

I also had to up the rlimit in one of the problematic functions in MerkleTree.Low.


secure_api/merkle_tree/MerkleTree.Low.fst
```diff
@@ -2220,7 +2220,7 @@ val mt_flush_to_:
                (U32.v i) (U32.v (Ghost.reveal j))))))
    (decreases (U32.v i))
 #restart-solver
-#push-options "--z3rlimit 800 --fuel 1 --ifuel 0 --retry 8"
+#push-options "--z3rlimit 1500 --fuel 1 --ifuel 0 --retry 8"
```
prosecco, line 2223: "restart solver + retry 8"!!!
This looks like overkill, but I do it all the time, so hey.

aseemr:
This retry is from the last round of changes, sorry about it. The assertion at line 2309 fails but it's hard for me to fix it as I don't have any familiarity with this code.

aseemr:
Thanks Marina!

PR #275, opened 2020-04-03
OCaml API

victor-dumitrescu:
This PR contains:
* a high level OCaml API for EverCrypt and HACL* (`bindings/ocaml`), which uses the already merged, Kremlin-generated Ctypes bindings in `dist/gcc-compatible`
* a small Makefile which prepares the source archive required to publish to opam (`tools/Makefile.opam`) 
* an updated `lib/c/Lib_RandomBuffer_System.c` which uses `SYS_getrandom` on Linux if available and `/dev/urandom` as fallback

The main entry points for the API are `EverCrypt.mli` and `Hacl.mli`, with `SharedDefs.ml` containing the definitions of modules which are implemented in both. It contains almost every primitive in EverCrypt and HACL*, only excluding those which might be too tricky to use securely (e.g., the block-based API for hashing). To further enhance safety and prevent misuse all high level functions also perform runtime checks for buffer sizes and disjointness (manually lifted from F* preconditons). Currently the API uses OCaml Bytes as the type for buffers but it has been written in a generic way in which Bigstring can also be used alongside or instead of Bytes, as required.

Still to do:
* ~~add the API to CI~~
* a few functions still lack unit tests

Many thanks to @protz for his extensive help with this work!


bindings/ocaml/EverCrypt.ml
```diff
@@ -0,0 +1,282 @@
...
+  let update (alg, incr_len, st) data =
+    assert (C.size data < max_input_len);
+    incr_len := Z.add !incr_len (Z.of_int (C.size data));
+    assert (Z.lt !incr_len (incremental_input_len alg));
+    everCrypt_Hash_Incremental_update st (C.ctypes_buf data) (C.size_uint32 data)
+  let finish (alg, _, st) dst =
+    assert (C.size dst = digest_len alg);
+    everCrypt_Hash_Incremental_finish st (C.ctypes_buf dst)
+  let free (_, _, st) =
+    everCrypt_Hash_Incremental_free st
```
msprotz, line 165: can this be attached as a finalizer on the OCaml GC block? would prevent misuses of the API

bindings/ocaml/EverCrypt.ml
```diff
@@ -0,0 +1,282 @@
...
+    Error err
+  let get_result r = match UInt8.to_int r with
+    | 0 -> Success ()
+    | n -> error n
+end
+
+module AEAD = struct
+  open Error
+  open Hacl_Spec
+  open EverCrypt_AEAD
```
msprotz, line 65: a function to free the key is missing here but I'm not sure it's even in the .fst file

dist/Makefile.tmpl
```diff
@@ -134,6 +134,14 @@ include ctypes.depend
 
 all: libocamlevercrypt.cmxa
 
+lib/Lib_RandomBuffer_System_bindings.cmx:
```
msprotz, line 137: why not copy them into the `lib` and `lib_gen` directories respectively, and let the automatic dependencies kick in? wouldn't work? in any case please use `-c $<` and `-o $@` to minimize the risk of typos

msprotz: by that I mean that if they were in the right directories you wouldn't have to manually list the ocamlopt invocation

dist/gcc-compatible/Lib_RandomBuffer_System_bindings.ml
```diff
@@ -0,0 +1,8 @@
+open Ctypes
```
msprotz, line 1: is this written by hand?

tools/Makefile.opam
```diff
@@ -0,0 +1,34 @@
+# Builds a source archive required for opam release in bindings/ocaml/archive
+# The gcc-compatible snapshot is packaged as `hacl-star-raw`
+# The OCaml bindings are packaged as `hacl-star`
```
msprotz, line 3: add a word about what to do next, and how to call `opam-publish`?

dist/Makefile.tmpl
```diff
@@ -154,7 +162,16 @@ endif
 lib/%_stubs.ml lib/%_c_stubs.c: lib_gen/%_gen.exe
 	$(DY)PATH=.:$(OPENSSL_HOME):$$$(DY)PATH $<
 
+# Compiling these files raises the pointer sign warning because the Ctypes declaration for bytes expects a char*, not a uint8_t*
+# Can be remove once ctypes is upgraded to 0.17
+lib/%_c_stubs.o: CFLAGS += -Wno-pointer-sign
```
msprotz, line 167: I already added this in the Makefiles, so you may have a conflict here when you merge, by the way

dist/gcc-compatible/Lib_RandomBuffer_System_bindings.ml
victor-dumitrescu: Yes, Lib_RandomBuffer_System_bindings.ml  and _gen.ml are written by hand since the C files are also written by hand.

dist/Makefile.tmpl
msprotz: you should remove line 80 of this makefile then

bindings/ocaml/EverCrypt.ml
```diff
@@ -0,0 +1,287 @@
...
+        (ptr everCrypt_AEAD_state_s)
+        (from_voidp everCrypt_AEAD_state_s null)
+    in
+    match UInt8.to_int (everCrypt_AEAD_create_in (alg_definition alg) st (C.ctypes_buf key)) with
+    | 0 -> Success (alg, st)
+    | n -> error n
+  let encrypt (alg, st) iv ad pt ct tag : unit result =
+    (* EverCrypt.AEAD.encrypt_pre *)
+    check_sizes alg iv tag;
+    assert (C.disjoint ct tag);
```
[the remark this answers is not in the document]
victor-dumitrescu: It's only checking for physical equality, since for Bytes if they're not the same object then they can't overlap.

PR #278, opened 2020-04-06
Removing `abstract` keyword

aseemr:
This PR removes the instances of the `abstract` keyword from the code. Following are some explanations:

Hacl.Spec.Curve25519.Field51.Definition.fst: I think just `[@"opaque_to_smt]` should suffice, it seems the idea is to expose to the solver only the type of the definition.

Hacl.Spec.Ed25519.Field56.Definition.fst: same.

Lib.Sequence.fsti: I didn't even know we allowed `abstract` keyword in the interface files, seems wrong. Anyway, we are hoping that we will be able to remove `abstract` keyword soon.

Lib.Sequence.fst: Corresponds to the .fsti change.


aseemr:
Thanks for merging it @s-zanella!

PR #280, opened 2020-04-08
ECDSA Update (agile, compression) PR

Frosne:
This PR introduced a couple of new methods for ecdsa, i.e. key compression and decompression, agil-ing ecdsa and a couple of clean-ups. 
Key compression and decompression now could present key in 3 possible ways, raw (x + y), compressed (02/03 + x), notCompressed (04 + x + y). 
Additionally, to the ecdsa interface I have added a reduction method that takes a random buffer of size n, and returns a buffer, that could be used for ecdsa signature.


Makefile.common
```diff
@@ -108,7 +108,7 @@ POLY_BUNDLE=-bundle 'Hacl.Poly1305_32=Hacl.Impl.Poly1305.Field32xN_32' \
   -bundle 'Hacl.Poly1305_128=Hacl.Impl.Poly1305.Field32xN_128' \
   -bundle 'Hacl.Poly1305_256=Hacl.Impl.Poly1305.Field32xN_256'
 NACLBOX_BUNDLE=-bundle Hacl.NaCl=Hacl.Impl.SecretBox,Hacl.Impl.Box
-ECDSA_BUNDLE=-bundle Hacl.Impl.ECDSA=Hacl.Impl.ECDSA,Hacl.Impl.ECDSA.*,Hacl.Impl.P256.*,Hacl.Hash.SHA2,Hacl.Hash.Core.SHA2,Hacl.Hash.Core.SHA2.Constants,Hacl.Hash.Definitions,Hacl.Impl.P256,Hacl.Spec.P256.*,Hacl.Impl.SolinasReduction,Hacl.Impl.LowLevel[rename=Hacl_ECDSA]
+ECDSA_BUNDLE= -bundle Hacl.Impl.ECDSA=Hacl.Impl.ECDSA,Hacl.Impl.ECDSA.*,Hacl.Impl.P256.*,Hacl.Hash.SHA2,Hacl.Hash.Core.SHA2,Hacl.Hash.Core.SHA2.Constants,Hacl.Hash.Definitions,Hacl.Impl.P256,Hacl.Spec.P256.*,Hacl.Impl.SolinasReduction,Hacl.Impl.LowLevel,Hacl.Impl.LowLevel.RawCmp[rename=Hacl_ECDSA]
```
msprotz, line 111: can the files starting with `Hacl.Impl.LowLevel` go within the `Hacl.Impl.P256` namespace? I'm afraid it's a little generic and we won't know if someone uses that prefix elsewhere which algorithm it goes with

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Signature.Agile.fst
```diff
@@ -0,0 +1,319 @@
...
+
+let ecdsa_signature_step12 alg mLen m result = 
+  assert_norm (pow2 32 < pow2 61);
+  assert_norm (pow2 32 < pow2 125);
+  push_frame(); 
+    let h0 = ST.get() in 
+  let sz: FStar.UInt32.t = hash_len alg in
+  let mHash = create sz (u8 0) in    
+  
+  begin
```
msprotz, line 65: to simplify the reasoning and proof performance, you might benefit from:

```
let statically_agile_hash (a: hash_alg): Hacl.Hash.Definitions.update_st a =
  match a with
  | SHA2_256 -> hash_256 m mLen mHash
  | ...
```
and then you can have your main proof go through with `ifuel 0`, leaving only the definition above to have `ifuel 1`

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Signature.Agile.fst
```diff
@@ -0,0 +1,319 @@
...
+  let a = Spec.ECDSA.exponent_spec (fromDomain_ kInv) in 
+  lemmaFromDomain kInv;
+
+  power_distributivity (kInv * modp_inv2_prime (pow2 256) prime_p256_order) (prime_p256_order - 2) prime_p256_order;
+  power_distributivity_2 kInv (modp_inv2_prime (pow2 256) prime_p256_order % prime_p256_order) (prime_p256_order - 2);
+  lemma_mod_mul_distr_r (pow kInv (prime_p256_order - 2)) (pow (modp_inv2_prime (pow2 256) prime_p256_order) (prime_p256_order - 2)) prime_p256_order;
+
+  lemma_pow_mod_n_is_fpow prime_p256_order (pow2 256 % prime_p256_order) (prime_p256_order - 2);
+  
+  let inverse2_256 = 43790243014242295660885426880012836369732278457577312309071968676491870960761 in
```
msprotz, line 132: one good thing for readability would be to try to avoid magic constants -- for instance, here, you could do:

```
let inverse2_256 = normalize_expr (modp_inv2_prime (pow2 256) prime_p256_order) in
FStar.Pervasives.normalize_term_spec inverse2_256 (modp_inv2_prime (pow2 256) prime_p256_order);
```

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Signature.Agile.fst
```diff
@@ -0,0 +1,319 @@
...
+
+open Hacl.Impl.P256.Signature.Common
+
+module H = Spec.Agile.Hash
+module Def = Spec.Hash.Definitions
+
+open Spec.Hash.Definitions
+open Hacl.Hash.Definitions
+
+#set-options "--z3rlimit 100"
```
msprotz, line 40: I would recommend setting fuel and ifuel to zero here, as mentioned in your previous pull request

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Signature.Agile.fst
```diff
@@ -0,0 +1,319 @@
...
+    reduction_prime_2prime_order x x;
+  pop_frame();
+    isZero_uint64_CT x
+
+
+val lemma_power_step6: kInv: nat -> Lemma 
+  (Spec.ECDSA.exponent_spec (fromDomain_ kInv) == toDomain_ (pow kInv (prime_p256_order - 2)))
+
+let lemma_power_step6 kInv = 
+  let a = Spec.ECDSA.exponent_spec (fromDomain_ kInv) in
```
msprotz, line 123: for big lemmas like this, I really like `calc` statements, because they really help with proof maintenance -- it would be a good idea, for robustness of the code, to try to write calc statements if applicable

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Signature.Agile.fst
```diff
@@ -0,0 +1,319 @@
...
+
+      lemmaFromDomain (as_nat h0 r * as_nat h0 da); 
+      lemma_felem_add (as_nat h0 r * as_nat h0 da) (as_nat h0 z);
+      lemma_power_step6 (as_nat h0 kFelem); 
+
+      lemmaFromDomain (fromDomain_ br0);
+      lemmaToDomain br1;
+      assert_norm ((modp_inv2_prime (pow2 256) prime_p256_order * pow2 256) % prime_p256_order = 1);
+       
+      lemma_mod_mul_distr_l (fromDomain_ br0 * modp_inv2_prime (pow2 256) prime_p256_order) (br1 * pow2 256 % prime_p256_order) prime_p256_order;
```
msprotz, line 182: same thing about calc -- if this proof breaks and I need to fix it, I would have a much easier time with a `calc` statement

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Signature.Hashless.fst
```diff
@@ -184,46 +184,43 @@ val ecdsa_signature_core: r: felem -> s: felem -> mLen: size_t -> m: lbuffer uin
...
       
       let kFelem = nat_from_bytes_be (as_seq h0 k) in 
       as_nat h1 r == xN % prime_p256_order /\
       as_nat h1 s == (z + (as_nat h1 r) * as_nat h0 privKeyAsFelem) * pow kFelem (prime_p256_order - 2) % prime_p256_order /\
       (
-	if as_nat h1 r = 0 || as_nat h1 s = 0 then 
-	  uint_v flag == pow2 64 - 1
-	else 
-	  uint_v flag == 0
+  if as_nat h1 r = 0 || as_nat h1 s = 0 then
```
msprotz, line 195: your editor seems to hesitate between tabs and spaces -- spaces are of course preferred in HACL*, so it's good that you're avoiding tabs, but here indentation seems off

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Verification.Agile.fst
```diff
@@ -88,29 +95,50 @@ let ecdsa_verification_step1 r s =
...
   isRCorrect && isSCorrect
 
-
-inline_for_extraction noextract
-val ecdsa_verification_step23: hashAsFelem : felem -> mLen: size_t -> m: lbuffer uint8 mLen -> Stack unit
-  (requires fun h -> live h hashAsFelem /\ live h m)
-  (ensures  fun h0 _ h1 -> modifies (loc hashAsFelem) h0 h1 /\
-  (
+inline_for_extraction
+val ecdsa_verification_step23: alg: hash_alg {SHA2_256? alg \/ SHA2_384? alg \/ SHA2_512? alg (* and blake one day *)}
```
msprotz, line 99: since this refinement was already used in the type of ecdsa_signature_step12, I would suggest:
```
let is_supported_alg = function | SHA2_256 | SHA2_284 | SHA2_512 -> true | _ -> false
let supported_hash_alg = a:hash_alg { is_supported_alg a }
```

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Verification.Agile.fst
```diff
@@ -88,29 +95,50 @@ let ecdsa_verification_step1 r s =
...
+
+let ecdsa_verification_step23 alg mLen m result = 
+  assert_norm (pow2 32 < pow2 61);
+  assert_norm (pow2 32 < pow2 125);
+  push_frame(); 
+    let h0 = ST.get() in 
+  let sz: FStar.UInt32.t = hash_len alg in
+  let mHash = create sz (u8 0) in    
+  
+  begin
```
msprotz, line 121: the shared helper I mentioned earlier might come in handy here, too

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Verification.Agile.fst
```diff
@@ -88,29 +95,50 @@ let ecdsa_verification_step1 r s =
...
-  let mHash = create (size 32) (u8 0) in
-  hash_256 m mLen mHash;
-  toUint64ChangeEndian mHash hashAsFelem;
-  let h1 = ST.get() in
-  lemma_core_0 hashAsFelem h1;
-  reduction_prime_2prime_order hashAsFelem hashAsFelem;
-  Spec.ECDSA.changeEndianLemma (uints_from_bytes_be #U64 #_ #4 (as_seq h1 mHash));
-  uints_from_bytes_be_nat_lemma #U64 #_ #4 (as_seq h1 mHash);
+
+let ecdsa_verification_step23 alg mLen m result =
```
msprotz, line 113: this function is exactly the same as `ecdsa_signature_step12` -- please eliminate the code duplication

if the function needs to be duplicated in the C code (which I would not recommend), do:
```
inline_for_extraction
let shared_func () x y z =
  ...

// MyFile.fst
let first_copy = shared_func ()

// MyOtherFile.fst
let second_copy = shared_func ()
```

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Verification.Hashless.fst
```diff
@@ -0,0 +1,581 @@
...
+    let z0_zero = eq_0_u64 f0 in
+    let z1_zero = eq_0_u64 f1 in
+    let z2_zero = eq_0_u64 f2 in
+    let z3_zero = eq_0_u64 f3 in
+
+    z0_zero && z1_zero && z2_zero && z3_zero
+
+
+(* This code is not side channel resistant *)
+val isMoreThanZeroLessThanOrderMinusOne: f:felem -> Stack bool
```
msprotz, line 68: why the different convention here? `camelCase` vs `with_underscores`

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Verification.Agile.fst
```diff
@@ -1,4 +1,4 @@
-module Hacl.Impl.ECDSA.P256SHA256.Verification
+module Hacl.Impl.ECDSA.P256SHA256.Verification.Agile
```
msprotz, line 1: I feel like the `Agile` contradicts the `P256SHA256` bit in here. This file should probably be renamed.

code/ecdsap256sha2/Hacl.Impl.LowLevel.fst
```diff
@@ -701,151 +920,179 @@ let lemma_320_64 a b c d e u =
...
     (
       let f0 = Lib.Sequence.index (as_seq h0 f) 0 in 
-      as_nat h1 result + uint_v c * pow2 64 * pow2 64 * pow2 64 * pow2 64  = uint_v f0 * as_nat h0 f
-
-    ) /\
-
+      as_nat h1 result + uint_v c * pow2 64 * pow2 64 * pow2 64 * pow2 64  = uint_v f0 * as_nat h0 f) /\
+      (
+  let f = as_seq h0 f in 
+  let f0 = Lib.Sequence.index f 0 in
```
msprotz, line 932: for this and all other occurrences below, I would suggest defining helpers that return a tuple, for readability

code/ecdsap256sha2/Makefile
```diff
@@ -14,7 +14,7 @@ FSTAR_ROOTS += ../tests/Hacl.Test.ECDSA.fst
 
 #BASE_FLAGS+= -funroll-loops 4
 
-CFLAGS += -I ../../../lib/c -march=native -mtune=native -O3
+CFLAGS += -I ../../../lib/c -march=native -mtune=native -O3 -Wno-unused-function
```
msprotz, line 17: which function triggers this warning? kremlin should eliminate all unused functions, and if it does not, it's probably a bug -- do you have a repro?

msprotz, line 17: which function triggers this warning? kremlin should eliminate all unused functions, and if it does not, it's probably a bug -- do you have a repro?

code/ecdsap256sha2/Makefile
```diff
@@ -32,11 +32,13 @@ dist/libecdsa.a: dist/Hacl_ECDSA.o
 dist/%.o: dist/Makefile.basic
 	$(MAKE) -C dist -f Makefile.basic $(notdir $@)
 
-dist/Hacl_Test_ECDSA.exe: dist/Hacl_Test_ECDSA.o \
+dist/Hacl_Test_ECDSA.exe: dist/Hacl_ECDSA.o dist/Hacl_Test_ECDSA.o \
```
msprotz, line 35: this will trigger two parallel invocations of a recursive make in dist which will race with each other -- I would suggest depending only on `dist/libecsda.a` and changing `dist/%.o` above to be `dist/libecsda.a`

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Signature.Agile.fst
msprotz: note: I saw that almost all other places have great calc statements -- this might be coming from a previous version of the code, or might just be a particular case, but if calc didn't work for you, don't hesitate to file a bug

Makefile.common
Frosne: Sure, but the idea of not putting any specific header was that the code in this file is not p256/ecdsa specific :) It is just buffer-based arithmetics

code/ecdsap256sha2/Hacl.Impl.ECDSA.P256SHA256.Verification.Agile.fst
Frosne: :)

msprotz:
Can you remove code/ecdsa/dist from the PR? Thanks!

code/ecdsap256/Hacl.Impl.ECDSA.MontgomeryMultiplication.fst
```diff
@@ -233,7 +234,7 @@ let reduction_prime_2prime_with_carry x result  =
     let cin = Lib.Buffer.index x (size 4) in 
     let x_ = Lib.Buffer.sub x (size 0) (size 4) in 
         recall_contents prime256order_buffer (Lib.Sequence.of_list p256_order_prime_list);
-    let c = Hacl.Impl.LowLevel.sub4_il x_ prime256order_buffer tempBuffer in
+    let c = Hacl.Impl.P256.LowLevel .sub4_il x_ prime256order_buffer tempBuffer in
```
msprotz, line 237: some renamings seem to have worked out poorly (extra space) -- you also have inconsistent indentation... have you considered using `fstar --print_in_place` to automatically re-indent your files?

code/ecdsap256/Hacl.Impl.P256.LowLevel.fst
```diff
@@ -163,7 +171,66 @@ let copy_conditional out x mask =
   lemma_eq_funct_ (as_seq h1 out) (as_seq h0 out);
   lemma_eq_funct_ (as_seq h1 out) (as_seq h0 x)
 
+(*
```
msprotz, line 174: please remove this bit as it's now in another file

code/ecdsap256/Hacl.Impl.P256.LowLevel.fst
```diff
@@ -520,82 +595,288 @@ let mul1_add f1 u2 f3 result =
...
+    as_nat h2 r * uint_v f2 * pow2 64 * pow2 64 + 
+    as_nat h2 b2 * pow2 64 * pow2 64 + 
+    as_nat h3 r * uint_v f3 * pow2 64 * pow2 64 * pow2 64;
 
-#pop-options
+    (==)
+    {    
+      assert(Lib.Sequence.index (as_seq h2 bk1) 0 == Lib.Sequence.index (as_seq h2 out) 0);
+      assert(Lib.Sequence.index (as_seq h2 bk1) 1 == Lib.Sequence.index (as_seq h2 out) 1)
+    }
```
msprotz, line 792: why is `eq_intro` not working in those cases?

Makefile.common
```diff
@@ -108,7 +108,11 @@ POLY_BUNDLE=-bundle 'Hacl.Poly1305_32=Hacl.Impl.Poly1305.Field32xN_32' \
   -bundle 'Hacl.Poly1305_128=Hacl.Impl.Poly1305.Field32xN_128' \
   -bundle 'Hacl.Poly1305_256=Hacl.Impl.Poly1305.Field32xN_256'
 NACLBOX_BUNDLE=-bundle Hacl.NaCl=Hacl.Impl.SecretBox,Hacl.Impl.Box
-ECDSA_BUNDLE=-bundle Hacl.Impl.ECDSA=Hacl.Impl.ECDSA,Hacl.Impl.ECDSA.*,Hacl.Impl.P256.*,Hacl.Hash.SHA2,Hacl.Hash.Core.SHA2,Hacl.Hash.Core.SHA2.Constants,Hacl.Hash.Definitions,Hacl.Impl.P256,Hacl.Spec.P256.*,Hacl.Impl.SolinasReduction,Hacl.Impl.LowLevel[rename=Hacl_ECDSA]
+P256_BUNDLE_RHS=Hacl.Impl.ECDSA.*,Hacl.Hash.SHA2,Hacl.Hash.Core.SHA2,Hacl.Hash.Core.SHA2.Constants,Hacl.Hash.Definitions,Hacl.Impl.SolinasReduction,Hacl.Impl.P256.*[rename=Hacl_P256] \
```
msprotz, line 111: per our discussion on Slack, you may want to use `rename-prefix` instead of `rename` to make sure that the resulting names in the C code starts with Hacl_P256 (and are not just moved Hacl_P256.c)

Makefile.common
```diff
@@ -108,7 +108,11 @@ POLY_BUNDLE=-bundle 'Hacl.Poly1305_32=Hacl.Impl.Poly1305.Field32xN_32' \
...
+  -bundle Hacl.SolinasReduction.Lemmas
+P256_BUNDLE=-bundle Hacl.Interface.P256=$(P256_BUNDLE_RHS)
+ECDSA_BUNDLE=-bundle Hacl.Interface.P256.ECDSA=$(P256_BUNDLE_RHS)
```
msprotz, line 114: this one and the one below are not used anywhere -- I would just suggest removing them

dist/c89-compatible/Hacl_Interface_P256_DH.c
```diff
@@ -0,0 +1,36 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+
+#include "Hacl_Interface_P256_DH.h"
+
+uint64_t Hacl_Interface_P256_DH_ecp256dh_i(uint8_t *result, uint8_t *scalar)
+{
+  return Hacl_Impl_P256_DH_ecp256dh_i(result, scalar);
```
msprotz, line 29: I think here because there's a call from Hacl_Interface_P256_DH_ecp256dh_i to Hacl_Impl_P256_DH_ecp256dh_i, then Hacl_Impl_P256_DH_ecp256dh_i is forced to appear in Hacl_P256.h instead of remaining private (hidden)

dist/c89-compatible/Hacl_P256.h
```diff
@@ -0,0 +1,217 @@
...
+
+#ifndef __Hacl_P256_H
+#define __Hacl_P256_H
+
+#include "Hacl_Kremlib.h"
+#include "Hacl_Hash.h"
+#include "Hacl_Spec.h"
+
+
+uint64_t Hacl_Impl_P256_LowLevel_sub4_il(uint64_t *x, const uint64_t *y, uint64_t *result);
```
msprotz, line 40: all of these `Impl_` functions are forced to appear in the header because `Hacl_Interface_P256_ECDSA.c` is calling into them -- you would actually get a cleaner header with a single interface file instead of three separate files

code/ecdsap256/Hacl.Impl.P256.LowLevel.fst
Frosne: Solved

code/ecdsap256/Hacl.Impl.P256.LowLevel.fst
Frosne: Because I did not know about it's existence :)

Makefile.common
Frosne: Solved

polubelova:
Thanks @Frosne  for the PR! Just a few comments:
- What's the reason for having the following three files:
	code/ecdsap256/Hacl.Interface.P256.ECDSA.fsti
	code/ecdsap256/Hacl.Interface.P256.DH.fsti
	code/ecdsap256/Hacl.Interface.P256.fsti (it has the same set of functions as the two files above)
- Not sure whether it makes sense to use `[@Comment]` for functions marked as `inline_for_extraction noextract`
- Can you leave some comments in the C code for new functions describing their API?
(for example, what kind of buffer `b` does the `decompression_not_compressed_form` function expect?)
- Can you clean up a bit the tests/p256-test.c file? :)

Yeah, the proofs of the `mul` function are huge (`wide_as_nat h1 out = as_nat h0 r * as_nat h0 f`). I see no reason for that, but it doesn't matter for now.

Frosne:
Thanks @polubelova for the comments to the PR! 
Just a few remarks:

The initial idea was to divide the full code into three intersection sets: 
Code for ECDSA, code for DH, and code for both. The code is quite huge and I don´t find the idea of just giving to everyone this big chuck of beautifully written code is good. 

NB:  Having that I can´t pass CI with it (it fails with some errors [probably because of it, to check]), most probably I will erase it.

Frosne:
> as for the next PR, it would be nice to have only a high-level spec of p256 in specs/ :)

Ok :)

msprotz:
Congrats on a nice PR!

I agree about the high-level spec. Isn't it something you had started already with Santiago? One that just uses the high-level math and not just the algorithm ladder.

For the next PR, let's try to make changes more modular and easier to review. There was a lot of stuff going on here at the same time and I think that may have contributed to the overhead.

PR #281, opened 2020-04-14
A configure script for ARM

msprotz:
I am starting on proper work to ensure that "make" in dist/gcc-compatible works on ARM platforms and compiles as much as we can.

In addition to build complexities, this already uncovered several errors, such as calls into ASM symbols coming from Vale not guarded behind if EverCrypt.TargetConfig.x64

I have not fixed these yet but I'd like to this landed so that I can setup github actions to, essentially, go run make make in dist/gcc-compatible and make sure our code continues to compile on ARM.


msprotz:
Thanks for the reviews!

PR #282, opened 2020-04-17
Remove custom lemma, use FStar.Math.Lemmas instead

mtzguido:
Just a tiny cleanup. This lemma made sense before when its version in `FStar.Math.Lemmas` was wrong.


msprotz:
Looks good to me. There might be other instances in which some lemmas from fstar.math.lemmas were duplicated in Vale for the sake of having `pos` instead of `nat` -- I remember doing a pass a long while ago to remove many of those but there might still be some leftovers

PR #283, opened 2020-04-20
Attempt to fix JS doc auto-publication

denismerigoux:
Does not rely on `make all` doc generation, but rather directly generates the doc from `bindings/js` in `ci.sh`.


msprotz:
See #292 for the actual root cause of the problem.

PR #284, opened 2020-04-21
Bump Vale version to 0.3.13

mtzguido:
Needed after some changes to F*, namely that `Tm_delayed` labels are now `"Tm_delayed"` instead of `"Tm_delayed-resolved"` and the moving of `Lemma` into `FStar.Pervasives`.

CI has run over this with an old version of F* and so it has failed, trying to trigger a new one now.


mtzguido:
The windows build for the PR (not the head commit) came back green, see https://msr-project-everest.visualstudio.com/Everest/_build/results?buildId=43493&view=results. I thought that would make the check green here in GitHub, but seems not. Anyway just kicked a new one for the commit.

PR #285, opened 2020-04-21
Significantly widen the testing matrix for HACL*

msprotz:
This is essentially about testing the generated C code on a variety of platforms, OSes, and compiler toolchains. We don't need the advanced Everest CI for this since we don't care about having F* installed!

Quoting from Slack:
- gcc-7 / aarch64 bionic (passing, minimal tests)
- gcc-7 / linux64 bionic (passing, all C tests and extracted Low* tests)
- gcc-5.4 / linux64 xenial (failing because of an OCaml binding issue, same as above + OCaml package build & test)
- clang7 / linux64 xenial (failing because of the intrinsic, all C tests and extracted Low* tests)
- clang / xcode 10 (failing because of: error: inline assembly requires more registers than available, same as above)
- clang / xcode 11 (failing because of the intrinsic, same as above)as part of this CI work:
- I put dist/c/test under version control
- I wrote a mini-Makefile in dist/c/test so that one can run the extracted C tests without recompiling the world
- these tests require "full kremlib" so I'm contemplating putting generated kremlib C code under version control so that it's easy to build without verifying/extracting kremlib


msprotz:
@parno @R1kM: I could not find a way to feature-test for the buggy LLVM version -- I tried all sorts of mini-tests that I was hoping would indicate whether the compiler we use has the register allocation failure, but none of them seem to capture the bug, and since I cannot repro on my machine, I gave up trying to use Travis as an oracle and now grep for the buggy version in `cc --version`

here's an excerpt from the diff listing all the things I tried:
```
+  # For posterity, here are a list of things that don't work
+  # - checking if cc will compile curve25519-inline.h
+  # - checking if cc will compile fsqr2 as is
+  # - checking if cc will compile fsqr2 marked as extern
+  # - checking if cc will compile fsqr2 as is with an extern caller in the file to
+  # prevent it from being eliminated
+  #
+  # In short, I couldn't figure out a minimal testcase for the error:
+  #   ./curve25519-inline.h:595:5: error: inline assembly requires more registers than available
+  #
+  # Furthermore, this error only seems to happen on the exact config Travis
+  # uses.
+  # - Installing XCode 10.0 (for MacOS 10.14, Catalina) does not give me the right LLVM build
+  # - Installing XCode 10.1 (for MacOS 10.14, Catalina) does not give me the right LLVM build
+  # - Installing XCode 10.3 (for MacOS 10.14, Catalina) bails because my OSX is too recent
+  cc --version | grep -q clang-1000.11.45.5
```

@Frosne @karthikbhargavan I fixed lib_intrinsics.h to disable the first definition in the case of clang -- please check you're happy with the fix

now with my two fixes we're only left with the OCaml bindings bugs which, once fixed, will allow this PR to be merged

msprotz:
I wrote a test based on https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81300 that detects whether the compiler miscompiles the intrinsics, triggered at `./configure`-time. If it does, it uses @Frosne 's fallback implementation (in C) for the intrinsics.

The test correctly triggers for CC=gcc-5.4 -- interestingly, it also triggers for a specific build of clang-7 on linux but not on other architectures. I haven't been able to reproduce...

msprotz:
The test is not right because it ignores the bug about the argument order. I'm switching to a custom-made test that seems more reliable...

msprotz:
> The test correctly triggers for CC=gcc-5.4 -- interestingly, it also triggers for a specific build of clang-7 on linux but not on other architectures. I haven't been able to reproduce...
this was because on this configuration, we had `CC=clang` but the configure script was hardcoding `cc` instead of `$CC`. 

I've switched to a test based on a diff that checks whether the (minimized) testcase extracted from ECDSA exhibits different results with/without fallback implementations of intrinsics; if it does, it's a buggy compiler

I think we should be good for merging now

PR #290, opened 2020-04-26
clean up C tests

polubelova:


tests/blake2-arm-test.c
```diff
@@ -8,439 +8,100 @@
...
+  uint8_t *in5 = vectors2s[3].input;
+  int key_len5 = vectors2s[3].key_len;
+  uint8_t *key5 = vectors2s[3].key;
+  int exp_len5 = vectors2s[3].expected_len;
+  uint8_t *exp5 = vectors2s[3].expected;
+
+  ok = print_test2s(in_len2,in2,key_len2,key2,exp_len2,exp2) && ok;
+  ok = print_test2s(in_len3,in3,key_len3,key3,exp_len3,exp3) && ok;
+  ok = print_test2s(in_len4,in4,key_len4,key4,exp_len4,exp4) && ok;
+  ok = print_test2s(in_len5,in5,key_len5,key5,exp_len5,exp5) && ok;
```
msprotz, line 104: minor remark: you could do
```
bool ok = true;
for (int i = 0; i < vectors2b/sizeof(blake2b_test_vector); ++i) {
  ok &= print_test2b(vectors2b[i].input, ...);
}
// same for vectors2s
```

that way if someone adds more test vectors, no need to change the test for them to be taken into account

tests/blake2_vectors.h
```diff
@@ -0,0 +1,322 @@
...
+} blake2b_test_vector;
+
+typedef struct {
+  uint8_t *input;
+  size_t input_len;
+  uint8_t *key;
+  size_t key_len;
+  uint8_t *expected;
+  size_t expected_len;
+} blake2s_test_vector;
```
msprotz, line 24: why not share the two struct definitions?

tests/blake2_vectors.h
```diff
@@ -0,0 +1,322 @@
+#pragma once
+
+// This is cleaner than defining things in the main function, and also allows
+// for-looping over the vectors right than copy-pasting N test bodies.
+//
+// This can be easily moved to a header to be shared between blake2 tests to
```
msprotz, line 6: this comment should be updated (it has now been moved into a header)

tests/test_helpers.h
```diff
@@ -21,3 +21,34 @@ static inline bool compare_and_print(size_t len, uint8_t* comp, uint8_t* exp) {
...
+}
+
+static __inline__ cycles cpucycles_end(void)
+{
+  uint64_t rax,rdx,aux;
+  asm volatile ( "rdtscp\n" : "=a" (rax), "=d" (rdx), "=c" (aux) : : );
+  return (rdx << 32) + rax;
+}
+
+static inline void print_time(uint64_t count, clock_t tdiff, cycles cdiff){
```
msprotz, line 43: why the copy-paste of a definition here? sounds like print_time doesn't need to be behind an ifdef

PR #295, opened 2020-05-02
Protect against spurious dist refreshes and also remove some old stuff from the ci script

msprotz:
There was a bad CI nightly job that errored out with:

[STDERR]/bin/bash: /usr/bin/time: No such file or directory

but resulted in a refresh of the dist directory that removed everything because errors weren't caught. This PR tries to fix it while removing some old stuff that was lying around.

Adding Chris: let us know if someone still uses vale-only CI on the vale branch.


parno:
I'm not using them at this point.

PR #298, opened 2020-05-08
OCaml API: Don't run unit tests which require unsupported features 

victor-dumitrescu:
* Small change to the OCaml interface of EverCrypt.AutoConfig2
* Don't run some of the HACL* unit tests for the OCaml bindings if CPU doesn't support required features


msprotz:
Spurious build failure on Windows. Let's merge this.

msprotz:
Thanks Benjamin, thanks Marina!

PR #299, opened 2020-05-09
Fix the nightly build so that it keeps extracted C tests under version control

msprotz:
This wasn't caught by yesterday's PR, and is currently breaking the build...


msprotz:
This is because Travis runs extracted C tests *and* hand-written C tests as part of its "lightweight CI".

PR #300, opened 2020-05-09
Extend EverCrypt CPU detection to support avx512

R1kM:
This PR uses Vale to detect whether an Intel x64 processor supports the avx512 extensions needed for HACL intrinsincs.
It checks whether AVX512VL + AVX512F + AVX512DQ + AVX512BW are enabled.


providers/evercrypt/EverCrypt.AutoConfig2.fsti
```diff
@@ -31,6 +31,7 @@ val has_adx: getter Vale.X64.CPU_Features_s.adx_enabled
 val has_sse: getter Vale.X64.CPU_Features_s.sse_enabled
 val has_movbe: getter Vale.X64.CPU_Features_s.movbe_enabled
 val has_rdrand: getter Vale.X64.CPU_Features_s.rdrand_enabled
+val has_avx512: getter Vale.X64.CPU_Features_s.avx512_enabled
```
msprotz, line 34: a comment indicating which of the avx512 flags you're currently testing for would be great... because we'll need this comment to be in sync with the compiler flags passed in the makefile

R1kM: Good point, this is specified in CPU_Features_s, but I added a comment in Autoconfig. Thanks!

PR #302, opened 2020-05-11
Basic building blocks for adding proper treatment to zero-out memory

msprotz:
@Kachoc as promised, here's the implementation with the delicate bits taken care of... there's only a couple things left to fix, per our earlier discussion:
- rewire the functions in lib/c/Lib_Memzero.c to call the implementation in lib/c/Lib/Memzero0.c
- fix the latter to use the various implementations described in FStarLang/kremlin#146, including memset_s for OSX which appears to require `#define __STDC_WANT_LIB_EXT1__ 1` followed by `#include <string.h>`

Thanks,

Jonathan


msprotz:
Please see commit message in kremlin for the details of what was needed on the kremlin side: https://github.com/FStarLang/kremlin/commit/05a25d74e2940985ccb3e4d4573523bf9a115ad9

msprotz:
Some more details:
- the reason why I had to write Lib_Memzero0.ml is that some specs in ECDSA, which are under the `Spec.` prefix, get extracted to ML (everything under `Spec` does), but somehow require `Lib.Buffer`, so this means that even though Lib.Buffer is supposed to be Low\* code and specs are supposed to be high-level code, there's still a dependency arrow from ML-extracted spec to Lib.Buffer, and because I fixed `salloc` to call Lib.Memzero0, then this file needs an ML version

To be reviewed once @Kachoc pushes a couple more commits on this branch to add proper C calls.

msprotz:
I've guarded memset_s behind OSX because I'm assuming that if you're on OSX, your toolchain supports memset_s. This is true of the configurations we test on Travis (10.13 XCode 10 and 10.14 XCode 11). https://developer.apple.com/documentation/kernel/2876438-memset_s?language=objc indicates 10.13+

Ideally we'd have a proper configure script that does actual feature detection -- this would be the best solution, but requires time, trial and error, and possibly getting into autoconf/automake. I've done a substantial amount of it already for:
- working around buggy XCode 10.1
- working around buggy GCC
- proper #ifdef-ing of subborrow depending on GCC version and not for clang
- etc.
so unfortunately my motivation is low to write a configure detection step. (This stuff takes days and days to get right.) My thinking is therefore:
- assume RtlZeroMemory (windows)
- assume memset_s (OSX)
- assume explicity_bzero (linux, freebsd)
- assume explicit_memset (netbsd)
all via #ifdef detection, and #warning otherwise.

msprotz:
Also note that `memset_s` is a function of your `libc` too, so then it depends on what your libc developers think of C11 annex K. On Linux, they've explicitly said no, see https://patchwork.ozlabs.org/project/glibc/patch/CAKDKvuzWYf3GcXYs4ED8XLyy58nzmvxRV84xwsKKZjPpVSFQug@mail.gmail.com/

msprotz:
Build failure on Windows:
```
[STDERR]Lib_Memzero0.c:28:5: error: implicit declaration of function ‘SecureZeroMemory’ [-Werror=implicit-function-declaration]
[STDERR]   28 |     SecureZeroMemory(pnt, len);
[STDERR]      |     ^~~~~~~~~~~~~~~~
[STDERR]Lib_Memzero0.c:28:22: error: ‘pnt’ undeclared (first use in this function); did you mean ‘int’?
[STDERR]   28 |     SecureZeroMemory(pnt, len);
[STDERR]      |                      ^~~
[STDERR]      |                      int
```

PR #303, opened 2020-05-11
OCaml API: Various fixes

victor-dumitrescu:
* fixes and cleanup in the build process, notably invoking dist/gcc-compatible/configure when installing hacl-star-raw
* changes to CPU features runtime checks, which previously made the whole build of hacl-star fail when a feature required by any primitive was not present


bindings/ocaml/EverCrypt.ml
```diff
@@ -111,12 +111,14 @@ end
 
 module Chacha20_Poly1305 : Chacha20_Poly1305 =
   Make_Chacha20_Poly1305 (struct
+    let reqs = []
```
msprotz, line 114: I would add a comment here saying that run-time checks are already performed by the evercrypt API and that as such the OCaml bindings layer does not need to check for anything

PR #304, opened 2020-05-12
OCaml API: Fix for opam build

victor-dumitrescu:
Small change to the way in which hacl-star-raw.opam is generated


tools/Makefile.opam
```diff
@@ -31,8 +31,9 @@ release-hacl-star: prepare-raw
 	mv raw hacl-star/
 	mv hacl-star/raw/hacl-star-raw.opam hacl-star/
 	sed -i 's/make/make "-C" "raw"/g' hacl-star/hacl-star-raw.opam
```
msprotz, line 33: please use $(SED) -- your usage of `sed -i` is non-portable and assumes GNU sed

msprotz: see the top-level Makefile for how this is defined, and possibly move the auto-detection to Makefile.include or something so that it can be shared

victor-dumitrescu: Given that this Makefile is only ever expected to be manually run when submitting to opam, is this worth changing?

msprotz: it just means I can't run it on my laptop (osx) so I'm taking this as a firm commitment that you'll be the one taking care of submitting any further updates to OPAM :)

PR #305, opened 2020-05-12
CI: do not hardcode gcc-7

tahina-pro:
No longer useful now with the base Linux CI container using gcc-9. This change was suggested by @msprotz 

I claim that this is the only change needed to make HACL* Linux CI work with the new Linux CI base image based on Ubuntu 20.04, where the default gcc is now gcc-9


msprotz:
actually it looks like the base image *has* propagated, so we should be able to merge this soon... thanks!

tahina-pro:
yes, I waited for the base image to propagate before opening this PR

PR #306, opened 2020-05-13
Proof tweak for some smt pat changes in F* ulib

aseemr:
We are in the process of cleaning up F* ulib smt patterns to remove the 271 (theory symbols appearing in patterns) warning. I just pushed to F* master changes in List and Seq. This caused one non-linear arithmetic assertion in Lib.Buffer.fst to fail. This PR adds a restart-solver around that proof to make it work.


msprotz:
Closing because this change was cherry-picked while merging another PR. Thanks Aseem!

PR #308, opened 2020-05-25
AEAD encryption/decryption with key expansion on the stack

Kachoc:
For AEAD, the persistent state can be a burden to carry around, and it is desirable to have encryption/decryption functions which perform key expansion and use the stack to allocate the state (not the heap).
This merge is about adding a `StackInline` version of `create_in` as well as the appropriate encryption/decryption functions.


Kachoc:
For review: @msprotz

PR #309, opened 2020-05-31
Fix the build on ARM

msprotz:
This is a followup to @Kachoc's earlier pull request, where in the review I failed to notice that this was breaking the build on ARM. (I noticed that the nightly builds were no longer regenerating the snapshot, and did some investigation today.)

I documented the issue as comments; essentially, the one-shot encrypt/decrypt functions for AES-GCM contained unguarded references to the X64 ASM routines, meaning that on the ARM build there was an unresolved linker reference.

My fix was to make these functions perform proper run-time detection. Since the run-time CPU feature flags are partial, and never give you the absence of CPU features, just their presence (if init has been called), there is no way to eliminate the UnsupportedAlgorithm case statically. So, I made two "no_check" variants which never return UnsupportedAlgorithm at the expense of them being inline_for_extraction noextract.


providers/evercrypt/EverCrypt.AEAD.fsti
```diff
@@ -329,19 +337,50 @@ let encrypt_expand_st (a: supported_alg) =
...
+      match r with
+      | Success ->
+          B.(modifies ((loc_union (loc_buffer cipher) (loc_buffer tag))) h0 h1) /\
+          S.equal (S.append (B.as_seq h1 cipher) (B.as_seq h1 tag))
+            (Spec.encrypt #a (B.as_seq h0 k) (B.as_seq h0 iv) (B.as_seq h0 ad) (B.as_seq h0 plain))
+      | UnsupportedAlgorithm ->
+          if does_runtime_check then
+            B.(modifies loc_none h0 h1)
+          else
+            false
```
R1kM, line 357: Minor: Shouldn't this be False?

providers/evercrypt/EverCrypt.AEAD.fsti
```diff
@@ -420,7 +461,10 @@ let decrypt_expand_st (a: supported_alg) =
   dst: B.buffer uint8 { B.length dst = B.length cipher } ->
   Stack error_code
     (requires fun h0 ->
-        config_pre a /\
+        (if does_runtime_check then
+          true
```
R1kM, line 465: Minor: Shouldn't this be True?

providers/evercrypt/EverCrypt.AEAD.fsti
```diff
@@ -437,20 +481,30 @@ let decrypt_expand_st (a: supported_alg) =
         Some? plain /\ S.equal (Some?.v plain) (B.as_seq h1 dst)
       | AuthenticationFailure ->
         None? plain
+      | UnsupportedAlgorithm ->
+          if does_runtime_check then
+            B.(modifies loc_none h0 h1)
+          else
+            false
```
R1kM, line 488: Same as above

PR #310, opened 2020-06-01
Make the opam packages work on ARM

victor-dumitrescu:
This pull request makes the `hacl-star-raw` and `hacl-star` opam packages build on ARM by:
* Having the `configure` script write a config.h which indicates whether x86 intrinsics are available and whether it is running on ARM 7/8 (thanks @msprotz)
* Excluding all of the auto-generated bindings for .c files that cannot be compiled from `hacl-star-raw`
* In `hacl-star`, using `cppo` to preprocess `Hacl.ml/mli` to exclude the API for primitives which have not been compiled and have no corresponding binding. Same for tests.


bindings/ocaml/Hacl.ml
```diff
@@ -1,32 +1,36 @@
...
-module Hacl_Poly1305_256 = Hacl_Poly1305_256_bindings.Bindings(Hacl_Poly1305_256_stubs)
 module Hacl_HKDF = Hacl_HKDF_bindings.Bindings(Hacl_HKDF_stubs)
 module Hacl_NaCl = Hacl_NaCl_bindings.Bindings(Hacl_NaCl_stubs)
 module Hacl_Blake2b_32 = Hacl_Blake2b_32_bindings.Bindings(Hacl_Blake2b_32_stubs)
-module Hacl_Blake2b_256 = Hacl_Blake2b_256_bindings.Bindings(Hacl_Blake2b_256_stubs)
 module Hacl_Blake2s_32 = Hacl_Blake2s_32_bindings.Bindings(Hacl_Blake2s_32_stubs)
-module Hacl_Blake2s_128 = Hacl_Blake2s_128_bindings.Bindings(Hacl_Blake2s_128_stubs)
 module Hacl_ECDSA = Hacl_ECDSA_bindings.Bindings(Hacl_ECDSA_stubs)
 
+#ifndef BROKEN_INTRINSICS
```
msprotz, line 23: so I would actually use a fresh define, such as "NO_VEC_256". BROKEN_INTRINSICS is actually orthogonal to whether 128- and 256-bit algorithms are buildable... it's just a switch to use X86 intrinsics for addcarry/subborrow, or to use C implementations

whether 256-bit and 128-bit versions are available is another dimension.
- arm7: none are available
- arm8: 128-bit only

so I would make sure the configure script writes out one or two of these #defines when detecting ARM

bindings/ocaml/Hacl.ml
```diff
@@ -341,3 +299,51 @@ module ECDSA = struct
...
+  end)
+
+module Chacha20_Poly1305_256 : Chacha20_Poly1305 =
+  Make_Chacha20_Poly1305 (struct
+    let reqs = [AVX2]
+    let encrypt = Hacl_Chacha20Poly1305_256.hacl_Chacha20Poly1305_256_aead_encrypt
+    let decrypt = Hacl_Chacha20Poly1305_256.hacl_Chacha20Poly1305_256_aead_decrypt
+  end)
+
+module Curve25519_64 : Curve25519 =
```
msprotz, line 318: can you reuse evercrypt_targetconfig.h for this module and guard the definition of curve64 behind EVERCRYPT_TARGETCONFIG_IS_X64?

bindings/ocaml/SharedFunctors.ml
```diff
@@ -113,7 +113,7 @@ module Make_EdDSA_generic (C: Buffer)
     (* Hacl.Ed25519.sign_expanded *)
     assert (C.size ks = 96);
     assert (C.size signature = 64);
-    assert (C.size msg + 64 <= max_size_t);
+    assert Z.(of_int (C.size msg) + ~$64 <= max_size_t);
```
msprotz, line 116: I suspect these overflow on 32-bit ARM. Have you gone through a general cleanup pass throughout all the bindings?

dist/configure
```diff
@@ -171,8 +171,8 @@ fi
 if ! detect_x64; then
   echo "$(uname -m) does not support x64 assembly, disabling Curve64"
   echo "BLACKLIST += Hacl_Curve25519_64.c $(ls Hacl_HPKE_Curve64_*.c | xargs)" >> Makefile.config
-  echo "$(uname -m) does not support _addcarry_u64, disabling ECDSA"
-  echo "BLACKLIST += Hacl_ECDSA.c $(ls Hacl_HPKE_P256_*.c | xargs)" >> Makefile.config
+  echo "$(uname -m) does not support _addcarry_u64, using a C implementation"
+  echo "#define BROKEN_INTRINSICS 1" >> config.h
```
msprotz, line 175: this change is good

bindings/ocaml/Hacl.ml
victor-dumitrescu: I gave it a go but the format of evercrypt_targetconfig.h doesn't work with cppo. I made the configure script write IS_X64 to config.h, but this does exclude the `defined(_M_X64)` check.

bindings/ocaml/SharedFunctors.ml
victor-dumitrescu: Yes, that was the reason I changed it throughout the bindings.

bindings/ocaml/Hacl.ml
msprotz: maybe you could write NOT_X64 so that the default remains the same and people can compile on windows without the need to run configure

victor-dumitrescu: So if i understand correctly this would actually be the same thing as BROKEN_INTRINSICS, which is written in `if ! detect_x64; then`

victor-dumitrescu: (Actually no sorry that's not exactly true since BROKEN_INTRINISICS is also defined when `! check_no_bug81300`)

bindings/ocaml/Hacl.ml
```diff
@@ -1,32 +1,38 @@
...
-module Hacl_Poly1305_256 = Hacl_Poly1305_256_bindings.Bindings(Hacl_Poly1305_256_stubs)
 module Hacl_HKDF = Hacl_HKDF_bindings.Bindings(Hacl_HKDF_stubs)
 module Hacl_NaCl = Hacl_NaCl_bindings.Bindings(Hacl_NaCl_stubs)
 module Hacl_Blake2b_32 = Hacl_Blake2b_32_bindings.Bindings(Hacl_Blake2b_32_stubs)
-module Hacl_Blake2b_256 = Hacl_Blake2b_256_bindings.Bindings(Hacl_Blake2b_256_stubs)
 module Hacl_Blake2s_32 = Hacl_Blake2s_32_bindings.Bindings(Hacl_Blake2s_32_stubs)
-module Hacl_Blake2s_128 = Hacl_Blake2s_128_bindings.Bindings(Hacl_Blake2s_128_stubs)
 module Hacl_ECDSA = Hacl_ECDSA_bindings.Bindings(Hacl_ECDSA_stubs)
 
+#ifndef IS_ARM_7
```
msprotz, line 23: this disables the bindings on X64, doesn't it? can you #ifdef it behind

#if not IS_NOT_X64 || IS_ARMV8

?

dist/Makefile.tmpl
```diff
@@ -85,9 +85,13 @@ Hacl_Blake2b_256.o Hacl_Poly1305_256.o Hacl_Chacha20_Vec256.o Hacl_Chacha20Poly1
 
 all: libevercrypt.$(SO)
 
-libevercrypt.$(SO): $(OBJS)
+libevercrypt.$(SO): config.h $(OBJS)
 	$(CC) $(CFLAGS) -shared -o $@ $^ $(LDFLAGS)
 
+# If the configure script has not ran, create an empty config.h
```
msprotz, line 91: typo: run?

.travis.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         - opam init --yes --bare
```
msprotz, line 16: here's one thing I just thought of: why not enable travis CI for the OCaml bindings in ARM as well?

PR #314, opened 2020-06-11
sha2-mb

polubelova:
this PR introduces Multi-Buffer SHA2 implementation (Scalar, AVX, AVX2)


prosecco:
Let's rename the folder to code/sha2-mb

code/sha2-vec/Hacl.Impl.SHA2.Generic.fst
```diff
@@ -0,0 +1,409 @@
...
+module SpecVec = Hacl.Spec.SHA2.Vec
+module VecTranspose = Lib.IntVector.Transpose
+module LSeq = Lib.Sequence
+
+
+#set-options "--max_fuel 0 --max_ifuel 0 --z3rlimit 50"
+
+(** Top-level constant arrays for the SHA2 algorithms. *)
+
+let h224 : x:glbuffer uint32 8ul{witnessed x Constants.h224 /\ recallable x} =
```
msprotz, line 32: any way we can share this with the old-sha2 and put it into a static header?

code/sha2-vec/Hacl.Spec.SHA2.EquivScalar.fst
```diff
@@ -0,0 +1,933 @@
...
+  assert (hash_length a == word_length a * hash_word_length a);
+
+  let aux (i:nat{i < hash_length a}) : Lemma ((finish a st).[i] == (PadFinish.finish a st).[i]) =
+    BSeq.index_uints_to_bytes_be #(word_t a) #SEC hash_final_w i;
+    BSeq.index_uints_to_bytes_be #(word_t a) #SEC #8 st i in
+
+  Classical.forall_intro aux;
+  eq_intro (finish a st) (PadFinish.finish a st)
+
+//TODO: move to Lib.Sequence.Lemmas
```
msprotz, line 608: aren't these somewhere already?

code/sha2-vec/Hacl.Spec.SHA2.fst
```diff
@@ -0,0 +1,306 @@
...
+
+open FStar.Mul
+open Lib.IntTypes
+open Lib.Sequence
+module C = Spec.SHA2.Constants
+module S = FStar.Seq
+
+open Spec.Hash.Definitions
+
+#set-options "--z3rlimit 20 --max_fuel 0 --max_ifuel 0"
```
msprotz, line 11: this seems to be reusing a lot of definitions already present elsewhere

Kachoc:
Very neat code indeed! I agree with @protz  about trying not to duplicate code with the old SHA.
The most important remarks have already been made but I have a few (minor) comments.

This is more for information, because so far I don't see how I could ask for @polubelova to modify her code to account for my future PR, but I noticed there will likely be collisions with the work started by @R1kM on incremental blake and which I took on. So let's say we'ill probably have to coordinate a bit to minimize the pain of merging those branches (or put differently, once this branch gets merged into master, I'll probably have a few questions to prepare my PR - if as much code as possible is factorized, as suggested @protz , it will probably help me a lot, though).
For instance, there is [Hacl.Spec.SHA2.Equiv.fst](https://github.com/project-everest/hacl-star/blob/e260a1dbfe36cbb5ef95f2672a8f91d123bab17e/code/sha2-vec/Hacl.Spec.SHA2.Equiv.fst) which will (I think) overlap with [Spec.Hash.Incremental.Lemmas.fst](https://github.com/project-everest/hacl-star/blob/son_blake/specs/lemmas/Spec.Hash.Incremental.Lemmas.fst). Some other definitions like [update_last](https://github.com/project-everest/hacl-star/blob/e260a1dbfe36cbb5ef95f2672a8f91d123bab17e/code/sha2-vec/Hacl.Spec.SHA2.fst#250) will probably overlap too. More generally, the last block size is computed differently in the incremental blake branch, to make sure there is always some data to give to `update_last`. Still, I think the overlapping is definitely bearable on my side.

Another small remark which is not specifically about this branch but which I drop here so that it may be heard and cause a change in programming practices: I think it is not a good idea to declare long sequences definitions like the ones from [Spec.SHA2.Constants.fst](https://github.com/project-everest/hacl-star/blob/karthik_sha2_vec/specs/Spec.SHA2.Constants.fst) as `inline_for_extraction`. The reason is that at extraction time, many spec functions which more or less directly reference those definitions and are present in assertions in the code get expanded, causing significant blowups for the normalizer (I encountered the issue for blake). I'd rather advocate for using `normalize` (or else) at the exact place where they actually need to be expanded (i.e.: where we allocate the arrays in the code).

msprotz:
Thanks @Kachoc -- regarding the list issue, is there any chance you could push a commit on this PR to establish a style, with comments on how to do it in a more optimal way in the future? Even just one instance might be enough to make sure future code is written in the way your propose. (Or just copy-paste a diff here or on slack.) Thanks for looking at this!

Kachoc:
@protz regarding the list issue, for Blake I remove the `inline_for_extraction` from the `list_sigma` definition in `Spec.Blake2.fst`:
```
noextract
let list_sigma: list_sigma_t =
  [@inline_let]
  let l : list sigma_elt_t = [
    size  0; size  1; size  2; size  3; size  4; size  5; size  6; size  7;
    (* ... *)
    ] in
  assert_norm(List.Tot.length l == 160);
  l
```

And I updated the buffer definition in `Hacl.Impl.Blake2.Constants.fst` to look like that:

```
let sigmaTable : x:glbuffer Spec.sigma_elt_t 160ul{witnessed x Spec.sigmaTable /\ recallable x} =
  (**) norm_spec [delta_only [`%Spec.list_sigma]] Spec.list_sigma;
  createL_global (norm [delta_only [`%Spec.list_sigma]] Spec.list_sigma)
```

I think the `delta_only` is unnecessarily precise (on such a definition we could probably just use `normalize`)

Kachoc:
Update on the list issue: inserting manual calls to `norm` and its derivatives actually doesn't work, because they overwrite the normalization parameters used by Kremlin at extraction time and thus prevent proper inlining of the definitions tagged `inline_for_extraction`. The correct method is to use post-processing (apply the normalization to update the definition body right after it has been defined):
```
noextract
let pp_sigmaTable () : Tactics.Tac unit =
  Tactics.norm [delta_only [`%Spec.list_sigma]]; Tactics.trefl ()

[@(Tactics.postprocess_with pp_sigmaTable)]
let sigmaTable : x:glbuffer Spec.sigma_elt_t 160ul{witnessed x Spec.sigmaTable /\ recallable x} =
  createL_global Spec.list_sigma
```

karthikbhargavan:
Let us merge this. We can then address the concerns in another PR that integrates sha2-vec with the Agile API.

dist/gcc-compatible/Hacl_SHA2_Vec256.h
```diff
@@ -0,0 +1,102 @@
...
+K____uint8_t___uint8_t_;
+
+typedef struct K____uint8_t__K____uint8_t___uint8_t__s
+{
+  uint8_t *fst;
+  K____uint8_t___uint8_t_ snd;
+}
+K____uint8_t__K____uint8_t___uint8_t_;
+
+typedef struct K____uint8_t__K____uint8_t__K____uint8_t___uint8_t__s
```
beurdouche, line 52: @polubelova Can you do something about these?

dist/gcc-compatible/libintvector.h
```diff
@@ -290,6 +310,10 @@ typedef __m256i Lib_IntVector_Intrinsics_vec256;
 #define Lib_IntVector_Intrinsics_vec256_rotate_right64_56(x0) \
   (_mm256_shuffle_epi8(x0, _mm256_set_epi8(14,13,12,11,10,9,8,15,6,5,4,3,2,1,0,7,14,13,12,11,10,9,8,15,6,5,4,3,2,1,0,7)))
 
+#if !defined(__clang__) && defined(__AVX512VL__)
```
beurdouche, line 313: Is that really clang specific?

lib/Lib.IntVector.fst
```diff
@@ -473,16 +552,21 @@ let vec_load_le t w b =
...
 
 let vec_load_be t w b =
   match t,w with
   | U128,1 -> vec128_load_be b
   | _,1 -> Lib.ByteBuffer.uint_from_bytes_be #t #SEC b
   | U32,4 -> vec128_load32_be b
   | U64,2 -> vec128_load64_be b
+  | U32,8 -> vec256_load32_be b
+  | U64,4 -> vec256_load64_be b
+  | U128,2 -> admit()//vec256_load_be b
```
beurdouche, line 566: Why are those admitted?

specs/Spec.SHA2.fst
```diff
@@ -176,10 +176,10 @@ let shuffle_core_pre = shuffle_core_pre_
 (* Scheduling function *)
 
 (* Incremental Version *)
-let ws_pre_inner (a:sha2_alg) (block:block_w a) (i:nat{i < size_k_w a}) (ws:k_w a) : k_w a =
-    if i < block_word_length then
+let ws0_pre_inner (a:sha2_alg) (block:block_w a) (i:nat{i < block_word_length}) (ws:k_w a) : k_w a =
```
beurdouche, line 179: Argh, is this split really necessary? If it really is, no worries... :)

beurdouche (approved):

lib/Lib.MultiBuffer.fst
```diff
@@ -0,0 +1,92 @@
...
+  live h b0 /\ live h b1 /\ live h b2 /\ live h b3
+
+let live8 #a #len (h:mem) (b0 b1 b2 b3 b4 b5 b6 b7: lbuffer a len) =
+  live h b0 /\ live h b1 /\ live h b2 /\ live h b3 /\ live h b4 /\ live h b5 /\ live h b6 /\ live h b7
+
+let internally_disjoint4 #len #a (b0 b1 b2 b3: lbuffer a len) =
+  disjoint b0 b1 /\ disjoint b0 b2 /\ disjoint b0 b3 /\
+  disjoint b1 b2 /\ disjoint b1 b3 /\ disjoint b2 b3
+
+let internally_disjoint8 #len #a (b0 b1 b2 b3 b4 b5 b6 b7: lbuffer a len) =
```
msprotz, line 25: FYI there's a library to do this automatically rather than write all pairwise combinations by hand: https://github.com/FStarLang/FStar/blob/master/ulib/FStar.BigOps.fsti -- see https://github.com/FStarLang/FStar/blob/master/ulib/LowStar.Monotonic.Buffer.fsti#L1004 for an example of usage

lib/Lib.Sequence.fst
```diff
@@ -365,3 +365,67 @@ let rec index_generate_blocks #t len max n f i =
...
+  assert_norm (s.[5] == x5);
+  assert_norm (s.[6] == x6);
+  assert_norm (s.[7] == x7)
+
+let create16 #a x0 x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 x11 x12 x13 x14 x15 =
+  let l = [x0; x1; x2; x3; x4; x5; x6; x7; x8; x9; x10; x11; x12; x13; x14; x15] in
+  assert_norm (List.Tot.length l = 16);
+  createL l
+
+let create16_lemma #a x0 x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 x11 x12 x13 x14 x15 =
```
msprotz, line 414: same remark here, you can meta-program those:
https://github.com/FStarLang/FStar/blob/master/ulib/FStar.Seq.Properties.fsti#L666
is something I wrote in support of this style, although the line number may lead you to believe this technique is cursed

msprotz (approved):
This looks good to me, minor the two little remarks I had -- happy to help with those if you're interested in having a solution that works once and for all, rather than having to write disjoint32 very soon ;-).

Thanks for moving the constants into a static header, this will probably do a lot for performance.

I have two requests.
- Can you open a bug and list precisely in a bullet-point style what remains to be done to get rid of the old implementation? That way we'll know where to start once someone has time to get rid of the old stuff. As you saw trying to merge this again, it's hard to get back into some old piece of code you haven't touched in a while. So, a description of "next tasks" would be good. I'm happy to add my own list of tasks to the bug once you're done with yours.
- I would like to see even a brief paragraph describing this new API. The argument names to your function are two-letter long (e.g. `b1`) and I actually doubt that anyone will figure out how this API works on their own. It'd be a shame if no one used your great work because they can't figure out how to use the API! So can you please add a word to the documentation? https://hacl-star.github.io/HaclHash.html -- it should be as simple as modifying this: https://github.com/project-everest/hacl-star/blob/master/doc/HaclHash.rst

Happy to help with any of these tasks, please let me know. Congratulations again on a really nice piece of work!

karthikbhargavan (approved):

karthikbhargavan:
Marina, let's address Jonathan's 2 requests (open a bug, and document the API) together. I can help next week.

Makefile.common
```diff
@@ -98,6 +98,7 @@ CURVE_BUNDLE=-bundle Hacl.Curve25519_64=Hacl.Impl.Curve25519.Field64.Vale \
   $(CURVE_BUNDLE_BASE) -bundle Hacl.Curve25519_64_Local
 
 HASH_BUNDLE=-bundle Hacl.Hash.MD5+Hacl.Hash.Core.MD5+Hacl.Hash.SHA1+Hacl.Hash.Core.SHA1+Hacl.Hash.SHA2+Hacl.Hash.Core.SHA2+Hacl.Hash.Definitions=Hacl.Hash.*[rename=Hacl_Hash]
+SHA2VEC_BUNDLE=-bundle Hacl.Impl.SHA2.Generic=Hacl.Impl.SHA2.*,Hacl.Hash.*[rename=Hacl_SHA2_Generic] -static-header Hacl.Impl.SHA2.Generic
```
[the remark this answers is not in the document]
polubelova: you are right. i've removed Hacl.Hash.* from sha2vec_bundle

dist/gcc-compatible/libintvector.h
polubelova: removed from this PR

lib/Lib.Sequence.fst
polubelova: thanks for the link! it works like a charm.

polubelova:
thanks a lot for all the comments and suggestions!

@karthikbhargavan was working on the Low* implementation of sha2-mb, maybe he can address some of the above comments.

Karthik, sure, we can discuss it next week.

prosecco:
shall we merge this?

msprotz:
If you document the API and list the remaining items in a bug, you have my blessing! :-)

karthikbhargavan:
Protz want a documented API which should not be hard to add. Bitrot is a bigger concern.

msprotz:
I'm strongly in favor of merging this, but please write a minimum amount of documentation and list the remaining items, otherwise the knowledge will be lost and we'll totally struggle to remember what needs to be done to clean up. There will be quite a bit of duplication after this merge, so I want to make sure the issues are properly documented.

I can help with listing the remaining issues (brief call together?) and the documentation shouldn't take more than 10 minutes. Not something exhaustive, just a new page in the doc that explains what multi-buffer is, where to find it, and how to use it.

Thanks.

PR #319, opened 2020-06-26
Update Supported.rst (P256, Blake2)

franziskuskiefer:
It appears that P256 and blake2 are on master.


msprotz:
Thanks Franziskus!

Issue #316, opened 2020-06-21
MSVC build failed


sopium:
* MSVC does not have `x86intrin.h`. Use `intrin.h` instead?

  https://github.com/project-everest/hacl-star/blob/24b295a645bd8462d20e2c2db1cb8cd303aa781d/dist/msvc-compatible/lib_intrinsics.h#L10

* `low` and `high` macros leak from here:

  https://github.com/project-everest/hacl-star/blob/24b295a645bd8462d20e2c2db1cb8cd303aa781d/dist/kremlin/kremlib/dist/minimal/fstar_uint128_msvc.h#L33-L34

   making this fail to compile:

  https://github.com/project-everest/hacl-star/blob/24b295a645bd8462d20e2c2db1cb8cd303aa781d/dist/msvc-compatible/Hacl_ECDSA.c#L738

msprotz:
Ok for high/low, thanks -- our Windows CI has been broken for a couple weeks and this flew under the radar.

For x86intrin.h, I could've sworn it worked on MSVC back when we had CI. Would be able to confirm that this: https://stackoverflow.com/questions/11228855/header-files-for-x86-simd-intrinsics is right and that including `immintrin.h` works everywhere?

msprotz:
I did a little bit of archaeology and indeed we weren't compiling Hacl_ECDSA in our MSVC test scenario, even back when Windows CI was working. We should probably have a light travis run to make sure dist/msvc-compatible always compiles on MSVC.

If you have time, a PR on https://github.com/FStarLang/kremlin/ to add an `#undef low #undef high` at the end of fstar_uint128_msvc.h would be very helpful (and would be propagated to hacl-star via the nightly build).

If you have more time, and can confirm #include immintrin.h works, I'd be more than happy to approve a pull request.

Finally, if you have even more time, a PR to add a quick and dirty Travis windows run (e.g. cd dist/msvc-compatible && cl.exe /I ../kremlin/include /I ../kremlin/kremlib/dist/minimal) to prevent that sort of errors in the future would be make both your life and ours easier I suspect :)

sopium:
I did some quick tests, and `immintrin.h` works for MSVC (and clang) but not for GCC < 9, because it does not include `_addcarry_u64` and `_subborrow_u64` then:

https://gcc.godbolt.org/z/G1ei02

msprotz:
@Frosne has expressed an interest in giving some Windows testing & CI a go.

Perhaps
```
[...]

PR #323, opened 2020-07-06
Lightweight Windows CI

msprotz:
This compiles the generated code with MSVC, something we've been badly missing for a while.

The way to get this to work was found by searching around for Travis/MSVC examples and it seems to work quite well, modulo the usual error-return-code issues between msys shell and native windows programs.

Note that this barely compiles the files, and does not attempt to link them or run tests. This can be performed later by someone else.


msprotz:
Note that of course I had to fix #316 to get a green on Windows, which was the whole point of doing this.

msprotz:
Nudged by @polubelova I actually added compilation of the MSVC assemblies with the Microsoft assembler, along with linking everything together (but I'm not sure this tests much).

PR #324, opened 2020-07-07
OCaml API: Minor build fix, new release

victor-dumitrescu:
This PR fixes a small bug in the Makefile in dist/gcc-compatible which resulted in the bytecode archive not being picked up when installing `hacl-star-raw`.

### DIFF
--- a/dist/gcc-compatible/Makefile
+++ b/dist/gcc-compatible/Makefile
@@ -191,7 +191,7 @@ ocamlevercrypt.cma: $(ALL_BINDINGS) $(ALL_ML_STUBS_CMO) $(CTYPES_CMO) $(ALL_C_ST
 	ocamlfind mklib -o ocamlevercrypt $(CTYPES_CMO) $(ALL_C_STUBS) -ccopt -L. -cclib -levercrypt
 
 install-hacl-star-raw: ocamlevercrypt.cmxa ocamlevercrypt.cma libevercrypt.a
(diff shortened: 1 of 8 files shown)


dist/gcc-compatible/Makefile
```diff
@@ -191,7 +191,7 @@ ocamlevercrypt.cma: $(ALL_BINDINGS) $(ALL_ML_STUBS_CMO) $(CTYPES_CMO) $(ALL_C_ST
 	ocamlfind mklib -o ocamlevercrypt $(CTYPES_CMO) $(ALL_C_STUBS) -ccopt -L. -cclib -levercrypt
 
 install-hacl-star-raw: ocamlevercrypt.cmxa ocamlevercrypt.cma libevercrypt.a
-	ocamlfind install hacl-star-raw META $(CTYPES_ML) $(CTYPES_CMX) $(CTYPES_CMO) $(CTYPES_CMI) $(wildcard *.cmxa *.cma *.a *.so *.dll) config.h
+	ocamlfind install hacl-star-raw META $(CTYPES_ML) $(CTYPES_CMX) $(CTYPES_CMO) $(CTYPES_CMI) ocamlevercrypt.cma ocamlevercrypt.cmxa $(wildcard *.a *.so *.dll) config.h
```
msprotz, line 194: I suspect this will not work if you invoke "make install-hacl-star-raw" from a clean build because $(wildcard) will be expanded at Makefile read-time and will expand to nothing

example:

```
jonathan@absinthe:/tmp $ cat Makefile 
all:
	echo $(wildcard *.foo)
	touch bar.foo
jonathan@absinthe:/tmp $ make
echo 

touch bar.foo
jonathan@absinthe:/tmp $ make
echo bar.foo
bar.foo
touch bar.foo
```

PR #331, opened 2020-07-21
ECDSA Bug in reduction

Frosne:
This PR solves a problem with missing reduction at the end of ECDSA verification procedure, adds several wycheproof tests for DH and ECDSA and adds one more precondition for DH. Additionally, there was a change in the verification procedure. Before the function 'point_add' there was a check for point equality. If the points are identical, the point_double routine is called. 

A lot of thanks to @franziskuskiefer for enormous help with ecdsa :)


Frosne:
Additionally, there was a change in the verification procedure. Before the function 'point_add' there was a check for point equality. If the points are identical, the point_double routine is called.

specs/ecdsap256/Spec.ECDSA.fst
```diff
@@ -472,9 +472,14 @@ let ecdsa_verification_agile alg publicKey r s mLen m =
       let u1D, _ = montgomery_ladder_spec u1 (pointAtInfinity, basePoint) in
       let u2D, _ = montgomery_ladder_spec u2 (pointAtInfinity, publicJacobian) in
 
-      let sumPoints = _point_add u1D u2D in
-      let pointNorm = _norm sumPoints in
+      let sumD = if  _norm u1D =  _norm u2D then
```
msprotz, line 475: can you add a comment here or something explaining the subtlety?

code/ecdsap256/Hacl.Impl.ECDSA.P256.Verification.Agile.fst
```diff
@@ -464,7 +558,7 @@ let ecdsa_verification_core alg publicKeyBuffer hashAsFelem r s mLen m xBuffer t
   r
 
 
-[@ (Comment "  This code is not side channel resistant")]
+(* [@ (Comment "  This code is not side channel resistant")] *)
```
msprotz, line 561: did the come become side-channel resistant or is there another issue related to C comments?

code/ecdsap256/Hacl.Impl.ECDSA.P256.Verification.Agile.fst
```diff
@@ -301,22 +397,34 @@ val ecdsa_verification_step5_1:
...
   ecdsa_verification_step5_0 points pubKeyAsPoint u1 u2 tempBuffer;
   let pointU1G = sub points (size 0) (size 12) in
-  let pointU2Q = sub points (size 12) (size 12) in
-  point_add pointU1G pointU2Q pointSum buff;
-  norm pointSum pointSum buff;
+  let pointU2Q = sub points (size 12) (size 12) in 
+
+  let equalX = ecdsa_verification_step5_1 points in 
+  begin
+  match equalX with
```
msprotz, line 424: `if`?

code/ecdsap256/Hacl.Impl.ECDSA.P256.Verification.Agile.fst
```diff
@@ -273,8 +274,103 @@ let ecdsa_verification_step5_0 points pubKeyAsPoint u1 u2 tempBuffer =
...
+      let pointU2Q = gsub points (size 12) (size 12) in 
+      r = (
+	_norm (fromDomainPoint (point_prime_to_coordinates (as_seq h0 pointU1G))) =
+	_norm (fromDomainPoint (point_prime_to_coordinates (as_seq h0 pointU2Q))))
+    )
+  )
+
+let ecdsa_verification_step5_1 points = 
+  push_frame();
+    let tmp = create (size 112) (u64 0) in
```
msprotz, line 356: this is superficial, but aligning `let`s is generally considered good form (many other instances of this all throughout your code), because upon first sight one may think that they're under an `if` or some other constructor

tests/ecdhp256-test.c
```diff
@@ -89,8 +93,134 @@ int main()
...
+			fDU = success;
+		
+		if (w_vectors[i].flag != 1) 
+		{
+
+			ok = ok && ((success & fDU) == w_vectors[i].flag);
+			if (success == 0) {
+				printf("%s\n", "");
+				ok = ok && compare_and_print(32, result, w_vectors[i].sharedKey);
+				}
```
msprotz, line 118: the confusing indentation makes your code genuinely hard to follow. please your C test through `clang-format -i`

tests/ecdhp256-test.c
```diff
@@ -89,8 +93,134 @@ int main()
...
 
 
+	printf("%s\n", "Wycheproof tests ECDH: ");
+		
+	for (int i = 0 ; i< sizeof(w_vectors)/sizeof(ecdhp256_w_i); i++)
+	{
+		printf("ECDH Wycheproof Test %d ", i );
+
+		uint8_t* decompressedPoint = (uint8_t*) malloc (sizeof (uint8_t) * 64);
+		bool flagDecompress = Hacl_P256_decompression_not_compressed_form(w_vectors[i].publicKey, decompressedPoint);
```
msprotz, line 103: the documentation states that Hacl_P256_decompression_not_compressed_form returns true if and only if the decompression was successful, but the name `flagDecompress` seems to indicate something else -- let's use `ok`, or `success`, or something else

tests/ecdhp256-test.c
```diff
@@ -89,8 +93,134 @@ int main()
...
+
+		uint64_t success = Hacl_P256_ecp256dh_r(result, decompressedPoint, w_vectors[i].privateKey);
```
msprotz, line 105: let's keep in mind that C tests are where our C users are looking for inspiration when it comes to how to use our API, so I would like to set the bar pretty high

why is this `success` not chained (via `&&`) with the result of the previous call to Hacl_P256_decompression_not_compressed_form?

tests/ecdhp256-test.c
```diff
@@ -89,8 +93,134 @@ int main()
...
+
+		uint64_t fDU = 0;
+		if (flagDecompress = true)
```
msprotz, line 108: - assignment instead of comparison? this condition always evaluates to true, wasn't this flagged as a warning from the C compiler?
- what does `fDU` mean? let's use meaningful names, especially if we expect people to look at this code
- even imagining `==` instead of `=` I still don't understand the logic, please explain with a comment

tests/ecdhp256-test.c
```diff
@@ -89,8 +93,134 @@ int main()
...
+			fDU = success;
+		
+		if (w_vectors[i].flag != 1) 
+		{
+
+			ok = ok && ((success & fDU) == w_vectors[i].flag);
```
msprotz, line 114: the documentation states that `success` is non-zero if an error has occured

you now seem to be doing some bit-wise logic on the return value `success` -- are there some error codes that are encoded in the return value? what are the possible values for `w_vectors[i].flag`?

tests/ecdhp256-test.c
```diff
@@ -89,8 +93,134 @@ int main()
...
+	printf("%s\n", "Wycheproof tests ECDSA: ");
+		
+	for (int i = 0 ; i< sizeof(w_ecdsa_vectors)/sizeof(ecdsap256_w_i) ; i++)
+	{
+		printf("ECDSA Wycheproof Test %d", i );
+		
+		uint32_t mLen = w_ecdsa_vectors[i].mLen;
+		
+		static uint8_t pk_ [64] = {0};
+		for (int j = 0; j < 32; j++)
```
msprotz, line 147: consider using `memcpy` for greater clarity

tests/ecdhp256-test.c
```diff
@@ -89,8 +93,134 @@ int main()
...
+		{
+			pk_[j] = w_ecdsa_vectors[i].publicX[j];
+			pk_[j+32] = w_ecdsa_vectors[i].publicY[j];
+		}
+
+
+		bool resultVerification2 = Hacl_P256_ecdsa_verif_p256_sha2(mLen, w_ecdsa_vectors[i].message, pk_, w_ecdsa_vectors[i].r, w_ecdsa_vectors[i].s);	
+		
+		if ((w_ecdsa_vectors[i].flag == 18446744073709551615U) && (resultVerification2 == true))
```
msprotz, line 156: please don't use magic constants and #define them with a meaningful name

Frosne: All done :)

tests/ecdhp256-test.c
```diff
@@ -31,6 +35,13 @@ int main()
 {
 	uint8_t* result = (uint8_t*) malloc (sizeof (uint8_t) * 64);
 	bool ok = true;
+	uint8_t* pk = (uint8_t*) malloc (sizeof (uint8_t) * 64);
```
msprotz, line 38: where is the corresponding call to `free`?

tests/ecdhp256-test.c
```diff
@@ -31,6 +35,13 @@ int main()
...
+	uint8_t* decompressedPoint = (uint8_t*) malloc (sizeof (uint8_t) * 64);
+	uint64_t incorrect = 18446744073709551615U;
+	uint64_t correct = 0;
+
+// The start of NIST ECDH Initiator tests
```
msprotz, line 43: can you separate each flavor of tests into its own function? e.g.
```
bool test_nist() {
}

bool test_compression () {
}

...

int main () {
  if (!test_compression())
    return EXIT_FAILURE;
  if (!test_nist())
    return EXIT_FAILURE;
  return EXIT_SUCCESS;
}
```

or something like that
```

File: tests/ecdhp256-test.c [hunk 14]
```diff
@@ -89,8 +101,75 @@ int main()
...
 
+// The start of Wycherproof tests
+
+	printf("%s\n", "Wycheproof tests ECDH: ");
+		
+	for (int i = 0 ; i< sizeof(w_vectors)/sizeof(ecdhp256_w_i); i++)
+	{
+		printf("ECDH Wycheproof Test %d \n", i );
+
+		bool flagDecompressSuccessful = Hacl_P256_decompression_not_compressed_form(w_vectors[i].publicKey, decompressedPoint);
```
msprotz, line 112: can you remind me why one function returns a `bool` for success (0 == error, 1 == success) and the other one a `uint64_t` (0 == success, MAX_UINT64 == error)? I'm very concerned about C clients getting this wrong

tests/ecdhp256-test.c
```diff
@@ -31,6 +35,13 @@ int main()
 {
 	uint8_t* result = (uint8_t*) malloc (sizeof (uint8_t) * 64);
 	bool ok = true;
+	uint8_t* pk = (uint8_t*) malloc (sizeof (uint8_t) * 64);
+	uint8_t* decompressedPoint = (uint8_t*) malloc (sizeof (uint8_t) * 64);
+	uint64_t incorrect = 18446744073709551615U;
```
msprotz, line 40: no magic constants please... include `<stdint.h>` and use UINT64_MAX for readability

tests/ecdhp256-test.c
```diff
@@ -89,8 +101,75 @@ int main()
...
+	printf("%s\n", "Wycheproof tests ECDH: ");
+		
+	for (int i = 0 ; i< sizeof(w_vectors)/sizeof(ecdhp256_w_i); i++)
+	{
+		printf("ECDH Wycheproof Test %d \n", i );
+
+		bool flagDecompressSuccessful = Hacl_P256_decompression_not_compressed_form(w_vectors[i].publicKey, decompressedPoint);
+		uint64_t success = Hacl_P256_ecp256dh_r(result, decompressedPoint, w_vectors[i].privateKey);
+		
+		if (w_vectors[i].flag != 1)
```
msprotz, line 115: what does  		`if (w_vectors[i].flag != 1)` indicate? I'm not familiar with the wycheproof test data

msprotz: sometimes you do != 1 and in some other places you compare the value of the flag with UINT64_MAX or 0 so I'm not sure what this is supposed to stand for

msprotz:
something that crossed my mind: remember that your C tests are not spec tests (which have to be pure) so you're free to split your code into separate functions and do early-returns if something is wrong rather than chaining the value through `ok &&=`

franziskuskiefer:
While this is not specific to the issue I want to bring it up here. We can either work on it in this PR or make a separate one.
As I brought up already, there's no way to generate P256 keys right now. If I understand correctly, the only exposed function is `reduction_8_32` right now. But the function reduces the given random bytes to fit into the curve scalar, which is something we must not do when generating keys. To generate keys we must do some sort of rejection sampling, i.e. generate random values until we have one that's ok. Computing modulo on the random value results in non-uniform values.
I'm not sure what's the best function to expose here. But something like `isMoreThanZeroLessThanOrderMinusOne` (without the minus one) would work.

Frosne:
If we decide on one *good* style for the code API, I don´t mind changing the code to follow it.

franziskuskiefer:
I agree that return values are an issue in the API right now. Returning `boolean` is the only sensible thing to do imho. I don't see how doing the int to bool cast could be used as a side channel. There's no information there to extract (as long as the values for true and false are always the same).
So I'd support returning bools here and everywhere else for success and failure.

msprotz:
Here's a potential addition to Lib.RawIntTypes that would allow casting:

```
// Lib.RawIntTypes.fsti
val unsafe_bool_of_u64 (x: uint64 { v x == 0 \/ v x == pow2 64 - 1 }):
  (b:bool { b <==> v x == 0 })

// Lib.RawIntTypes.fst
let unsafe_bool_of_u64 = function
  | 0UL -> true
  | _ -> false
```

and if you want to minimize the amount of changes required, here's a suggestion on how to structure your post-conditions, which also allows segregating the usage of unsafe code to wrappers:

```
// In the fsti:
let ecdh_i_post x y h0 (r: uint64) h1 =
  ...

val ecdh_i: x:... -> y:... -> Stack bool (...) (fun h0 r h1 -> ecdh_i_post x y h0 (if r then u64 0 else u64 (pow2 64 - 1)) h1)

// In the fst:
let ecdh_i_old (x: ...) (y: ...): Stack bool (...) (ecdh_i_post x y)

let ecdh_i x y = Lib.RawIntTypes.unsafe_bool_of_uint64 (ecdh_i_old x y)
```

that way you don't need to copy/paste any post-condition. What do you think?

msprotz (approved):
This looks good to me! Thanks for fixing the return values, I believe it makes the API much more uniform now. As you've seen, the OCaml bindings were already doing return-value conversion so this makes things much more streamlined and easier to deal with.

Some minor points that I don't think should prevent you from merging:
- I'm still mildly concerned that the pairs of malloc/free (as opposed to stack allocation) will look weird if someone tries to take inspiration from the C tests.
- You're the only one using tabs in your tests in the tests directory, which makes your tests stand out in comparison to others. I'm happy to run `clang-format` on your test files, let me know.

I trust that Karthik has done an in-depth review of the F* code -- can you get him to do a second review of this? Or maybe ping Marina to look at your F* code?

Once you get a second review, please merge as soon as you get a green.

Thanks,

Jonathan

code/ecdsap256/Hacl.Impl.P256.LowLevel.RawCmp.fst
```diff
@@ -29,3 +30,14 @@ inline_for_extraction noextract
...
 
 let eq_0_u64 a = eq_u64_nCT a (u64 0)
+
+
+
+inline_for_extraction noextract
+val unsafe_bool_of_u64 (x: uint64 { v x == 0 \/ v x == pow2 64 - 1 }):
+  (b:bool { b <==> v x == 0 })
+
+let unsafe_bool_of_u64 x =
```
polubelova, line 40: you can write it as `FStar.UInt64.(u64_to_UInt64 x =^ 0uL)`, in this way the extracted code will look nicer.

polubelova (approved):

PR #332, opened 2020-07-29
Incremental blake2

Kachoc:
This pull request adds support for incremental hash for blake2.

Blake2 has the following differences with the other MD-based hashes:
	1. we need to pass how much data has been hashed up to now when hashing a block
	2. blake2 uses a different hash function for the very last block, which disallows eagerly calling update/update_multi when receiving data

For those reasons, we had to make the following changes to the specification functions and the code interfaces.

## Specification
In order to solve the first point, we added an `extra_state` to `words_state` which is now a pair. This `extra_state` is `uint64` for blake2s, `uint128` for blake2b and `unit` for the others, and is updated when calling `update`. `uint` was chosen instead of `nat` to avoid cumbersome additional refinements and preconditions on `update` to prove that it never overflows (though it should be possible to work with `nat`).

The second point was solved in the spec by modifying the computation of the last block for `hash_incremental` and streaming hash. The number of complete blocks to process and the size of the remaining data were initially computed as follows:
```
let num_blocks = total_length / block_length
let rem = total_length % block_length
```

We now do this (see [split_at_last_lazy_nb_rem](https://github.com/project-everest/hacl-star/blob/2a3b10ca35d0a9b34a5eccc5379d936360e0164d/lib/Lib.UpdateMulti.fst#L150) - all the spec split functions more or less directly refer to it in order to factorize the code and the proofs):

```
let num_blocks = total_length / block_length
let rem = total_length % block_length
let num_blocks, rem =
	if rem = 0 && num_blocks > 0
	then num_blocks - 1, block_length
	else num_blocks, rem
```

## Implementation
Because of the first point, we had to update the signatures of the hash implementations.

The [`Hacl.Hash.Definitions.update_st`](https://github.com/project-everest/hacl-star/blob/2a3b10ca35d0a9b34a5eccc5379d936360e0164d/code/hash/Hacl.Hash.Definitions.fst#L146), `Hacl.Hash.Definitions.update_multi_st` and `Hacl.Hash.Definitions.update_last_st` now take an `extra_state` parameter.

For the streaming functor, we have the following signature changes:
	* the [`update_multi`](https://github.com/project-everest/hacl-star/blob/2a3b10ca35d0a9b34a5eccc5379d936360e0164d/code/streaming/Hacl.Streaming.Interface.fsti#L346) and [`update_last`](https://github.com/project-everest/hacl-star/blob/2a3b10ca35d0a9b34a5eccc5379d936360e0164d/code/streaming/Hacl.Streaming.Interface.fsti#L365) functions of the `Hacl.Streaming.Interface.block` functor now take an additional `prevlen` parameter
	* the `totlen`parameter from [`update_last`](https://github.com/project-everest/hacl-star/blob/2a3b10ca35d0a9b34a5eccc5379d936360e0164d/code/streaming/Hacl.Streaming.Interface.fsti#L365) was replaced with `last_len`: **this might cause confusion in the client code** (see EverCrypt below)

Of course, those changes resulted in **similar updates in the EverCrypt hash function signatures** (see [`EverCrypt.Hash.update`](https://github.com/project-everest/hacl-star/blob/2a3b10ca35d0a9b34a5eccc5379d936360e0164d/providers/evercrypt/EverCrypt.Hash.fsti#L246), [`EverCrypt.Hash.update_multi`](https://github.com/project-everest/hacl-star/blob/2a3b10ca35d0a9b34a5eccc5379d936360e0164d/providers/evercrypt/EverCrypt.Hash.fsti#L268) and [`EverCrypt.Hash.update_last`](https://github.com/project-everest/hacl-star/blob/2a3b10ca35d0a9b34a5eccc5379d936360e0164d/providers/evercrypt/EverCrypt.Hash.fsti#L303)).

Besides, regarding the implementation details, the internal buffer is now lazily flushed (to take into account the point 2 above): if the buffer is full at some point, we wait for the next call to `update` with non empty data to flush it (this gives the invariant that as long as some data has been *seen*, the internal buffer is not empty, which means that the call to `update_last` will process data).

Finally, we updated the HMAC and HKDF implementations to add implementations using the different variants of blake2 (blake2s, blake2b, default and vectorized implementations).


Kachoc:
@R1kM the end of the work you started on hash incremental, finally ready for the PR. :) I would be happy if you could make a review.

Kachoc:
@protz for review

code/hash/Hacl.Hash.Definitions.fst
```diff
@@ -14,21 +15,67 @@ open FStar.Mul
...
+  [@inline_let] let a = get_alg i in
+  match a with
+  | MD5 | SHA1 | SHA2_224 | SHA2_256 | SHA2_384 | SHA2_512 -> state_word_length a
+  | Blake2S | Blake2B -> UInt32.v (4ul *. Blake2.row_len (to_blake_alg a) (get_spec i))
+
+inline_for_extraction
+type state (i:impl) =
+  b:B.buffer (impl_word i) { B.length b = impl_state_length i }
+
+(* TODO: make generic in the implementation *)
```
msprotz, line 56: is this todo obsolete?

code/hash/Hacl.Hash.Definitions.fst
```diff
@@ -49,45 +97,63 @@ let hash_len (a: hash_alg): n:size_t { v n = hash_length a } =
...
   (requires (fun h ->
     HS.is_stack_region (HS.get_tip h)))
-  (ensures (fun h0 s h1 ->
+  (ensures (fun h0 (s, v) h1 ->
     M.(modifies M.loc_none h0 h1) /\
     B.live h1 s /\
     B.frameOf s == HS.get_tip h0 /\
-    Seq.equal (B.as_seq h1 s) (Spec.Agile.Hash.init a) /\
-    LowStar.Monotonic.Buffer.alloc_post_mem_common s h0 h1 (Spec.Agile.Hash.init a)))
+    (as_seq h1 s, v) == Spec.Agile.Hash.init (get_alg i) /\
```
msprotz, line 132: this seems overly precise; where did you get the inspiration for this post-condition? the equality of domains should be given to you already by StackInline, same thing for the equality of the tips, and the `unused_in` should probably be replaced by `fresh_loc`

code/hkdf/Hacl.HKDF.fst
```diff
@@ -178,3 +182,25 @@ let expand_sha2_512: expand_st SHA2_512 =
...
+let extract_blake2s_32: extract_st Blake2S =
+  mk_extract Blake2S Hacl.HMAC.compute_blake2s_32
+
+let expand_blake2b_32: expand_st Blake2B =
+  mk_expand Blake2B Hacl.HMAC.compute_blake2b_32
+
+let extract_blake2b_32: extract_st Blake2B =
+  mk_extract Blake2B Hacl.HMAC.compute_blake2b_32
+
+val while_items: ni : nat -> ni_max: nat -> nw: nat{nw > 0} ->
```
msprotz, line 198: ?

code/streaming/Hacl.Streaming.Blake2.fst
```diff
@@ -0,0 +1,1077 @@
...
+open Hacl.Impl.Blake2.Generic
+module Impl = Hacl.Impl.Blake2.Generic
+module Incr = Spec.Hash.Incremental
+
+/// An instance of the stateful type class for blake2
+/// =========================================================
+
+#set-options "--max_fuel 0 --max_ifuel 0 --z3rlimit 200"
+
+/// Some lemmas I could not find in Math.Lemmas
```
msprotz, line 45: would be good to add to FStar.Math.Lemmas

Kachoc:
@msprotz I addressed the issues you pointed out in my last commits, put aside the arithmetic lemmas: I'm waiting for https://github.com/FStarLang/FStar/pull/2110 to be accepted before updating the files so that they reference the lemmas I added to FStar.Math.Lemmas

Kachoc:
You mean: replace `gupdate` by `update2` ? Indeed, some things are broken, I'm working on it...

msprotz:
> You mean: replace `gupdate` by `update2` ? Indeed, some things are broken, I'm working on it...

yes, `gupdate` looks like it's a ghost function

Kachoc:
@msprotz I introduced a versioning scheme, and when using attributes of the form `[@ deprecated "gupdate_multi" ]` there indeed are warning messages generated at compilation

Kachoc:
Yes, that's a good idea. We can discuss it on Slack

specs/Spec.Blake2.fst
```diff
@@ -528,7 +528,7 @@ let blake2s d kk k n = blake2 Blake2S d kk k n
 
 val blake2b:
     d:bytes
-  -> kk:size_nat{kk <= 64 /\ (if kk = 0 then length d < pow2 128 else length d + 128 < pow2 128)}
+  -> kk:size_nat{kk <= 64 /\ (if kk = 0 then length d < pow2 64 else length d + 64 < pow2 64)}
```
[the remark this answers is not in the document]
Kachoc: Thanks for pointing this out, it is actually a mistake: the spec should use the values given by the RFC, while the implementation limitations should be (are) written in the stateful functions signatures. I replaced 64 by 128.

specs/Spec.Hash.Definitions.fst
```diff
@@ -55,6 +77,8 @@ let max_input_length: hash_alg -> Tot pos = function
   | MD5 | SHA1
   | SHA2_224 | SHA2_256 -> pow2 61 - 1
   | SHA2_384 | SHA2_512 -> pow2 125 - 1
+  | Blake2S -> pow2 64 - 1
+  | Blake2B -> pow2 128 - 1
```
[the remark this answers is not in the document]
Kachoc: Doesn't conflict anymore now (see above)... Btw, I wondered why it didn't make the proofs fail, because there indeed is a conflict between `blake2b` and `max_length`. What happens is that in the spec proofs we always use the generic `blake2` which uses `max_limb` (which is correct), and in the implementation we either use buffers whose lengths are smaller than 2^32 (when we hash all the data at once), or we incrementally process data whose length is given by the RFC, but once again we use the generic blake2 spec...

specs/Spec.Hash.Definitions.fst
```diff
@@ -164,22 +261,23 @@ let size_hash = hash_length
 (** Padding *)
 
 (* Number of zeroes that should go into the padding *)
-let pad0_length (a:hash_alg) (len:nat): Tot (n:nat{(len + 1 + n + len_length a) % block_length a = 0}) =
+let pad0_length (a:hash_alg{not (is_blake a)}) (len:nat): Tot (n:nat{(len + 1 + n + len_length a) % block_length a = 0}) =
```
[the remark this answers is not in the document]
Kachoc: I replaced all those occurrences with calls to `is_md`

providers/evercrypt/EverCrypt.Hash.fsti
```diff
@@ -23,6 +23,13 @@ open Hacl.Hash.Definitions
 unfold
 let alg = hash_alg
 
+// TODO: we use this predicate to define functions which don't support blake, so
+// as to keep the old hash API. Remove once we don't need the old API anymore.
+let is_blake (a : alg) =
```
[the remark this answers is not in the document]
Kachoc: Indeed : I removed this redefinition

dist/c89-compatible/Hacl_HMAC.h
```diff
@@ -71,5 +72,37 @@ Hacl_HMAC_compute_sha2_512(
   uint32_t data_len
 );
 
+typedef struct K____uint32_t__uint64_t_s
```
msprotz, line 75: can you do a quick investigation and see why this appears here, and whether it's a kremlin bug, or that some other file needs it? (use `krml -d reachability`)

Makefile
```diff
@@ -769,6 +772,8 @@ dist/wasm/Makefile.basic: MERKLE_BUNDLE = -bundle 'MerkleTree,MerkleTree.*'
 dist/wasm/Makefile.basic: CTR_BUNDLE =
 dist/wasm/Makefile.basic: DEFAULT_FLAGS += -bundle 'EverCrypt,EverCrypt.*'
 
+dist/wasm/Makefile.basic: BLAKE2_BUNDLE += -bundle 'Hacl.Blake2b_256,Hacl.Blake2s_128'
```
msprotz, line 775: bad merge: I believe I had already added the right thing for you just above, so this line is superfluous

Makefile
```diff
@@ -907,6 +912,11 @@ dist/ccf/Makefile.basic: HACL_OLD_FILES =
 dist/ccf/Makefile.basic: POLY_BUNDLE =
 dist/ccf/Makefile.basic: P256_BUNDLE=-bundle Hacl.P256,Hacl.Impl.ECDSA.*,Hacl.Impl.SolinasReduction,Hacl.Impl.P256.*
 dist/ccf/Makefile.basic: HPKE_BUNDLE = -bundle 'Hacl.HPKE.*'
+dist/ccf/Makefile.basic: BLAKE2_BUNDLE=-bundle Hacl.Impl.Blake2.Constants \
+  -static-header Hacl.Impl.Blake2.Constants \
+  -bundle Hacl.HKDF.Blake2b_256,Hacl.HMAC.Blake2b_256,Hacl.Blake2b_256,Hacl.Hash.Blake2b_256 \
+  -bundle Hacl.HKDF.Blake2s_128,Hacl.HMAC.Blake2s_128,Hacl.Blake2s_128,Hacl.Hash.Blake2s_128 \
```
msprotz, line 918: this one needs the two `Streaming` flavors added

Makefile
```diff
@@ -928,7 +938,7 @@ dist/mozilla/Makefile.basic: NACLBOX_BUNDLE = -bundle Hacl.NaCl
 dist/mozilla/Makefile.basic: E_HASH_BUNDLE =
 dist/mozilla/Makefile.basic: MERKLE_BUNDLE = -bundle MerkleTree.*,MerkleTree
 dist/mozilla/Makefile.basic: CTR_BUNDLE =
-dist/mozilla/Makefile.basic: BLAKE2_BUNDLE = -bundle Hacl.Impl.Blake2.*,Hacl.Blake2b_256,Hacl.Blake2s_128,Hacl.Blake2b_32,Hacl.Blake2s_32
+dist/mozilla/Makefile.basic: BLAKE2_BUNDLE = -bundle Hacl.Impl.Blake2.*,Hacl.Blake2b_256,Hacl.Blake2s_128,Hacl.Blake2b_32,Hacl.Blake2s_32,Hacl.HMAC.Blake2b_256,Hacl.HMAC.Blake2s_128,Hacl.HKDF.Blake2b_256,Hacl.HKDF.Blake2s_128
```
msprotz, line 941: this one needs the two `Streaming` flavors added

dist/Makefile.tmpl
```diff
@@ -80,8 +80,8 @@ CFLAGS		+= -Wno-parentheses -Wno-deprecated-declarations -Wno-\#warnings -Wno-er
 CFLAGS_128 ?= -mavx
 CFLAGS_256 ?= -mavx -mavx2
 
-Hacl_Blake2s_128.o Hacl_Poly1305_128.o Hacl_Chacha20_Vec128.o Hacl_Chacha20Poly1305_128.o: CFLAGS += $(CFLAGS_128)
-Hacl_Blake2b_256.o Hacl_Poly1305_256.o Hacl_Chacha20_Vec256.o Hacl_Chacha20Poly1305_256.o: CFLAGS += $(CFLAGS_256)
+Hacl_Poly1305_128.o Hacl_Chacha20_Vec128.o Hacl_Chacha20Poly1305_128.o Hacl_Blake2s_128.o Hacl_HMAC_Blake2s_128.o Hacl_HKDF_Blake2s_128.o Hacl_Streaming_Blake2s_128: CFLAGS += $(CFLAGS_128)
```
msprotz, line 83: you need the `.o` for `Hacl_Streaming_Blake2s_128` (same below)

dist/c89-compatible/Hacl_HMAC.h
Kachoc: After investigation, this structure is used in `Hacl_HMAC_compute_blake2b_32` : see [here](https://github.com/project-everest/hacl-star/blob/904acaf6a11492513c26be0da2172eab780b5c2c/dist/gcc-compatible/Hacl_HMAC.c#L646) (and there is a similar problem for blake2s...)
It seems that the problem comes from [this line](https://github.com/project-everest/hacl-star/blob/904acaf6a11492513c26be0da2172eab780b5c2c/code/hmac/Hacl.HMAC.fst#L291), and where `alloca` is actually [Hacl.Hash.Core.Blake2.alloca_blake2b_32](https://github.com/project-everest/hacl-star/blob/904acaf6a11492513c26be0da2172eab780b5c2c/code/hash/Hacl.Hash.Core.Blake2.fst#L82)

PR #336, opened 2020-08-13
ocaml bindings: fix some misquoted path

tahina-pro:
The ocamlfind package destination directory processed by dune is passed as an argument to bash. On Windows (Cygwin), that path may be a Windows path with backslashes (\), thus we are caught into a quoting pitfall.
This pull request tries to solve this issue by having bash process the path instead of dune.
(A clean solution in the future would be to use the dune `copy` rule, but it currently does not support paths outside of the current building directory.)

### DIFF
--- a/bindings/ocaml/dune
+++ b/bindings/ocaml/dune
@@ -9,4 +9,4 @@
 
 (rule (targets config.h) (deps)
  (action
-  (copy %{lib:hacl-star-raw:config.h} .)))
+  (bash "cp %{lib:hacl-star-raw:config.h} .")))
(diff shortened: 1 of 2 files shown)


victor-dumitrescu:
I can't check whether this works on Windows or not, otherwise LGTM

tahina-pro:
It works on Windows, I tested it separately, see for instance https://github.com/project-everest/everest@5ac4b88ca53b3a2b9dffc29c7e1ca752da44c6f0 (that's specifically what used to fail, and the main reason why I opened this pull request.)

tahina-pro:
Thank you both for your reviews!

PR #338, opened 2020-08-17
Allow cross-compilation for ARM

franziskuskiefer:
This is a first stab at cross-compiling for iOS. It also fixes compilation on ARMv7.
Note that this only touches the files in `gcc-compatible` for now and should be moved up a directory before merging.


msprotz:
Hi Franziskus, thanks for taking a stab at this! Much appreciated!

Some questions about your pull request... trying to understand a bunch of stuff.

First, is there any way we can simplify the logic? Here's an attempt at this (diff from master):

```diff
diff --git a/dist/gcc-compatible/configure b/dist/gcc-compatible/configure
index 4de80e64d..7411a7032 100755
--- a/dist/gcc-compatible/configure
+++ b/dist/gcc-compatible/configure
@@ -91,11 +91,12 @@ detect_uint128 () {
 }
 
 detect_x64 () {
-  [[ $(uname -m) == "x86_64" ]]
+  [[ $build_arch == "x86_64" ]]
 }
 
 detect_arm () {
-  [[ $(uname -m | cut -c 1-3) == "arm" ]] || [[ $(uname -m | cut -c 1-7) == "aarch64" ]];
+  [[ $build_arch == "arm" ]] || \
+  [[ $build_arch == "aarch64" ]] ||
 }
 
 detect_arm_cc () {
@@ -152,11 +153,27 @@ EOF
 echo -n > Makefile.config
 echo -n > config.h
 
-if detect_arm; then
+# Parse command line arguments.
+all_args=("$@")
+while [ $# -gt 0 ]; do
+    case "$1" in
+        -arm) build_arch="arm" ;;
+        -aarch64) build_arch="aarch64" ;;
+    esac
+    shift
+done
+
+if [[ ! -z "$build_arch" ]]; then
+  echo "Doing cross build for $build_arch. Make sure that your toolchain is configured properly."
+else
+  build_arch=$(uname -m)
+fi
+
+if detect_arm; then
   echo "... detected ARM platform"
-  echo "... $(uname -m) does not support 256-bit arithmetic"
+  echo "... $build_arch does not support 256-bit arithmetic"
   echo "BLACKLIST += $(ls *CP256*.c *_256.c *_Vec256.c | xargs)" >> Makefile.config
-  echo "... $(uname -m) does not support legacy vale stubs"
+  echo "... $build_arch does not support legacy vale stubs"
   echo "BLACKLIST += evercrypt_vale_stubs.c" >> Makefile.config
   echo "CFLAGS += -DLib_IntVector_Intrinsics_vec256=\"void *\"" >> Makefile.config
   if detect_arm_cc; then
@@ -167,13 +184,14 @@ if detect_arm; then
     echo "$CC cannot compile 128-bit vector arithmetic, disabling"
     echo "BLACKLIST += $(ls *CP128*.c *_128.c *_Vec128.c | xargs)" >> Makefile.config
     echo "#define IS_ARM_7 1" >> config.h
+    echo "CFLAGS += -DLib_IntVector_Intrinsics_vec128=\"void *\"" >> Makefile.config
   fi
 fi
 
 if ! detect_x64; then
-  echo "$(uname -m) does not support x64 assembly, disabling Curve64"
+  echo "$build_arch does not support x64 assembly, disabling Curve64"
   echo "BLACKLIST += Hacl_Curve25519_64.c $(ls Hacl_HPKE_Curve64_*.c | xargs)" >> Makefile.config
-  echo "$(uname -m) does not support _addcarry_u64, using a C implementation"
+  echo "$build_arch does not support _addcarry_u64, using a C implementation"
   echo "#define BROKEN_INTRINSICS 1" >> config.h
   echo "#define IS_NOT_X64 1" >> config.h
 else
```

Second, I don't understand why you have to patch libintvector.h -- if you're on a cross-compilation toolchain, isn't it the case that the target system headers and the corresponding C cross-compiler should define proper compiler flags? For instance, I would expect that if you cross-compile to iOS, then neither __x86_64__ or _M_X64 will be defined, so I don't understand why a third define is needed to override those two. (I'm assuming you have exported a proper cross-compiling CC.)

Third, I don't understand the interaction between build_target and build_arch. If you're compiling, say, for iOS, shouldn't you already pass either `-arm` or `-aarch64`? Why is another flag needed?

Fourth, SDKROOT seems like something you would export from your own invocation. For instance, I assume you would have a little wrapper script:

```bash
SDKROOT=...
CC=$SDKROOT/.../arm-eabi-hf-whatever-cc
AR=$SDKROOT/...
./configure -aarch64
```

That way we don't need to upgrade our own configure script whenever Apple changes the location of their SDK.

There's probably a bunch of stuff I'm missing, since I'm not a cross-compilation expert, so please bear with me :). Once we're happy with the patch, we'll have to i) see whether this can be put under CI, ii) see how to document this and iii) see about patching dist/configure instead.

Thanks again!

Jonathan

franziskuskiefer:
Thanks for the feedback. I guess iOS is a somewhat weird target to configure. Let me try to answer your questions. This is actually the first time I did this for iOS so I might have also missed some things here.

I think first it's important to know that we of course use Apple `clang` here. And `clang` is a cross-compilatation tool in itself, i.e. no separate toolchain, like `CC=...`, is defined but the switch is a simple `-target ...` argument on the same `clang` (thinking about it, this might have caused the issues that required me to touch `libintvector` where I simply didn't set the target correctly).

To your first and third point:
ARM is not equal to iOS. Some reference for iOS and ARM compilation is [here](https://developer.apple.com/documentation/xcode/building_a_universal_macos_binary). With macOS moving to ARM as well there will be mostly ARM soon. But macOS is still pretty different from iOS (for now), like we can only have static libraries.
Therefore I think it would be good to know the target OS and architecture. For generic cross-compilation though your version should probably work. Maybe we can also do that in a first step.

To your second point:
I'm actually not entirely sure what mac defines here. But for some reason something like `x86_64` was defined. I have to look into why that is the case. So ideally this shouldn't be necessary.

To your fourth point:
I think we should define the `SDKROOT` here. There's no version in here (it symlinks to the actual version). So there shouldn't be issues on toolchain changes unless Apple changes paths in XCode. But then we probably have to touch something anyway.

> i) see whether this can be put under CI

yeah we should definitely put this under CI

> ii) see how to document this and

With the build setup and `configure` getting more complex we should add a proper documentation for building (with a `--help` and in a `README.md` or something). 

> iii) see about patching dist/configure instead

Right, this should eventually go into `dist/configure`.

Maybe its better to start with a generic `arm` and `aarch64` cross-compilation target. I just had iOS at hand to try. Let me try to make this a generic ARM patch.

franziskuskiefer:
Here another idea of how to do this. The Android ARM toolchains are pretty good so I decided to use them as a first testing target.

To test this you can use the NDK toolchains and run `./configure -target aarch64-linux-android` and `make libevercrypt.a` to produce a static library for aarch64 Android.

The changes in `Blake2` are necessary due to #339 and shouldn't end up in this PR.

dist/gcc-compatible/curve25519-inline.h
```diff
@@ -1,4 +1,5 @@
 #ifdef __GNUC__
+#if !(defined(__aarch64__) || defined(_M_ARM64) || defined(__arm__) || defined(_M_ARM))
```
msprotz, line 2: this is the only remark that I have: for the sake of being future-proof, it'd be good to make this:

```
#if defined(__x86_64__) || defined(_M_X64)
```

however, this file is auto-generated, so we can either:

- fix the generator in vale/code/crypto/ecc/curve25519/Inline25519.ml
- understand why this is being included in non-X64 builds... and how this previously used to work -- my theory: it's the include chain EverCrypt_Curve25519.h --> Hacl_Curve25519_64.h --> curve25519-inline.h that it causing trouble, and ideally, curve25519-inline.h would be included exclusively from Hacl_Curve25519_64.c, not Hacl_Curve25519_64.h

the latter is an easy fix, the former just requires a small option in kremlin (`-add-internal-include`) and a one-liner in the root Makefile

which solution would you prefer?

msprotz: also to be noted that I don't understand how this worked previously on our ARM travis CI

dist/gcc-compatible/Makefile.basic
```diff
@@ -14,7 +14,7 @@ endif
 CFLAGS 	+= -I. -I $(KREMLIN_HOME)/include -I $(KREMLIN_HOME)/kremlib/dist/minimal
 CFLAGS 	+= -Wall -Wextra -Werror -std=c11 -Wno-unused-variable \
   -Wno-unknown-warning-option -Wno-unused-but-set-variable \
-  -Wno-unused-parameter -Wno-infinite-recursion \
+  -Wno-unused-parameter -Wno-infinite-recursion -Wno-unused-function \
```
msprotz, line 17: why did you need this one, out of curiosity?

dist/gcc-compatible/configure
```diff
@@ -79,23 +88,23 @@ int main() {
   return 0;
 }
 EOF
-  $CC -I. -I../kremlin/include -I../kremlin/kremlib/dist/minimal -march=native -O3 $file -o $out && $out > $out.test
-  $CC -I. -I../kremlin/include -I../kremlin/kremlib/dist/minimal -DBROKEN_INTRINSICS -march=native -O3 $file -o $out && $out > $out.ref
+  $CC $CROSS_CFLAGS -I. -I../kremlin/include -I../kremlin/kremlib/dist/minimal -march=native -O3 $file -o $out && $out > $out.test
```
msprotz, line 91: does this -march=native override the CROSS_FLAGS? same below and in a bunch of other invocations

dist/gcc-compatible/Makefile.include
```diff
@@ -1,5 +1,5 @@
 USER_TARGET=libevercrypt.a
-USER_CFLAGS=-march=native -mtune=native -Wno-unused
+#USER_CFLAGS=-march=native -mtune=native -Wno-unused
```
msprotz, line 2: this is also an auto-generated file, so we'd have to remove these flags from the main Makefile (line 645), then tweak Makefile.tmpl to say something like:

```
# Neither the main Makefile or the configure script specified -march, default to native
ifeq ($(subst -march,,$(USER_CFLAGS)$(CFLAGS)),$(USER_CFLAGS)$(CFLAGS))
CFLAGS += -march=native -mtune=native
```
```

dist/gcc-compatible/configure
msprotz: I guess you have to rely on either $CROSS_CFLAGS to specify it and, otherwise, to be -march=native ... if this is the default behavior of gcc (i.e. ignore subsequent -march flags) then I guess we're good

dist/gcc-compatible/Makefile.basic
franziskuskiefer: `vale_impl_of_alg`, `poly1305_vale`, and `has_adx_bmi2` are `static` functions that aren't used but always get built.

franziskuskiefer:
closing in favor of #342

PR #340, opened 2020-08-19
make sure clang understands that r0-r3 are constants

franziskuskiefer:
Fixes #339

PR #341, opened 2020-08-20
fix neon clang (make sure clang understands that r0-r3 are constants)

### LINKED ISSUE #339: ARM builds with clang are broken
`clang` doesn't like the constants in `Hacl_Impl_Blake2_Constants_rTable_S` (`error: argument to '__builtin_neon_vsriq_n_v' must be a constant integer`). The integers here really are constants, but `clang` doesn't understand this. So instead of using `Hacl_Impl_Blake2_Constants_rTable_S` something like

```
#define Hacl_Impl_Blake2_Constants_rTable_S_0 16
#define Hacl_Impl_Blake2_Constants_rTable_S_1 12
#define Hacl_Impl_Blake2_Constants_rTable_S_2 8
#define Hacl_Impl_Blake2_Constants_rTable_S_3 7
```

could be used.

https://github.com/project-everest/hacl-star/blob/51054f0b49464d49654792dbdeb08460262bd4c5/dist/gcc-compatible/Hacl_Blake2s_128.c#L168

franziskuskiefer:
Fixes #339
Replacing #340


polubelova (approved):
seems fine to me. thanks!

PR #344, opened 2020-09-04
Add missing module alias

wintersteiger:
Not sure how this made it through the CI checks, my F* complains about the unknown module.

### DIFF
--- a/providers/evercrypt/fst/EverCrypt.Hash.fst
+++ b/providers/evercrypt/fst/EverCrypt.Hash.fst
@@ -8,6 +8,7 @@ module B = LowStar.Buffer
 module IB = LowStar.ImmutableBuffer
 module HS = FStar.HyperStack
 module ST = FStar.HyperStack.ST
+module M = FStar.Modifies
 
 module AC = EverCrypt.AutoConfig2
 module SC = EverCrypt.StaticConfig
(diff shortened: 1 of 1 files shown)


msprotz:
Should be a warning only, right? FStar.Modifies is outdated and is no longer used anymore so we shouldn't add a new edge in the dependency graph. I suspect the module alias is defined in the fsti but not in the fst (see EverCrypt.Hash.fsti:64) which is why you're getting a warning

The right fix is to remove module M = LowStar.Modifies in the fsti and replace every occurrence of `M.` with `B.` as `LowStar.Modifies` is now an alias for `LowStar.Buffer`

The second best fix is to define `M` in the fst to silence the warning, but then to at least define it in a consistent way with the fsti.

Thanks,

Jonathan

wintersteiger:
No, my build failed. I'll try replacing it with `B.`.

msprotz:
Did your build fail because of an unrelated error? You may have seen that in the build log but the latest verification run of today with the latest F* revision worked, so could it be a very recent breakage?


~ jonathan

wintersteiger:
I may have had an old F*, at some point earlier today I updated to the lastest master, but I can't remember whether that was before or after.

wintersteiger:
Yeah, that must have been it. When I tried it again just now it was only a warning. Fixed correctly now.

msprotz (approved):
Looks good to me, thanks for fixing that pesky warning! Please request review from anyone else (or ask for a review on slack) so that this can be merged.

Thanks,

Jonathan

polubelova (approved):

PR #347, opened 2020-09-07
Fix leaks in EverCrypt.Hash

Kachoc:
A small update to fix memory leaks in EverCrypt.Hash


msprotz (approved):
Thanks Son. Please request review from a second person so that you can merge!

R1kM (approved):

PR #352, opened 2020-09-12
Have Vale use 32-bit encoding for "xor64 r, r"

Chris-Hawblitzel:
(suggested by Uros Bizjak for WireGuard crypto to save some code bytes)


parno (approved):

msprotz (approved):
👍

Issue #353, opened 2020-09-14
at_exit hook registered potentially multiple times


raphael-proust:
In the current implementation, the `Gc.full_major` callback is set to execute `at_exit` in two distinct `init` functions (`AEAD.init` and `Hash.init`).

These functions can be called multiple times (a lot in the life of a long-running process). This can lead to a small memory-leak where the global `exit_function` callback grows slowly but surely, and a very lengthy process-termination that is spent scanning the whole heap repeatedly.

PR #354, opened 2020-09-14
Only register at-exit callback once

raphael-proust:
This aims to solve issue https://github.com/project-everest/hacl-star/issues/353

Note: this has not yet been tested.

### DIFF
--- a/bindings/ocaml/EverCrypt.ml
+++ b/bindings/ocaml/EverCrypt.ml
@@ -43,6 +43,8 @@ module Error = struct
     | n -> error n
 end
 
+let at_exit_full_major = lazy (at_exit Gc.full_major)
+
 module AEAD = struct
   open Error
   open Hacl_Spec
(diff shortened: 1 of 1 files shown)


msprotz:
Thanks for the bugfix Raphaël! Would you or @victor-dumitrescu have any idea how to test this?

victor-dumitrescu:
I've tested this using a simple loop that initialises an incremental hash and does a few updates. Indeed, the GC is invoked on exit the same number of times as the Hash was initialised, which I didn't expect when I first wrote this code. After applying this patch the GC is invoked only once. 

The difference in execution time is clearly noticeable. Initialising a SHA-256 hash 10.000 times, each time doing 10 updates, takes ~8s now and < 0.5s with this patch.

victor-dumitrescu (approved):

msprotz (approved):
Thanks. We still need CI to run and sadly it doesn't run for PRs from external forks. @victor-dumitrescu any chance you could push this to a private branch and just send a link here once you get a green, so that we can confirm this doesn't break anything? Thanks!

victor-dumitrescu:
CI is passing on branch [vdum_354](https://github.com/project-everest/hacl-star/tree/vdum_354).

msprotz:
Thanks!

For the record: I used my "administrator privileges" to merge this branch, because CI does not happen for PRs from external forks, and @victor-dumitrescu was able to confirm the change is green. This is to avoid having to open a separate PR and fragmenting the discussion.

PR #355, opened 2020-09-16
Changes to use of nested assert norms in support of upstream F* changes

nikswamy:
Thanks @polubelova.

This also works without the upcoming F* changes so it should be mergeable.
Also, as a general point of style, `norm_spec` instead of assert_norm for these usages should be more idiomatic.


polubelova (approved):
thanks!

msprotz (approved):
Thanks very much!!

PR #357, opened 2020-09-17
Add _GNU_SOURCE/_BSD_SOURCE explicit_bzero

wintersteiger:
CCF uses the MUSL library (as provided by Open Enclave), which defines `bzero` [here](https://github.com/openenclave/openenclave/blob/048a5c3df27220fbbba13bcfaba1ebf15a3d15cd/3rdparty/musl/musl/include/strings.h#L18) and `explicit_bzero` [there](https://github.com/openenclave/openenclave/blob/048a5c3df27220fbbba13bcfaba1ebf15a3d15cd/3rdparty/musl/musl/include/string.h#L81). The build warning message explicitly asks for a pull request, so here you go!

Full disclosure: I have no idea about the security implications of adding `defined(_GNU_SOURCE) || defined(_BSD_SOURCE)` here.

### DIFF
--- a/lib/c/Lib_Memzero0.c
+++ b/lib/c/Lib_Memzero0.c
@@ -36,7 +36,7 @@ void Lib_Memzero0_memzero(void *dst, uint64_t len) {
     SecureZeroMemory(dst, len);
   #elif defined(__APPLE__) && defined(__MACH__)
     memset_s(dst, len_, 0, len_);
-  #elif (defined(__linux__) && !defined(LINUX_NO_EXPLICIT_BZERO)) || defined(__FreeBSD__)
+  #elif (defined(__linux__) && !defined(LINUX_NO_EXPLICIT_BZERO)) || defined(__FreeBSD__) || defined(_GNU_SOURCE) || defined(_BSD_SOURCE)
     explicit_bzero(dst, len_);
   #elif defined(__NetBSD__)
     explicit_memset(dst, 0, len_);


msprotz:
Thanks for the PR!

According to https://man7.org/linux/man-pages/man7/feature_test_macros.7.html , the intention behind _GNU_SOURCE and _BSD_SOURCE is to allow the programmer to control what system headers define.

In other words, if you're dealing with musl, you may do:

```
#define _BSD_SOURCE
#include <string.h>
```

in order to make sure the definition for `explicit_bzero` is enabled.

In this patch, I'm not sure what effect this achieves: Kremlin's default Makefiles always enable _GNU_SOURCE for some endianness macros to be visible, so this test would always be true.

The right solution would be to feature-test in `configure` and then read the results in the chain of if-defs, or a dirtier solution might be for CCF to compile Lib_Memzero0.c with -D__linux__ -D_GNU_SOURCE to make sure the chain of if-defs takes the right branch.

wintersteiger:
Right, that won't work then. I'd like to avoid adding custom compile flags for each file, but I don't think there's a different macro in MUSL that we can make this conditional on. @achamayou what's your opinion? Maybe add `_GNU_SOURCE` to all of EverCrypt in the cmake scripts?

achamayou:
@wintersteiger I assume you mean our cmake scripts? That sounds good, I don't think there's a particular problem with that.

The only other thing I can think of is detect musl and a musl ver >= whenever this was added, but that's quite specific...

wintersteiger:
@achamayou yeah, I meant the CCF cmake scripts. I looked for a MUSL macro too, but they don't want any of that (e.g https://wiki.musl-libc.org/faq.html#Q:-Why-is-there-no-%3Ccode%3E__MUSL__%3C/code%3E-macro?).

achamayou:
@wintersteiger hah! that's not at all what I expected. Let's go with -D__linux__ in our cmake then I think.

PR #358, opened 2020-09-18
Enable -O2 for Merkle tree benchmarks in debug mode

wintersteiger:
The Merkle tree benchmarks ran out of stack space because tail-call optimization isn't performed in debug mode. This fix enables `-O2` on `MerkleTree.c` in debug mode. In release mode everything worked as expected. The main offending function was `LowStar.RVector.free_elems` in case someone wants to de-recursify it.


msprotz (approved):
Thanks!

polubelova (approved):

PR #360, opened 2020-09-29
test-bindings-ocaml: do not install hacl-star-raw if already there

tahina-pro:
So far one couldn't run `make test` twice in a row, because `make test` was running `make install-hacl-star-raw` which was installing the `hacl-star-raw` OCaml package, which ocamlfind refuses to do if it is already there. (Consequently, `./everest test` cannot be run twice in a row, either.)
This commit solves this issue by explicitly checking that `hacl-star-raw` is not already installed on the system.

I claim that this commit is required by the fact that a component of Project Everest will now require the `hacl-star` OCaml package to be installed. (namely: EverParse, cf. https://github.com/project-everest/everparse/commit/c21543a7feb8a569646b472d2030833888e44d00#r42820061)
Currently, `hacl-star-raw` is installed as part of `./everest test`, but I will soon need to install it along with `hacl-star` as part of `./everest make`, because EverParse will depend on it. See project-everest/everest@a48d201d0b56a43f146c0df1e571184af35446e3


msprotz:
Shouldn't it *remove* old installs rather? Because if there's a fix and I try to run the CI target on my machine, it'll just pick up the old version of hacl-star-raw

tahina-pro:
One could argue that `make test-bindings-ocaml` would be meant to only test the `bindings/ocaml` directory stricto sensu, independently of the `hacl-star-raw` package.

The problem is that, if `hacl-star-raw` and `hacl-star` are installed  after `make` (by the user or by `./everest make`), but if `hacl-star-raw` is rebuilt and reinstalled as part of `make test`, then there might be `*.cm*` inconsistencies between the leftover `hacl-star` and the newly reinstalled `hacl-star-raw`

msprotz:
Excellent point. Can we at least have a warning then, with `$(warn`, if there's already a hacl-star-raw installed?

tahina-pro:
Not sure I understand: by `$(warn`, do you mean the GNU make `$(warning ...)` function? If so, then this would be interpreted when reading the Makefile, not when executing the recipe, and so the code would look like:
```
hacl_star_raw_path := $(shell ocamlfind query hacl-star-raw)
hacl_star_raw_not_found := $(.SHELLSTATUS)
ifeq($(hacl_star_raw_not_found),0)
$(warning hacl-star-raw already installed)
test-bindings-ocaml:
else
test-bindings-ocaml: gcc-compatible
        +$(MAKE) -C dist/gcc-compatible install-hacl-star-raw
endif
        cd bindings/ocaml && $(LD_EXTRA) dune runtest
```
Is that what you are suggesting?

msprotz:
Ah I didn't remember that $(warning ...) was processed at recipe-parsing time rather than at recipe-execution time

In that case, I think just an `echo WARN: ...` would be enough. I just don't want anyone to waste time on this because they didn't realize the freshly compiled hacl-star-raw was not being installed.

msprotz (approved):
Looks good to me, thanks for taking my remarks into account. ~Please ask @victor-dumitrescu for review.~ let's wait for the second review then merge

victor-dumitrescu (approved):

PR #363, opened 2020-10-06
Remove kremlin submodule

### LINKED ISSUE #362: Broken submodule
https://github.com/project-everest/hacl-star/commit/05e0b0049aa4f6f2ca362ef44ff4548dbfda8f7b#diff-e111e895b719e3cbf8a5ac9cd36dba1b added a git submodule to `dist/test/c/kremlin`. But I don't see any `.gitmodule` file or other submodule definition. This breaks using `hacl-star` as a git submodule

```
git submodule update --init --recursive
fatal: No url found for submodule path 'hacl-star/dist/test/c/kremlin' in .gitmodules
```

franziskuskiefer:
fixes #362

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "dist/test/c/kremlin"]
-	path = dist/test/c/kremlin
-	url = https://github.com/FStarLang/kremlin.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "dist/test/c/kremlin"]
+	path = dist/test/c/kremlin
+	url = https://github.com/FStarLang/kremlin.git


franziskuskiefer:
Looks like the ci `script.sh` does this. Then this probably shouldn't get checked in in the first place?

msprotz:
This never should've been checked in -- I don't know why @Frosne had a submodule there, and why it was part of her PR. This must have been added after I reviewed the code, or maybe I missed it. Removing this it the right fix, which I understand is what your patch is proposing. I also see that the CI, which is designed to deal with these issues, has removed it: see https://github.com/project-everest/hacl-star/commit/d78d9c4d871a1a371b770fc053f32e17b0e2f6a6 and search for "submodule" (I couldn't figure out a way to send a direct link to the relevant part of the diff).

Can you confirm that everything is back to normal?

franziskuskiefer:
> Can you confirm that everything is back to normal?

My local copy is a little messed up now after toying with this. But the `git pull` on CI works again. Locally it's fine after manually patching up git too. So I think this should solve it.

msprotz:
Closing since the CI job removed that git submodule before I could merge this PR

PR #365, opened 2020-10-06
Finish cross-compilation work

msprotz:
This supersedes #342 and adds the following:
- we no longer hardcode any -mtune or -march in the Makefiles
- users are free to pass `CFLAGS="-march=native -mtune=native"` if they grew up using Gentoo
- otherwise, we don't specify anything and rely on the compiler's default value for -march and -mtune (most likely core2, it seems)
- allow disabling explicit_bzero when cross-compiling on Linux for an older toolchain

For safety:
- we still, on Azure CI only, compile dist/gcc-compatible with -march=native -mtune=native
- we still, on Azure CI only, compile dist/portable-gcc-compatible with -mtune=generic (the lowest supported value)

I'm opening up a fresh PR because I was working on a separate branch but most of the work here was done, of course, by the excellent @franziskuskiefer .


tahina-pro (approved):

.docker/build/build.sh
```diff
@@ -33,6 +33,13 @@ function vale_test() {
       make -j $threads vale.build -k
 }
 
+# This is just out of complete obsessiveness about CI: we check that the code
+# compiles on antedeluvian machines: mtune=generic is the lowest target GCC
```
franziskuskiefer, line 37: "antediluvian" ?

dist/Makefile.tmpl
```diff
@@ -23,18 +23,22 @@
...
 
+-include Makefile.config
+
 # 1. The usual pseudo auto-configuration
 
-UNAME		= $(shell uname)
-MARCH		= $(shell uname -m)
+# TODO: this should all move to the configure script
+# TODO: Makefile.config: configure; ./$<
+# TODO: include Makefile.config
```
franziskuskiefer, line 32: You can check out the [configure](https://github.com/project-everest/hacl-star/blob/franziskus/dist-build-update/dist/evercrypt/configure) and [Makefile](https://github.com/project-everest/hacl-star/blob/franziskus/dist-build-update/dist/evercrypt/Makefile) that do these things already. They are pretty aggressive in consolidating things. I'll leave it up to you if you want to incorporate these changes in this PR or if we do something like this in a follow up.

dist/configure
```diff
@@ -144,36 +150,140 @@ int main () {
...
+# This is being decomposed into the following information.
+# * target_arch: the target architecture, e.g. aarch64, arm, x86_64, i386
+# * target_sys: the target system, e.g. none, linux, win32, darwin
+# * target_abi: the target abi, e.g. eabi, gnu, android, macho
+
+if [ ! -z "$build_target" ]; then
+  echo "Doing cross build for $build_target. Make sure that your toolchain is configured properly."
+  if [[ "$build_target" == "aarch64-none-linux-android" ]]; then
+    # NOTE: We assume that the Android environment is set up correctly.
+    #       Make sure that `crtbegin_so.o` and `crtend_so.o` can be found.
```
franziskuskiefer, line 204: Depending on the build we'll need more here. But let's do that in a follow up when we actually have CI builds for this case as well.

dist/configure
```diff
@@ -144,36 +150,140 @@ int main () {
...
+
+    if [ -z $NDK ]; then
+      echo "Compiling for aarch64-none-linux-android requires an NDK toolchain."
+      echo "Then please set at least NDK appropriately. SYSROOT and NDK_TOOLCHAIN might be necessary as well."
+      echo "Get your toolchain here: https://developer.android.com/ndk/downloads/"
+      exit 1
+    fi
+    if [[ $SYSROOT == "" ]]; then
+      SYSROOT="$NDK/toolchains/llvm/prebuilt/darwin-x86_64/sysroot"
```
franziskuskiefer, line 213: These are MacOS specific right now. I'm fine with leaving this in here but we should do a follow up for Android builds soonish to clean this up.
Alternatively we could not try to set any `SYSROOT` or `NDK_TOOLCHAIN` here and require the caller to do it.

franziskuskiefer (approved):
Thanks for finishing this up!
I think there are a couple things that we should follow up on (see inline comments). But this is a good first step to get the basics in.

dist/Makefile.tmpl
msprotz: I'd like to do a follow-up if that's ok... since it's likely going to require a few tweaks.

I looked at your changes and they seem good. From a quick glance at your files, I only have two remarks:
- CFLAGS_{128,256} should be determine by the configure script, shouldn't they?
- I'd like to disable -march=native,-mtune=native by default -- this is causing quite a few problems on windows where people tend to share binaries, and if compiled on a host that has a processor that's too recent, it'll generate illegal instructions errors -- note that openssl doesn't enable -march=native,-mtune=native by default

Do you mind opening up a PR to follow up?

msprotz: (If I open the PR I can't review it)

franziskuskiefer: Yeah let's do this in a follow up. This was more meant as a note to follow up on 🙂 
I'd like to decide on the route first, then I'll update the branch and do a PR.

msprotz:
I'm merging this to get things moving... two issues remain:
- compilation on ARMv6 (Raspberry Pi) seems broken, @victor-dumitrescu to investigate
- more work with @franziskuskiefer on cleaning up dist/ and making things more palatable

PR #368, opened 2020-10-08
don't use libintvector.h on 32-bit arm

franziskuskiefer:
`libintvector.h` shouldn't be used on 32-bit ARM. I added `!defined(__ARM_32BIT_STATE)` for this.

### DIFF
--- a/dist/c89-compatible/INFO.txt
+++ b/dist/c89-compatible/INFO.txt
@@ -1,4 +1,4 @@
 This code was generated with the following toolchain.
-F* version: 0e284e249aefb378ac63d4f8e251372f636d965d
+F* version: 9ac1d8b6f4575008f380bbd3640d6b06c817db3e
 KreMLin version: f3656741e080d038f5f96a2864666e58ba34b831
 Vale version: 0.3.13
--- a/dist/c89-compatible/libintvector.h
+++ b/dist/c89-compatible/libintvector.h
@@ -419,7 +419,8 @@ typedef __m256i Lib_IntVector_Intrinsics_vec256;
   (_mm256_permute2x128_si256(x1, x2, 0x31))
 
 
-#elif defined(__aarch64__) || defined(_M_ARM64) || defined(__arm__) || defined(_M_ARM)
+#elif (defined(__aarch64__) || defined(_M_ARM64) || defined(__arm__) || defined(_M_ARM)) \
+      && !defined(__ARM_32BIT_STATE)
 #include <arm_neon.h>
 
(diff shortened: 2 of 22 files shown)


karthikbhargavan:
Can you say which parts of libintvector do not work on arm-32? Should be easy to fix.

franziskuskiefer:
There's at least https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vzip2q_u64 which is not supported on ARM 32.

franziskuskiefer:
@karthikbhargavan let me know how you want to fix this. But we should unbreak arm32 in some way.

prosecco:
Disabling is good for now. I need to find a way to test arm32On Oct 9, 2020 09:24, Franziskus Kiefer <notifications@github.com> wrote:
@karthikbhargavan let me know how you want to fix this. But we should unbreak arm32 in some way.

msprotz:
Thank Franziskus. Curious to hear if your change to `configure` fixes @victor-dumitrescu's issue? In any case, looks good to me, thanks for looking into this!!

msprotz (approved):
Looks good to me! 👍

franziskuskiefer:
Just noting another small change in here to make sure you've seen it. The `Lib_IntVector_Intrinsics_vec*` defines go into `config.h` now instead of the `Makefile.config`. This way they are later available as well (the Rust bindings need them again when compiling for example).

msprotz:
Yes I noticed and this seems quite cleaner than passing `-D` so I'm all in favor. Thanks for also cleaning up the ARM test and making sure the comment is up-to-date, much appreciated!

msprotz:
One thing that bothers me is that we don't have any CI scenario for this and we're bound to break 32-bit ARM again in the future (as @victor-dumitrescu has noticed). Maybe @victor-dumitrescu or you have ideas on how to test this?

msprotz:
Or maybe we could just cross-compile. Right now our only scenario is aarch64 cross-compilation.

franziskuskiefer:
Cross-compiling is certainly something we could add as a minimum target to make sure we don't break builds. 32-bit arm is pretty rare nowadays such that it'll be difficult to get a proper CI running that can actually test the build.

victor-dumitrescu:
> Thank Franziskus. Curious to hear if your change to `configure` fixes @victor-dumitrescu's issue? In any case, looks good to me, thanks for looking into this!!

I can confirm that this does indeed fix the build issue on my Raspberry Pi, thanks!

victor-dumitrescu (approved):

franziskuskiefer:
waiting for #372 before merging

PR #369, opened 2020-10-09
Do not return a promise when calling into the wasm code

raphael-proust:
This is a WIP of the general idea that I'm working towards. Ideally, instead of `api.js` giving me a fully realised module which contains functions that return promises (promises that are due to initialisation), I'd rather `api.js` give me a promise of something (module, object, whatever) which contains functions that return their result immediately.

### DIFF
--- a/bindings/js/api.js
+++ b/bindings/js/api.js
@@ -258,10 +258,6 @@ var HaclWasm = (function() {
     throw "Unimplemented !";
   };
 
-  var checkObj = {
-    checkIfInitialized: checkIfInitialized,
-  };
-
   var api_obj = {};
 
   // Creating object by mapping from api.json structure
@@ -271,14 +267,24 @@ var HaclWasm = (function() {
         api_obj[key_module] = {};
       }
       api_obj[key_module][key_func] = function(...args) {
-        return checkIfInitialized().then(function() {
-          return callWithProto(api_json[key_module][key_func], args);
-        });
+        if (isInitialized === false) {
+           throw 'Uninitialized';
+        };
+        return callWithProto(api_json[key_module][key_func], args);
       };
     });
   });
 
-  return Object.assign(checkObj, api_obj);
+  var getInitializedHaclModule = async function () {
+    await checkIfInitialized();
+    return api_obj;
+  };
+
+  var checkObj = {
+    getInitializedHaclModule: getInitializedHaclModule,
+  };
+
+  return checkObj;
 })();
 
 if (typeof module !== "undefined") {
--- a/bindings/js/api_test.js
+++ b/bindings/js/api_test.js
@@ -89,7 +89,7 @@ var passTest = async function(func_sig, func, msg, t) {
   console.log("Test #" + (t + 1) + " passed !");
 };
 
-async function checkTestVectors(func_sig, func, msg) {
+function checkTestVectors(func_sig, func, msg) {
   var number_of_tests = Infinity;
   func_sig.args.map(function(arg) {
     if (arg.tests !== undefined) {
@@ -104,18 +104,18 @@ async function checkTestVectors(func_sig, func, msg) {
     console.warn("No tests for " + msg + "!");
   }
   for (var t = 0; t < number_of_tests; t++) {
-    await passTest(func_sig, func, msg, t);
+    passTest(func_sig, func, msg, t);
   }
 }
 
-HaclWasm.checkIfInitialized().then(async function() {
+HaclWasm.getInitializedHaclModule().then(function(Hacl) {
   var tests = [];
   Promise.all(Object.keys(test_vectors).map(function(key_module) {
     Object.keys(test_vectors[key_module]).map(function(key_func) {
-      tests.push([test_vectors[key_module][key_func], HaclWasm[key_module][key_func], key_module + "." + key_func]);
+      tests.push([test_vectors[key_module][key_func], Hacl[key_module][key_func], key_module + "." + key_func]);
     });
   }));
   for (var i = 0; i < tests.length; i++) {
-    await checkTestVectors.apply(null, tests[i]);
+    checkTestVectors.apply(null, tests[i]);
   }
 });


denismerigoux:
Thanks @raphael-proust for this PR ! This is definitely a change we can incorporate here. Indeed, I was being overly defensive by checking the initialization for each HACL* API call. But you're right, this has the downside of transforming morally `sync` functions into `async`, which is not very practical (I know, I too have struggled a lot with `async` functions). So I approve this PR :)

denismerigoux:
Additional comments : 
* in the error message being thrown when the user forgot to initialize, change the basic one by somethign which provides help like `hacl.checkInitialized().then(() => ...)`
* Add a helper like `function getHaclModule() { return checkIfInitialized().then(() => hacl) ` that should be the goto promise to call to get the initialized module

denismerigoux (changes requested):
Additional comments : 
* in the error message being thrown when the user forgot to initialize, change the basic one by somethign which provides help like `hacl.checkInitialized().then(() => ...)`
* Add a helper like `function getHaclModule() { return checkIfInitialized().then(() => hacl) ` that should be the goto promise to call to get the initialized module

raphael-proust:
Good ideas! I'll make those changes soon and push them here.

Just posting something here: we basically use this wrapper when calling code from OCaml via `js_of_ocaml`:

```
let[@ocaml.warning "-21"] with_hacl f =
   let f = Js.wrap_callback f in
   Js.Unsafe.eval_string "global._HACL = require('./api.js');";
   let init_promise = Js.Unsafe.eval_string "_HACL.checkIfInitialized()" in
   Js.Unsafe.meth_call
     init_promise
     "then"
     Js.Unsafe.[| inject @@ wrap_callback f; inject @@ wrap_callback log |]
```

And our js_stubs declare a `_HACL` global.

raphael-proust:
One more thing: the init is very verbose, is it ok if in the PR I add a parameter to the initialisation function that's just a callback for logging? Is that something that'd be accepted? Otherwise the use within a CLI/`node` application is a bit hindered.

denismerigoux:
>  I add a parameter to the initialisation function that's just a callback for loggin

Currently, `checkIfInitialized()` returns a `Promise`, which you resolve by giving a callback using `.then(() => callback)`. I don't see why it is very verbose, maybe if you show me the changes you want to make I can give you a more precise opinion :)

raphael-proust:
I mean that the output of the program is very verbose. It basically display all the loading steps and results:

```
Custom imports: {}
Attempting to instantiate WasmSupport
Module WasmSupport successfully loaded
Adding exports into global import table:
  [
  'WasmSupport_align_64',
  'WasmSupport_check_buffer_size',
  'WasmSupport_betole32',
  'WasmSupport_betole64',
  'data_size'
]
This module has a data segment of size:  Global [WebAssembly.Global] {}
Next data segment will start at:  128

Attempting to instantiate FStar
Module FStar successfully loaded
Adding exports into global import table:
  [ 'FStar_UInt128_u32_64', 'FStar_UInt128_u32_32', 'data_size' ]
This module has a data segment of size:  Global [WebAssembly.Global] {}
Next data segment will start at:  128

Attempting to instantiate LowStar_Endianness
Module LowStar_Endianness successfully loaded
Adding exports into global import table:
  [ 'data_size' ]
This module has a data segment of size:  Global [WebAssembly.Global] {}
Next data segment will start at:  128

Attempting to instantiate Hacl_Hash
Module Hacl_Hash successfully loaded
Adding exports into global import table:
  [
  'Hacl_Hash_MD5_legacy_update_multi',
  'Hacl_Hash_MD5_legacy_update_last',
  'Hacl_Hash_MD5_legacy_hash',
  …
…
```

The idea would be to pass call `checkIfInitialized(logger)`. In practice, `logger` would be either not set (to not print anything) or it would be `((v) => { console.log(v); })` to print everything. But someone could have a smarter logger if they feel like it.

Alternatively, muting the output would also work.

denismerigoux:
Thanks for the clarification ! I see no problem in muting the console output; however most of it comes from the KreMLin-generated `loader.js` and not from `api.js`. But if you find a way to have these messages silenced no problems for me :)

raphael-proust:
Ok, I'll see if I find a way to make this work, thanks!

msprotz:
Yes you'll have to submit a PR on kremlin since these files get refreshed automatically as part of CI. They currently live here: https://github.com/FStarLang/kremlin/tree/master/kremlib/js

I would suggest the following workflow:
```bash
# modify kremlin's files
cd hacl-star
rm -rf dist/wasm/package.json
NODEPEND=1 make !$
# more commands to test the JS bindings
```
which at least will provide a faster loop for you

raphael-proust:
I've opened https://github.com/FStarLang/kremlin/pull/197 to allow setting a custom printer before loading.

I've modified this MR to do the following:

- forward the print-setting capacity of kremlin's loader from the aforementioned PR
- replace `checkIfInitialized` by `getInitializedHaclModule`

The new workflow, with `get…` works for my use-case (the `global` is very idiosyncratic though):

```
const HACL = require('./api.js');
HACL.setMyPrint((v) => { return; });
HACL.getInitializedHaclModule().then(function (HACL_initialized) {
    global._HACL = HACL_initialized;
    …
}
```

- Should `get…` be offered *alongside* `check…` or is the former sufficient for all use cases?
- I left an initialisation check for now, but it is actually useless with the new `get…` because it is impossible to call into an uninitialised function. (This is the primary reason to replace (rather than supplement) `check…` by `get…`.)
- Should the `get…` take an optional argument that's a printing function? The function `get…` would call the loader's `setMyPrint` before initialising.

denismerigoux:
To answer your questions: 

>     * Should `get…` be offered _alongside_ `check…` or is the former sufficient for all use cases?

We want to initialize the module once and get done with it so your `getInitializedHaclModule` is fine .

>     * I left an initialisation check for now, but it is actually useless with the new `get…` because it is impossible to call into an uninitialised function. (This is the primary reason to replace (rather than supplement) `check…` by `get…`.)

You can leave the check, it does not cost much and it will be useful for future hacking on this module because debugging where the unresolved promise comes from is not a fun game to play :)

>     * Should the `get…` take an optional argument that's a printing function? The function `get…` would call the loader's `setMyPrint` before initialising.

I don't think it's necessary, `getInitializedHaclModule` does not print anything (only the Kremlin-generated JS prints things).

denismerigoux (approved):
Thanks for the changes, I think it's good now, provided that the required PR have been merged in KreMLin (https://github.com/FStarLang/kremlin/pull/197)

raphael-proust:
Thanks for the answers. I've
- followed the suggestion to leave things as they are now,
- fixed indentation, and
- squashed the commits together.

I've also removed the printing part. This means that this PR doesn't depend on the kremlin one. This is because the no-promise feature is more important and not worth waiting on the other PR that is waiting on CLA.

msprotz:
@raphael-proust is there something preventing you from signing the CLA? let us know if we can help

@denismerigoux are there any tests to update now that the way to get access to the HACL object is different?

denismerigoux:
Can someone check that https://github.com/project-everest/hacl-star/blob/master/bindings/js/api_test.js is still working ? You just have to call it with `node api_test.js` and it should display a log of the tests and their results. I can't test it right now because I have to build HACL* again and I can't take an afternoon to have the build suceed (not until my paper deadlines are over).

beurdouche:
> Can someone check that https://github.com/project-everest/hacl-star/blob/master/bindings/js/api_test.js is still working ? You just have to call it with `node api_test.js` and it should display a log of the tests and their results. I can't test it right now because I have to build HACL* again and I can't take an afternoon to have the build suceed (not until my paper deadlines are over).

```
node:internal/modules/cjs/loader:903
  throw err;
  ^

Error: Cannot find module '/opt/hacl-star/bindings/js/loader.js'
Require stack:
- /opt/hacl-star/bindings/js/api.js
- /opt/hacl-star/bindings/js/api_test.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:900:15)
    at Function.Module._load (node:internal/modules/cjs/loader:745:27)
    at Module.require (node:internal/modules/cjs/loader:972:19)
    at require (node:internal/modules/cjs/helpers:88:18)
    at Object.<anonymous> (/opt/hacl-star/bindings/js/api.js:5:14)
    at Module._compile (node:internal/modules/cjs/loader:1083:30)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1112:10)
    at Module.load (node:internal/modules/cjs/loader:948:32)
    at Function.Module._load (node:internal/modules/cjs/loader:789:14)
    at Module.require (node:internal/modules/cjs/loader:972:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/opt/hacl-star/bindings/js/api.js',
    '/opt/hacl-star/bindings/js/api_test.js'
  ]
}
```

denismerigoux:
Thanks @beurdouche ! This is the error you get when you haven't called the right `make` target yet, I've forgotten to tell you. Actually I had autoumated the testing, you have to call 

    make test-wasm

From the root Makefile.

raphael-proust:
I fixed the tests. I had forgotten to check those.

As for the CLA, it's still waiting, both on my employer's input, but also I still have loads of questions about CLAs in general and I don't want to sign something before I actually understand it.

msprotz:
Pushed on a branch of the main repository because verification CI does not run for external forks. Branch rproust_no_promises

raphael-proust:
Has there been any progress into merging this?

I cannot see the details of the failing CI step because it requires logging into some service I don't have an account for.

PR #372, opened 2020-10-13
bump an rlimit

mtzguido:
We had a red build a few hours ago, but I couldn't reproduce it locally. Bumping this rlimit optimistically to see if that fixes it.


msprotz (approved):
thanks Guido let's hope this fixes it

mtzguido:
Managed to repro in the end... this is somewhat of a heisenbug. It's a definition in `Lib.Vec.Lemmas.fsti` that fails to typecheck when interleaved into `LIb.Vec.Lemmas.fst`. Both these files have a `#set-options` for using_facts_from, and they don't exactly agree. But that doesn't fix it... and in fact putting an empty `#set-options ""` in the fst just before the current one makes this work, so I suspect it's some tricky problem wrt interleaving and option setting, but I don't have a full hypothesis yet. Just removing the `set-options` from the fst also fixes it. Maybe we can do that to get back to green?

I can't explain why we got a green for this before, though. F* master at that point differs from the current master only slightly, and nothing in `ulib/` has changed.

msprotz:
Removing the set-options seems sensible. Also a good argument for using pairs of push/pop rather than set... unclear why this started failing on master. Maybe a hint replay was previously masking the snafu with the options setting?

mtzguido:
Just pushed a patch that seems to work locally, but I really have no clue what's going on. Push vs set doesn't seem to make a real difference (but I changed to push anyway), and I also remembered that pragmas on fstis are not interleaved into fsts. The empty set-options just before the push seems to fix this, no clue why. It's possible it was working before due to a hint hash replay, yes, but not sure how it could have been produced.

mtzguido:
It came back green. Maybe @polubelova is interested in the change. We should probably keep this commit hash marked, since this looks like an F* bug.

polubelova (approved):
thanks!

PR #378, opened 2020-10-23
Additions and enhancements to the JavaScript API

victor-dumitrescu:
This PR contains:

* Bindings for more existing wasm functions in api.json (NaCl, HMAC, keccak)
* Various bug fixes and enhancements to make these bindings work
* Bindings for P-256: relies on merging KreMLin support for extracting P-256 to wasm ([!198](https://github.com/FStarLang/kremlin/pull/198))
* P-256 point compression/decompression tests have also been added to the OCaml API, since they were missing


msprotz:
Thanks for the PR. Can you explain:
- the new size_offset parameter?
- why the ML tests for P256 have changed?

Thanks!

victor-dumitrescu:
The `size_offset` parameter is an integer that can be added to the `size` parameter to specify the size of a buffer. I needed this in order to specify the binding of the "easy" versions of Box functions, where the tag and the ciphertext are in one buffer, which has to be the size of the plaintext buffer + 16, for example here:
```json
        {
          "name": "c",
          "kind": "output",
          "type": "buffer",
          "size": "mlen",
          "size_offset": 16,
          "tests": [ ]
        },
 ```
I've also documented it [here](https://github.com/project-everest/hacl-star/blob/71fda46c5406a6b47f5adf6036b7a2999c405c1c/bindings/js/api.js#L133). Let me know if you think we should do this some other way.

The P-256 test is the same, I just changed some strings from decimal to hex, in order to use them in api.json. I had originally copied it from the C tests where it was written in a mixture of both and didn't bother changing it at the time.

**Comment by msprotz** [comment]
Ah it's just the name that confused me: I somehow thought that `offset` referred to an offset in memory and implied somehow that the size should be understood after pointer addition of 16 or something.

What do you think of doing something real quick, where size may be either `foo` (a parameter) or `foo+16` (parameter + extra constant size)? You could parse this in a really degenerate way (e.g. `if string.indexOf('+' >= 0) ...`) and it would enhance clarity of the API.

Thanks for the explanation regarding the tests, looks good!

**Comment by victor-dumitrescu** [comment]
I thought of doing that at first but decided against it because of the messier implementation, but maybe it's worth prioritising the clarity of the bindings.

**Comment by msprotz** [comment]
maybe something like this:
```
function parseLength(l) {
  var i = l.indexOf('+');
  if (i < 0)
    return [l,0];
  var base = l.substring(0, i).trim();
  var length = parseInt(l.substring(i+1));
  return [base,length];
}

var [ buf, extra ] = parseLength(...);
```

I don't think the parsing cost is too onerous, what do you think?

**Comment by victor-dumitrescu** [comment]
Just pushed an attempt at this. I actually also need (-) so it turned out a bit more convoluted. Let me know what you think.

File: bindings/js/api.js [hunk 1]
```diff
@@ -128,13 +130,19 @@ var HaclWasm = (function() {
         - 'name', the name of the argument which will be shown in the JS Doc
         - 'kind', either `input` or `output` of the function
         - 'type', either 'int' or 'boolean' or 'buffer'
-        - 'size', either an integer of a string which is the 'name' of another
+        - 'size', either an integer or a string which is the 'name' of another
           argument of type 'int'
+        - 'size_offset', an integer, which will be added to 'size' if the latter
```
msprotz, line 135: need to update here

bindings/js/api.js
```diff
@@ -161,6 +169,28 @@ var HaclWasm = (function() {
     return new Uint8Array(result);
   };
 
+  var parseSizeWithOp = function(arg, op, var_lengths) {
```
msprotz, line 172: rename to `evalSizeWithOp`

bindings/js/api.js
```diff
@@ -271,9 +306,10 @@ var HaclWasm = (function() {
         api_obj[key_module] = {};
       }
       api_obj[key_module][key_func] = function(...args) {
-        return checkIfInitialized().then(function() {
-          return callWithProto(api_json[key_module][key_func], args);
-        });
+        if (isInitialized === false) {
+           throw 'Nope';
```
msprotz, line 310: can we have a more helpful error message?

bindings/js/api_test.js
```diff
@@ -54,16 +54,9 @@ var passTest = async function(func_sig, func, msg, t) {
   });
   var result = await func.apply(null, args);
   if (func_sig.return.type !== "void") {
-    var result_val;
```
msprotz, line 57: @denismerigoux can you check this?

bindings/js/api.js
```diff
@@ -161,6 +169,28 @@ var HaclWasm = (function() {
...
 
+  var parseSizeWithOp = function(arg, op, var_lengths) {
+     if (arg.indexOf(op) >= 0) then
+       var terms = arg.split(op);
+       if (op === "+") {
+         return var_lengths[terms[0]] + parseInt(terms[1]);
+       } else if (op === "-") {
+         return var_lengths[terms[0]] - parseInt(terms[1]);
+       } else {
+         throw Error("Operator " + op + " not valid in `size` parameter, only at most one of {+,-} suppoted.")
```
msprotz, line 180: typo

msprotz (approved):
Looks good to me modulo:
- my small remarks
- Denis' approval for the changes to the JS API test harness

Let's wait for @denismerigoux to review this and then we'll be good to go.

Thanks @victor-dumitrescu for improving those bindings!

bindings/js/api_test.js
denismerigoux: So here @victor-dumitrescu you removed a small optimization I had made. Indeed, normally all API functions return an array containing the tuple of the results. But in the case where the tuple had one element, I was removing the array and directly returning the element. Do you remember why you changed this ?

denismerigoux (approved):
Modulo my small comment, I also approve the changes, thanks Victor!

bindings/js/api_test.js
victor-dumitrescu: There were a few things happening here. All functions in this category return a tuple of (result, output buffers) but we only care about the result at this stage, so it's fine to discard the rest. Here this was done if the function has any output buffers specified in api.json. However, even if that's not the case the function will return a tuple of (result, []) so I eliminated this check and I'm always taking the first part of the tuple as the result. Otherwise in this case result_val would be something like `[ true, [] ]`

Then, the result_val / expected_result check was broken, it should have been an inequality since if it's true we throw an error. The reason this wasn't being triggered is that this was comparing strings (as returned by `postprocessing`) with bool/ints (as returned by the function), which would always be false with === and so the test would pass. I added the posprocessing for the return value as well and fixed the check.

bindings/js/api.js
victor-dumitrescu: Sorry, I actually didn't mean to introduce this change, it was a test from another version of the file. Thanks for spotting it, I will reverse it.

bindings/js/api_test.js
denismerigoux: Oh thank you, I had not spotted this. Damn Javascript operators... I approve then :)

victor-dumitrescu:
Thanks for the review, I'll add the remaining tests and bindings and then merge.

msprotz:
Thanks Victor! Let's make sure we address Denis' remark here: https://github.com/project-everest/hacl-star/pull/378#discussion_r512521281 before merging. And then we're good.

victor-dumitrescu:
We already addressed it in the inline comment, I think there's no further change required

msprotz:
Ah thanks for some reason the github interface showed two different comment threads for the same issue. All good.

denismerigoux:
Correct, all good to go !

bindings/js/api.js
```diff
@@ -275,16 +276,13 @@ var HaclWasm = (function() {
       }
       return read_memory(pointer.value, size);
     });
-    if (return_buffers.length == 1) {
```
victor-dumitrescu, line 244: While adding tests for NaCl I had to make one last change to api.json on which I'd like to get @denismerigoux's approval before merging. Here we were flattening the return_buffers array only if it contained exactly one buffer, which is the case for almost all the functions in the API. A function would thus return
1. `[ ret ]` if it has no output buffers
2. `output_buf` if its return type is void and it has 1 output buffer
3. `[ ret, output_buf ]` if its return type is not void and it has 1 output buffer
4. `[ ret, [ output_buf1, output_buf2, ...]]` if its return type is not void and it has more than 1 output buffer

With these changes, case 1 and 3 remain unchanged and case 2 and 4 are now:
2. `[ output_buf ]` if its return type is void and it has 1 output buffer
4. `[ ret, output_buf1, output_buf2, ...]` if its return type is not void and it has more than 1 output buffer

So now, irrespective of return type and number of output values, all JS functions will return an array in which the first element is the return value (absent if return type if void) and the rest of the elements are output buffers.

victor-dumitrescu: I've made this change because I think it's a bit cleaner and it simplifies the logic required to consume these functions.

denismerigoux: I'm fine with that, it's indeed more regular

PR #380, opened 2020-11-03
support basic 32-bit cross compilation

franziskuskiefer:
This allows to compile a very basic 32-bit version of the gcc-compatible C code. We might be able to enable more, but this is the minimal thing we could do.


dist/gcc-compatible/configure
```diff
@@ -245,12 +250,21 @@ if [ ! -z "$build_target" ]; then
     CC="$TOOLCHAIN/bin/aarch64-none-linux-gnu-gcc"
     echo "CC=$CC" >> Makefile.config
     # Using GCC here doesn't require setting anything else.
+  elif [[ "$build_target" == "ia32" ]]; then
```
msprotz, line 253: the `ia32` name seems a little unfamiliar; I remember `i686` and `i386` being more common in, say, configure triples (e.g. i686-linux-gnu) or in `uname -m` return values (see https://stackoverflow.com/questions/45125516/possible-values-for-uname-m) for some information -- should we settle on i686 as the one possible value to indicate 32-bit x86?

dist/gcc-compatible/configure
```diff
@@ -100,6 +100,11 @@ detect_x64 () {
   [[ $target_arch == "x86_64" ]]
 }
 
+detect_x86 () {
+  [[ $target_arch == "i386" ]] || [[ $target_arch == "i686" ]] || \
```
msprotz, line 104: here too it feels like we're trying to match a lot of stuff; do you know if there are systems that may return `i86pc` or `ia32`? (just curious)

dist/gcc-compatible/configure
```diff
@@ -282,6 +296,20 @@ if detect_arm; then
...
+  echo "... $build_target does not support 256-bit arithmetic"
+  echo "BLACKLIST += $(ls *CP256*.c *_256.c *_Vec256.c | xargs)" >> Makefile.config
+  echo "... $build_target does not support legacy vale stubs"
+  echo "BLACKLIST += evercrypt_vale_stubs.c" >> Makefile.config
+  echo "#define Lib_IntVector_Intrinsics_vec256 void *" >> config.h
+  echo "$CC $CROSS_CFLAGS cannot compile 128-bit vector arithmetic, disabling"
+  echo "BLACKLIST += $(ls *CP128*.c *_128.c *_Vec128.c | xargs)" >> Makefile.config
+  echo "#define Lib_IntVector_Intrinsics_vec128 void *" >> config.h
+  echo "disabling the use of explicit_bzero"
+  echo "#define LINUX_NO_EXPLICIT_BZERO" >> config.h
```
msprotz, line 310: I would suspect that, say, a recent 32-bit linux distribution still supports bzero by virtue of having a recent glibc. Can we remove this bit and let the test at line 346 kick in?

msprotz (approved):
This looks good and I think it should unblock us for now. My remarks are mostly cosmetic and I don't have strong opinions about i686 vs ia32, I'm mostly curious.

Two things need to be fixed:
- let bzero detection kick in automatically
- copy those changes to `dist/configure` as well

In the long run, I'd like to understand whether anyone still cares about i686 -- I assume not. If people do, I'd love to understand why, and understand whether avx and avx2 are available in 32-bit mode or not, and whether we should make efforts to support these (weird) configurations.

Thanks very much @franziskuskiefer for submitting this -- this is great.

dist/gcc-compatible/configure
franziskuskiefer: `ia32` is the generic name for Intel 32-bit (https://en.wikipedia.org/wiki/IA-32). The issue is that there are so many different architectures that are 32-bit and they all have another architecture name (like the ones I match `detect_x86` and more).
I don't have strong opinions on the names we use here. The most generic way would be to not actually specify the target here but use a separate argument like `-m32` to indicate a 32-bit build. This would mimic the compiler flag we set anyway.

dist/gcc-compatible/configure
franziskuskiefer: Yeah I can do that.

dist/gcc-compatible/configure
msprotz: ok no need to change then -- thanks for the background, `ia32` is good!

msprotz: agreed that there is confusion between i386/i686 which are revisions of the architecture

.ci/script.sh
```diff
@@ -47,6 +47,12 @@ else
   make -C kremlin/kremlib/dist/generic -f Makefile.basic -j
   make -j -k
   popd
+
+  # Test 32-bit build; Tests don't work here yet.
```
msprotz, line 51: I'm under the impression this will install 32-bit built ocaml bindings for all x64 architectures, possibly leading to less test coverage... can we make this a separate line in travis?

franziskuskiefer: Sure, will do

franziskuskiefer:
@msprotz who else could r+ to get this landed?

beurdouche (approved):

beurdouche:
@protz Can you merge it, if you are still happy with this? Thanks !

msprotz:
Thanks @beurdouche !

PR #381, opened 2020-11-03
Remove noextract in specs and libs

nikswamy:
This PR removes all occurrences of noextract in specs and lib.

An upcoming change in F* will cause defintions marked `noextract` to be dropped even when extracting to OCaml (previously, noextract only impacted kremlin extraction). With this change, several definitions which are used in specs and needed in OCaml for spec testing were no longer available. Removing the qualifier fixes the problem.


beurdouche (approved):

msprotz (approved):
This looks good to me, feel free to merge once you get a green (and are up-to-date w.r.t. master).

Heads-up to @karthikbhargavan @polubelova @Frosne: now if a local Makefile doesn't do bundling properly, i.e. fails to use `BASE_FLAGS`, this will cause kremlin errors that might not have existed before. Fortunately, based on a quick grep, all local Makefiles currently use `BASE_FLAGS` so we should be good.

PR #382, opened 2020-11-05
EverCrypt.Hash.Incremental: add a --retry on flaky proof and update hints

mtzguido:
Got a failure here when trying out an everest build for FStarLang/FStar#2090.

Is the `--retry` acceptable? I ran this with rlimit=400 (so 2x) and it still failed, but it succeeds on the second try even with rlimit=100 (0.5x).


beurdouche (approved):

nikswamy:
ping @msprotz @karthikbhargavan 

Would be great to merge this PR or revise this particular proof. There are a couple of PR's queued up on F* that depend on this.

msprotz (approved):
This looks good to me. As a side note, officially requesting github reviews helps make sure this is ready for consideration.

nikswamy:
thanks!

PR #386, opened 2020-12-03
Deprecate two phase tc option

aseemr:
This option is true by default, and I think it is safe to say that unsetting it is not supported/maintained.


msprotz (approved):

beurdouche (approved):

beurdouche:
Thanks @aseemr ! :)

PR #389, opened 2020-12-28
New CI for dist

franziskuskiefer:
This is only part of the travis CI but should get us started with GH actions (we can add the other configs later).

`.github/workflows/dist.yml` implements the workflow for Windows, Linux, MacOS (all x64)
`.drone.yml` would add ARM64 and ARM32 CI. I don't have permission to activate the repo on drone.io though. So I don't know if it actually works. If we want to go for drone.io (for ARM), someone would have to enable it.


beurdouche (approved):

msprotz:
Neat!! Thanks so much Franziskus.

This looks good to me for the time being.
- What is drone.io and who needs to enable support for it? @beurdouche do you know how to enable that? (I don't)
- Do we have any plans for testing slightly older linux distros? What concerns me is making sure the GCC bugs don't reappear 

Thanks again!

msprotz (approved):
Approving but let's try to enable the ARM CI before we merge?

franziskuskiefer:
> Approving but let's try to enable the ARM CI before we merge?

Yeah I agree. But I don't have permissions to do it. The repo needs to be activated. https://cloud.drone.io/project-everest/hacl-star

franziskuskiefer:
> Neat!! Thanks so much Franziskus.
> 
> This looks good to me for the time being.
> - What is drone.io and who needs to enable support for it? @beurdouche do you know how to enable that? (I don't)
> - Do we have any plans for testing slightly older linux distros? What concerns me is making sure the GCC bugs don't reappear 
> 
> Thanks again!

We can either create docker images for older Linux/GCC and use them, or install older GCC on the default runner. Docker images is probably the easiest and makes it independent of the actual CI running it.

msprotz:
This should be enabled now. For reference, I've granted drone.io permission to the hacl-star repository

franziskuskiefer:
Looks like some ocaml dependency is making problems (https://cloud.drone.io/project-everest/hacl-star/8/1/2). I removed the ppa because it didn't work. 

> E: The repository 'http://ppa.launchpad.net/avsm/ppa/ubuntu hirsute Release' does not have a Release file.

We can either try to get this to work or not do the ocaml bindings on arm.
What do you think @msprotz ?

msprotz:
thanks Franziskus for looking into this!

@victor-dumitrescu you're building OCaml bindings on ARM, right? what setup do you use?

@franziskuskiefer is it the case that things don't work without the PPA? perhaps these packages have been merged into the main repository now which is why the PPA doesn't provide them anymore for "hirsute"?

franziskuskiefer:
> @franziskuskiefer is it the case that things don't work without the PPA? perhaps these packages have been merged into the main repository now which is why the PPA doesn't provide them anymore for "hirsute"?

Right, without the PPA the installs all appear to work but then there's an error switching opam. So I think the PPA shouldn't be needed anymore and all packages are in the main repositories. I think the following is the first error after calling `opam switch create 4.09.1 --yes`.

> [ERROR] The compilation of ocaml-base-compiler failed at "/root/.opam/opam-init/hooks/sandbox.sh build ./configure --prefix=/root/.opam/4.09.1 -C".

msprotz:
strange! I seem to vaguely remember that sandbox has to be disabled for opam to compile on arm -- this might be the magic flag that's required? https://www.reddit.com/r/ocaml/comments/fdg7a9/trouble_installing_utop_for_ocaml_on_ubuntu_1804/

hopefully Victor has some more info

victor-dumitrescu:
Hmm this issue doesn't ring any bell. I was building the bindings on a 32-bit Raspberry Pi (running Raspberry Pi OS) and I don't remember running into anything like this.

victor-dumitrescu:
(I stopped because Tezos no longer aims to support 32-bit ARM, but we'd still need the bindings to work on arm64)

msprotz:
strange -- I don't understand what the difference might be between the old travis environment and this one, but since we used to use bionic + Anil's PPA and now we're on hirsute + built-in OPAM, I guess there were bound to be changes

in any case, I did a little more research based on the error over `bwrap` in the build log and apparently sandboxing is indeed the issue, see https://github.com/ocaml/opam-repository/issues/12050#issuecomment-393478072 for a workaround (`--disable-sandboxing`)

so, looks like we should be able to re-enable aarch64 CI soon...!

franziskuskiefer:
Looks like this fixes the build! 🎉 
The Arm32 build doesn't go through because we'd probably have to set `--disable-ocaml` on it. But it looks like no one really needs Arm32 and we could just go with Arm64 only? @msprotz

msprotz:
thanks!! awesome

I don't have a strong opinion one way or another -- I did make the configure thing work on arm32 on my local raspberry pi, so if you change line 46 of .ci/script.sh to also catch arm32, we might get arm32 to work without any further changes

whether that's worth putting under CI and keeping track of is another story -- if you think that's not worth the effort, totally fine by me :)

franziskuskiefer:
Only changing the CI script doesn't appear to work for the arm32 tests. I removed the arm32 builds for now. If it's green, I'll merge. Then we have a baseline again and decide what else we want to add.

> chacha20-arm-test.c:(.text+0x2a4): undefined reference to `Hacl_Chacha20_Vec128_chacha20_encrypt_128'

msprotz:
Yeah looks like the target arm-test in tests/ assumes 128-bit implementations unconditionally. Probably something to fix in the future but all fine by me. Thanks for restoring some modicum of CI!

PR #390, opened 2021-01-06
update FrodoKEM to round 3 version

polubelova:


code/frodo/Hacl.Frodo.Random.fst
```diff
@@ -14,11 +14,14 @@ module S = Spec.Frodo.Random
...
 friend Lib.IntTypes
 
+#set-options "--z3rlimit 50 --fuel 0 --ifuel 0"
+
 let state = gcmalloc HyperStack.root 0uy 48ul
 
 let randombytes_init_ entropy_input =
   let h0 = HyperStack.ST.get () in
-  assume (as_seq h0 state == S.randombytes_init_ (as_seq h0 entropy_input))  
+  assume (as_seq h0 state == S.randombytes_init_ (as_seq h0 entropy_input))
```
msprotz, line 23: would be nice to add a comment here in case someone reviews this code to explain the assumption (i.e. that we can't "realize" randomness and thus have to assume it)

msprotz (approved):
This generally looks good to me but it'd be nice to describe what this PR does in a high-level way, if only for future reference. For instance:
- introduce a layer of agility for different variants of frodo parameters in spec
- make the implementation agile, too
- rely on partial evaluation to generate specialized variants of Frodo

Thanks,

Jonathan

karthikbhargavan (approved):

code/frodo/Hacl.Frodo.Random.fst
msprotz: note: with @Kachoc's most recent PR, you *may* be able to do something about that (but I'm not sure)

polubelova: I'm not sure either, but I know that for testing we need to drop the implementation of this function in favor of https://github.com/microsoft/PQCrypto-LWEKE/blob/669522db63850fa64d1a24a47e138e80a59349db/tests/rng.h#L70. It seems fine to me to leave it as is.

PR #391, opened 2021-01-22
bignum library

polubelova:
This PR introduces a verified bignum library that

- supports rsa-pss and ffdhe (2048-16384 bits)
- supports 32-bit and 64-bit machines
- provides the following operations
	* addition, subtraction
	* multiplication, squaring (textbook and karatsuba)
	* Montgomery modular reduction
	* modular exponentiation (binary and fixed-window methods; constant-time and variable-time impls)
	* load and store
- replaces the bignum code from curve25519_64_hacl (we keep this variant only for demonstration of how proofs work)


code/bignum/AUTHORS.md
```diff
@@ -0,0 +1 @@
+This code was primarily written by Marina Polubelova (INRIA).
```
msprotz, line 1: you may want to add something along the lines of "Jonathan Protzenko contributed the design of the type classes hierarchy" so that if there's questions about that you can deflect and not be bothered

code/bignum/Hacl.Bignum256.fst
```diff
@@ -0,0 +1,137 @@
...
+module BM = Hacl.Bignum.Montgomery
+module BE = Hacl.Bignum.Exponentiation
+module BR = Hacl.Bignum.ModReduction
+module BI = Hacl.Bignum.ModInv
+
+#set-options "--z3rlimit 50 --fuel 0 --ifuel 0"
+
+let _ = assert_norm (256ul = 64ul `FStar.UInt32.mul` 4ul)
+
+/// A note about the style of normalization used in this file. Normally,
```
msprotz, line 15: this comment is outdated

code/bignum/Hacl.Bignum256.fst
```diff
@@ -0,0 +1,137 @@
...
+
+module BN = Hacl.Bignum
+module BM = Hacl.Bignum.Montgomery
+module BE = Hacl.Bignum.Exponentiation
+module BR = Hacl.Bignum.ModReduction
+module BI = Hacl.Bignum.ModInv
+
+#set-options "--z3rlimit 50 --fuel 0 --ifuel 0"
+
+let _ = assert_norm (256ul = 64ul `FStar.UInt32.mul` 4ul)
```
msprotz, line 13: this assertion should probably mention n_limbs

code/bignum/Hacl.Bignum4096.fst
```diff
@@ -0,0 +1,137 @@
...
+module BM = Hacl.Bignum.Montgomery
+module BE = Hacl.Bignum.Exponentiation
+module BR = Hacl.Bignum.ModReduction
+module BI = Hacl.Bignum.ModInv
+
+#set-options "--z3rlimit 50 --fuel 0 --ifuel 0"
+
+let _ = assert_norm (4096ul = 64ul `FStar.UInt32.mul` 64ul)
+
+/// A note about the style of normalization used in this file. Normally,
```
msprotz, line 15: also outdated, and it looks like the assert above is perhaps already in the fsti?

msprotz (changes requested):
Thanks for a PR long in the making!

The code looks quite exemplary and clean. I had previous familiarity with the code from working together on the type classes so it looks mostly like a cleaned-up version of what we had before, with of course a ton of new exciting features like Karatsuba.

My main concern is that with an exemplary, high-quality development like this, there ought to be a README.md that gives readers a "tour" of the code. For instance, you could write a little bit about this code:
- there's a very strict design choice, which is that your Low* functions only conform to spec-level functions, and nothing more; this is evidence by the post-conditions, of the form `out == Spec.mult a b`, rather than of the form `interp out == interp a * interp b`
  - you could talk about why requiring the user to call a spec-level lemma is a good design choice, even though it may hide some preconditions and make the functions more dangerous to use --  and then say that the "safe API" you and I designed in the specialized versions is actually ok for unverified clients
  - you could also explain that that all of the reasoning is done on spec-level functions, and that, as a Low*/F* expert, you think this is the most viable style going forward
- there's a big space of choices for users: run-time vs. specialized, 32-bit vs. 64-bit, Karatsuba vs. regular multiplication... can you provide a guide, or at least some insights, for F* clients and C clients?

Note that I'm only suggesting topics that you're going to end up writing about in your PhD dissertation (presumably) so this shouldn't be any extra work for you. But mostly, I want to make sure that there's a little guide for this great piece of code, to encourage people to use and maintain in the future.

Two other points.
- Congrats on reusing this for the Curve code. It's really nice to show it's usable.
- I will need to check whether the C code that's in your PR matches the one that was given previously to Election Guard. I'll get back to you on that.

Congrats again on a great piece of work!

code/bignum/AUTHORS.md
polubelova: Sure! Thanks once again for your help! :)

msprotz (approved):
Thanks for the fast turnaround time! I'll let you hit the "merge" button. Congratulations again!

karthikbhargavan (approved):
Lots of great work here.

PR #397, opened 2021-02-05
Update the hash streaming functor for vectorized implementations

Kachoc:
This PR updates the hash streaming functor to make its internal buffer a multiple of blocks, and not exactly a block. This way, the functor fully exploit the performance provided by the vectorized implementations.


Kachoc:
I take the liberty of bringing the reviewers' attention to those lengths, which I'm not completely sure of:
https://github.com/project-everest/hacl-star/blob/dd3d72db24104695f13c904ef6142d715929cd22/code/streaming/Hacl.Streaming.Blake2.fst#L719
https://github.com/project-everest/hacl-star/blob/dd3d72db24104695f13c904ef6142d715929cd22/code/streaming/Hacl.Streaming.Poly1305.fst#L414

Kachoc:
@protz for your information

code/streaming/Hacl.Streaming.Functor.fst
```diff
@@ -682,8 +708,10 @@ let split_at_last_rest_eq (#index : Type0) (c: block index)
   let sz = rest c i total_len in
   ()
 
+// This rlimit is a bit crazy, but this proof is not stable. It usually
+// goes through fast, but a high rlimit is a security against regression failures.
```
msprotz, line 712: famous last words ;-)

msprotz: although the fact that non-linear arithmetic is disabled is, indeed, a good guarantee that it'll be robust

providers/evercrypt/fst/EverCrypt.Hash.Incremental.fst
```diff
@@ -75,6 +75,7 @@ let evercrypt_hash : block hash_alg =
     Hacl.Streaming.MD.max_input_length64
     hash_len
     block_len
+    block_len // No vectorization
```
msprotz, line 78: it would be interesting to have a larger buffer here and allow vectorization on the underlying blake2 implementations but that's another story

msprotz (approved):
Ok this all looks excellent to me. Thanks!

polubelova:
I agree that M32 could lead to incorrect computations of the block lengths. 

the correct formula for poly1305 should be the following:

```
match fs with
| M32 -> block_length (=16)
| M128 -> 2 * block_length (=32)
| M256 -> 4 * block_length (=64)
```

@karthikbhargavan  should know about the blake2 case.

code/streaming/Hacl.Streaming.Functor.fst
Kachoc: Yes, this doesn't protect against regression failures indeed... I mostly meant CI failures, for this exact PR ^^'
And I was quite surprised by the proof time variance, though I deactivated non-linear arithmetic - it would be worth investigating.

polubelova:
@Kachoc  can you please also add some C tests for blake2-streaming (all variants) and poly1305-streaming (only for M128 and M256) similar to this one: https://github.com/project-everest/hacl-star/blob/master/tests/poly1305-test-streaming.c?
Thanks!

Kachoc:
@polubelova I updated the blocks lengths for Poly1305 and added some C tests for Blake2 and Poly

polubelova (approved):
Thanks @Kachoc!

msprotz:
Side remark: `Hacl_Streaming_Functor_state_s___Lib_IntVector_Intrinsics_vec128___uint8_t_` is kind of a mouthful. Now that I've fixed the kremlin bug, maybe time to add a palatable abbreviation?

Kachoc:
I updated the blake2 internal buffers lengths and added abbreviations for the functor states @protz

PR #400, opened 2021-02-10
crypto_random implementation

Kachoc:
This PR introduces a blocking implementation for `Lib.RandomSequence.crypto_random`, which tries to be as close as possible to the spec, and comments `Lib.RandomBuffer.System.randombytes` as deprecated.


Kachoc:
tagging @protz  for information

lib/Lib.RandomBuffer.System.fsti
```diff
@@ -2,13 +2,40 @@ module Lib.RandomBuffer.System
...
 open Lib.IntTypes
 open Lib.Buffer
+module B = LowStar.Buffer
 
 open FStar.HyperStack.All
 
+open Lib.RandomSequence
+module R = Lib.RandomSequence
 
+/// WARNING: This function is DEPRECATED. Use ``random_crypto`` instead.
```
beurdouche, line 12: I think there is a kremlin macro to deprecate functions, isn't it @protz ?

beurdouche (approved):

lib/Lib.RandomBuffer.System.fsti
msprotz: yes! `[@@ deprecated "random_crypto" ]` which will give a deprecation warning in F* and in C as well via compiler-specific pragmas

lib/Lib.RandomBuffer.System.fsti
```diff
@@ -2,13 +2,40 @@ module Lib.RandomBuffer.System
...
+/// This function waits until it has received enough entropy before
+/// generating random bytes. Note that it may wait for some time.
+val crypto_random:
+     buf: buffer uint8
+  -> len: size_t{v len == length buf} ->
+  Stack unit
+  (requires (fun h0 ->
+    live h0 buf /\
+    disjoint buf entropy_p))
+  (ensures (fun h0 ok h1 ->
```
beurdouche, line 34: looks like you don't use `ok`...

Kachoc: Indeed, `ok` is a unit - removed it in last commit

franziskuskiefer:
> This PR introduces a blocking implementation for `Lib.RandomSequence.crypto_random`, which tries to be as close as possible to the spec, and comments `Lib.RandomBuffer.System.randombytes` as deprecated.

Is anyone still using blocking entropy sources in the real world?

Kachoc:
> > This PR introduces a blocking implementation for `Lib.RandomSequence.crypto_random`, which tries to be as close as possible to the spec, and comments `Lib.RandomBuffer.System.randombytes` as deprecated.
> 
> Is anyone still using blocking entropy sources in the real world?

It seems to be the case of Wireguard: https://git.zx2c4.com/wireguard-linux-compat/tree/src/noise.c#n493

franziskuskiefer:
> It seems to be the case of Wireguard: https://git.zx2c4.com/wireguard-linux-compat/tree/src/noise.c#n493

Interesting. Well, blocking makes sense but only shortly after boot or similar situations. In most cases this is just impossible to communicate to the user and should be handled on a higher level. Anyway, I was just curious where this is going 🙂

msprotz:
How do people normally deal with this issue? Do they just propagate a complete failure all the way up to the user if they can't find enough randomness?

Happy to hear your thoughts about API design here. I'm just not sure what is the common strategy... someone somewhere must write provisions for retrying, right?

beurdouche:
Unfortunately many people assume they got enough entropy in their output buffer. That sort of works on mainstream machines but not on smaller devices. This is why I am perfectly happy with blocking here.

franziskuskiefer:
> someone somewhere must write provisions for retrying, right?

Right, this is the tricky part. Blocking is just as bad as non-blocking without considering where it's used and how the error cases are handled. Most user space applications are fine with just not blocking because there's usually no way to recover from this. But during boot and on constraint devices someone (ideally a part that can give communicate with the user) has to call something like `wait_for_random_bytes`. This is tricky because the piece of code that's collecting entropy usually doesn't have a good way of telling the user about it. That's why I think it's important to have the entropy collection and checks for sufficient entropy separate (even if this may lead to a situation where the entropy is depleted between the two calls; applications that have to be 100% certain have to block again).

I don't want to block the change here. Just giving some food for thought on the API design 🙂

msprotz (approved):
Son and I chatted about this. Thanks @franziskuskiefer for the remarks. We actually don't know how to model something like the state of the generator and whether entropy has replenished or not with the current Low\* technology -- it might be feasible, but I don't see a trivial of doing so right now. So, seeing that this PR is a net improvement over the previous state of the API, I'm in favor of merging.

One thing that I think can still be improved is this: I think this API could distinguished between two types of errors. The first, when there's a system error (e.g. `CryptAcquireContext` returns an error), and the other, when there's not enough entropy and the number of bytes read wasn't enough and the user needs to retry. That way, the blocking API would return false only in the case of a very fatal error.

But once again, without a clear use-case in mind, it's hard to improve this API without at least some representative use-cases and a clear understanding of how we hope to do error management. So I think it's still worth merging.

PR #401, opened 2021-02-11
generic exponentiation

polubelova:
This PR introduces a low* implementation of `Lib.Exponentiation` that works for both bignum and ed25519 :)

The patch to the ed25519 code can be found here: https://github.com/project-everest/hacl-star/compare/polubelova_lexp...polubelova_ed25519. One needs to prove the properties about point addition and doubling.


code/bignum/Hacl.Impl.Exponentiation.fsti
```diff
@@ -0,0 +1,209 @@
...
+class lexp_to_exp (a_t:inttype_a) (len:size_t{v len > 0}) (ctx_len:size_t) = {
+  a_spec: Type0;
+  exp: S.exp a_spec;
+  linv_ctx: x:LSeq.lseq (uint_t a_t SEC) (v ctx_len) -> Type0;
+  linv: x:LSeq.lseq (uint_t a_t SEC) (v len) -> Type0;
+  refl: x:LSeq.lseq (uint_t a_t SEC) (v len){linv x} -> GTot a_spec;
+}
+
+
+// inline_for_extraction noextract
```
beurdouche, line 35: That would be useful no?

code/rsapss/Hacl.Spec.RSAPSS.fst
```diff
@@ -351,7 +363,13 @@ val rsapss_sign_compute_sgnt:
   -> dBits:size_nat{skey_len_pre t modBits eBits dBits}
   -> skey:lbignum t (2 * blocks modBits (bits t) + blocks eBits (bits t) + blocks dBits (bits t))
   -> m:lbignum t (blocks modBits (bits t)) ->
-  tuple2 bool (lseq uint8 (blocks modBits 8))
+  Pure (tuple2 bool (lseq uint8 (blocks modBits 8)))
```
beurdouche, line 366: Any specific reason for `tuple2` instead of the infix operator?

dist/gcc-compatible/Hacl_Bignum256.c
```diff
@@ -372,6 +331,93 @@ static inline void reduction(uint64_t *n, uint64_t nInv, uint64_t *c, uint64_t *
...
 
+static inline void
+mont_mul(uint64_t *n, uint64_t nInv_u64, uint64_t *aM, uint64_t *bM, uint64_t *resM)
+{
+  uint64_t c[8U] = { 0U };
+  uint32_t resLen = (uint32_t)8U;
+  memset(c, 0U, resLen * sizeof (uint64_t));
+  for (uint32_t i0 = (uint32_t)0U; i0 < (uint32_t)4U; i0++)
+  {
+    uint64_t uu____0 = bM[i0];
```
beurdouche, line 342: Rename?

dist/gcc-compatible/Hacl_Bignum256.c
```diff
@@ -372,6 +331,93 @@ static inline void reduction(uint64_t *n, uint64_t nInv, uint64_t *c, uint64_t *
...
+    uint64_t *res_ = c + i0;
```
beurdouche, line 343: Rename?

beurdouche (approved):
Looks really really nice :) 🥇

code/bignum/Hacl.Impl.Exponentiation.fsti
polubelova: not really, it is easier to initialize the resulting buffer with `one` before calling the exponentiation function.
for example, we can convert an integer to its montgomery representation using the precomputed constant `r2`. 
this means that we either need to store `r2` as a part of the `ctx` buffer or compute it internally (which is quite expensive).

code/rsapss/Hacl.Spec.RSAPSS.fst
polubelova: No reason :)

dist/gcc-compatible/Hacl_Bignum256.c
polubelova: yeah, can take a look

karthikbhargavan (approved):

PR #405, opened 2021-02-18
Build and config tweaks

victor-dumitrescu:
This PR is based on changes in #406.

Building hacl-star-raw fails if KREMLIN_HOME is not correctly set up. This should prevent the accidental building of slower snapshots, as we have seen in the past. I've also made some tweaks to the way the opam releases are packaged in order to avoid this in the future.


bindings/ocaml/hacl-star.opam
```diff
@@ -22,5 +22,5 @@ run-test: [
   ["dune" "runtest" "-p" name "-j" jobs]
 ]
 available: [
-  os-family != "bsd"
+  os-family != "bsd" & arch != "ppc64" & arch != "ppc32" & arch != "x86_32"
```
victor-dumitrescu, line 25: @msprotz This was [added at some point](https://github.com/ocaml/opam-repository/pull/18009) in opam-repository because the PowerPC build was failing. Do you think it's still necessary with the other changes in this PR? In particular it was failing because -mavx -mavx2 were being passed, which should no longer be the case.

msprotz (changes requested):
Hi Victor,

Thanks for this set of changes. Son and I overhauled the Makefile to be more future-proof and support new vectorized architectures in a more general way. This is in #406 -- is there any chance you could rebase your changes on top of #406? I think all of the bits about the OPAM package, checking kremlin home, adding `-thread` to `OCAMLC` are all good. Thanks!

bindings/ocaml/hacl-star.opam
msprotz: this should definitely be removed and it should work now

msprotz:
it would also be nice to try to understand why BSD is failing -- could this be detected as part of the configure script? something like "if neither windows, osx or linux, then supports_vale=false"

it would also be nice to fix up the configure script to see if the toolchain supports the intrinsic AES instructions, and if not, write something to that effect in config.h so that at least the troublesome parts of libintvector.h can be removed... but that all can be done later, of course -- it just depends on priorities!

dist/configure
```diff
@@ -296,13 +296,31 @@ if detect_arm; then
   fi
 fi
 
-if detect_x86; then
-  echo "... detected x86 platform"
+if detect_x64; then
+  echo "... detected x64 platform"
+  echo "CFLAGS_128 = -mavx" >> Makefile.config
```
franziskuskiefer, line 301: I see two issues with this approach.

This will probably work in most cases but AVX is also supported on some 32-bit systems and I'm not sure if x64 is a guarantee for avx (it's probably true in almost all cases though). But for an estimate this might be good enough.

Setting AVX on all compilation units is always a little dangerous because the compiler can add AVX instructions in parts where we don't expect it to do so. Ideally it would only be set on the files that need it. But that's a little difficult with the current setup.

franziskuskiefer (commented):
Thanks for tackling this!

> it would also be nice to fix up the configure script to see if the toolchain supports the intrinsic AES instructions, and if not, write something to that effect in config.h so that at least the troublesome parts of libintvector.h can be removed... but that all can be done later, of course -- it just depends on priorities!

Definitely something that should happen but maybe in a separate PR. The AES situation is a little more difficult I think because we'd have to remove all the AES related code/APIs as long as there's no non-NI version.

dist/configure
msprotz: I think CFLAGS_128 is appended to CFLAGS only for the files that strictly need it, see e.g. https://github.com/project-everest/hacl-star/blob/master/dist/Makefile.tmpl#L84

we have had the issue you describe it the past :( hence this trick about CFLAGS_128 appended to CFLAGS -- note that CFLAGS_128 is platform-specific and now depends on whether we're running with AVX, NEON, or soon the vector instructions for Z and Power

as to whether we can support x86 with AVX, that's another story, indeed... and probably another PR. that seems like a corner case to me but I also don't have a good grasp of these things so it could be more important than I think

msprotz (approved):
Thanks @victor-dumitrescu ! If this works for you and you can confirm on the OPAM side that all is good, then this is good for me.

victor-dumitrescu:
I'm waiting for my colleagues to re-test this on s390x but other than that I think this is good to go.

victor-dumitrescu:
This is now ready to be merged

franziskuskiefer (approved):

PR #408, opened 2021-02-24
Make the vector store_le/load_le functions more specialized

Kachoc:
This PR is necessary to add support for the IBMz vectorized implementations.

It makes a case disjunction on the type of vector elements in `store_le` and `load_le`.
The reason is that on x64 and on IBMz (which is big-endian), the consistency between the elements' order and the endianess is not the same.
For instance, if a vector 128 interpreted as an array of `uint32_t` has the following representation:
```
vector(uint32): [0x00112233, 0x44556677, 0x8899aabb, 0xccddeeff]
```

then, reinterpreted as an array of `uint64_t`, we get:
```
vector(uint64)(x64):  [0x4455667700112233, 0xccddeeff8899aabb]
vector(uint64)(IBMz): [0x0011223344556677, 0x8899aabbccddeeff]
```

When reading from bytes/writing to bytes, we thus need to know what kind of elements the vector stores, otherwise the data gets messed up.

Note that this is not a problem for the other vector operations, because even though the underlying vector type is the same, those operations are guarded by their types in the HACL code. In other words, it is not possible to use both `add32` and `add64` on the same vector, because there will be a typing issue in F*. In a sense, this PR simply makes the `store_le` and `load_le` operations consistent with the rest.

Finally, note that I added too many cases in `Lib.IntVector.fst` without the corresponding implementation in `libintvector.h`, rather than using generic `load_le`/`store_le` functions for the currently unused cases. The reason is that I think it's better to have compilation errors in the future when updating the code, rather than potentially nasty runtime errors because of endianess.

### DIFF
--- a/lib/Lib.IntVector.Intrinsics.fsti
+++ b/lib/Lib.IntVector.Intrinsics.fsti
@@ -83,10 +83,17 @@ val vec128_load32: u:uint32 -> vec128
 val vec128_load32s: x0:uint32 -> x1:uint32 -> x2:uint32 -> x3:uint32 -> vec128
 
 
-val vec128_load_le: b:lbuffer uint8 16ul -> ST vec128
+val vec128_load32_le: b:lbuffer uint8 16ul -> ST vec128
 			   (requires (fun h -> live h b))
 			   (ensures (fun h0 _ h1 ->  h1 == h0))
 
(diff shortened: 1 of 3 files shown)


Kachoc:
tagging @protz for information

beurdouche (approved):

polubelova (approved):
thanks!

PR #409, opened 2021-02-25
Fix the build for ARM

Kachoc:
Fixes the ARM neon detection in the configure file, which currently makes the build fail and was not detected by the CI before the merge of https://github.com/project-everest/hacl-star/pull/408

### DIFF
--- a/dist/configure
+++ b/dist/configure
@@ -162,8 +162,8 @@ detect_arm_neon () {
 
 int main () {
   uint8_t block[32] = { 0 };
-  Lib_IntVector_Intrinsics_vec128 b1 = Lib_IntVector_Intrinsics_vec128_load_le(block);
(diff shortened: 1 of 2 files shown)


beurdouche (approved):
Is gcc-compatible the only place where you need to update libintvector ?

Also the commit message makes no sense, there is nothing AMD related. 
x86_64?

Kachoc:
Sorry: this fixes ARM

msprotz (approved):
This looks good to me. The nightly build will refresh other copies of the configure script.

Kachoc:
> Is gcc-compatible the only place where you need to update libintvector ?
> 
> Also the commit message makes no sense, there is nothing AMD related.
> x86_64?

The important file is `dist/configure`, `dist/gcc-compatible` was updated only for a CI test, and the other `configure` files will be refreshed during the build.

PR #410, opened 2021-02-26
Generalize the build system for Vec128/256 and implement vectorized support for IBM SystemZ

Kachoc:
The following PR generalizes the build system and EverCrypt to target several architectures. Building on this addition, this PR also implements support for IBM SystemZ.

# Build system
The build system updates are made in several steps.

First, we moved as many macros as possible from `evercrypt_targetconfig.h` to the `configure` file, like the ones defining the target architecture.

We also renamed some macros, like `BROKEN_INTRINSICS` -> `COMPILE_INTRINSICS` (and applied/removed the proper negations) to make it clear that the macros guarding the code are purely compilation macros. In particular, `COMPILE_128` and `COMPILE_256` mean that we build the code for targets (potentially) supporting vectors 128/256: the current platform does not necessarily support them, and we may need to do dynamic checks for hardware support before calling the relevant functions.

We removed, whenever possible, the checks of the form `EverCrypt.TargetConfig.x64 && ... (* additional dynamic checks for hardware support *)`, to replace them with more general checks like `EverCrypt.TargetConfig.compile_128 && ... (* dynamic checks *)`which make the target architectures more abstract. Note that those extract, like before, to code of the form:
```
#if COMPILE_128
  if ... (* dynamic checks *) {
  ...
}
#endif
```

# 128-bit vector support for IBM SystemZ
This PR also adds support for the intrinsics of IBM SystemZ, for the versions >= z13. In order to make it easy to add support for more architectures, we also implemented various testing and debugging utilities for the vectorized functions.

The `tests` directory now contains a `vec128-test.c` file which contains very thorough tests for all the 128-bits vectorized functions currently used in Hacl. The tests were designed in order to be easy to debug, so that one can fix his implementations without much effort by looking at the tests output.

We modified the `libintvectors.h` file so that it is now possible to switch to a debugging mode by providing the proper compiler flags. This mode prints a debugging trace for the vectorized operations, in order to make it easy to understand why a primitive's output is not correct by comparing traces, in case the above test file was not enough. As of now, all the 128-bit functions used in Hacl output useful information when compiling with the debug mode flags.

# For future PRs
We leave as future work the addition of test vectors and debugging implementations for the functions which don't have any so far, that is to say the 128-bit functions which are not currently used in Hacl, and the 256-bit functions which I could not test with the machines I had at my disposal.


dist/gcc-compatible/evercrypt_targetconfig.h
```diff
@@ -6,27 +6,22 @@
...
+  // TODO: the macros defined in EverCrypt.TargetConfig.h don't extract as
+  // uppercase - fix that
+  target_architecture_name_x86,
+  target_architecture_name_x64,
+  target_architecture_name_arm7,
+  target_architecture_name_arm8,
+  target_architecture_name_systemz,
+  target_architecture_name_powerpc64,
+  target_architecture_name_unknown
+};
```
beurdouche, line 19: If you can extend Kremlin to extract uppercase if something like `[@CMacro]` is above the definition, that would be great.

providers/evercrypt/EverCrypt.TargetConfig.fsti
```diff
@@ -1,27 +1,61 @@
...
+/// completely by hand. Also note that we don't need any of the properties
+/// which would be provided by an enum in our proofs (we don't need to
+/// know that, say, [target_architecture_name_x86 <> target_architecture_name_z]).
+/// Finally, we use the prefix "target_architecture_name" to make sure the generated C
+/// macros don't clash with any other names, and that there is no ambiguity
+/// on their meaning - e.g. without the prefix, one may think [X86] is a flag
+/// and not a constant, and thus use it to guard some code working only on x86
+/// with [#if defined(X86)].
+
+// TODO: the below macros extract to C with lowercase names: fix that
```
beurdouche, line 15: Ahaha :)

bindings/ocaml/AutoConfig2.ml
```diff
@@ -17,8 +17,8 @@ let has_feature = function
   | SHAEXT -> everCrypt_AutoConfig2_has_shaext ()
   | AES_NI -> everCrypt_AutoConfig2_has_aesni ()
   | PCLMULQDQ -> everCrypt_AutoConfig2_has_pclmulqdq ()
-  | AVX -> everCrypt_AutoConfig2_has_avx ()
-  | AVX2 -> everCrypt_AutoConfig2_has_avx2 ()
+  | VEC128 -> everCrypt_AutoConfig2_has_vec128 ()
```
beurdouche, line 20: I am a bit skeptical about this renaming but ok, generally I agree with you that this shouldn't be architecture specific...

beurdouche (approved):

beurdouche:
Leaving an accept here so that you can merge easily after the next batch of changes...

Kachoc:
> Leaving an accept here so that you can merge easily after the next batch of changes...

Thanks a lot ! :)

beurdouche (changes requested):
One thing I didn't notice before is that some of the macro names are affecting not only the Evercrypt source but also the HACL*. Those files being consumed by many other clients need to have macro names identifiable such as: `<PROJECT>_SUPPORT_VEC128`, `HACL_SUPPORT_INLINE_ASM`, `EVERCRYPT_SUPPORT_VALE`because `COMPILE_128` doesn't mean much...

msprotz:
Thanks Benjamin, that's a sensible suggestion.

One thing I mentioned to Son earlier is that we should try to move away from "support" because it's too imprecise. There's actually two notions:
- static support at compile-time for generating code for e.g. vec128 or vale -- determined by the configure script, and generating a macro
- support at run-time for executing such code determined by a runtime test.

On x64 especially, the two are different notions, so my suggestion was to use "compile" rather than support for the former. So, maybe, EVERCRYPT_CAN_COMPILE_128 and so on, since all of the implementation multiplexing is done at the evercrypt level.

beurdouche:
Ok, "support" is kind of standard because it is quite rare to encounter these kind of macros for runtime checks but if you really prefer, that's not a blocker. Call it `EVERCRYPT_CAN_COMPILE_VEC128` and we have a deal ;)

Kachoc:
Ok, deal! Let me update the code :)

beurdouche (approved):
Thanks

providers/evercrypt/fst/EverCrypt.Poly1305.fst
```diff
@@ -126,17 +126,26 @@ let poly1305 dst src len key =
...
-  end else if EverCrypt.TargetConfig.x64 && avx then begin
+  end else if EverCrypt.TargetConfig.evercrypt_can_compile_vec128 && vec128 then begin
     Hacl.Poly1305_128.poly1305_mac dst len src key
 
-  end else if EverCrypt.TargetConfig.x64 && vale then begin
-    poly1305_vale dst src len key
+  end else if EverCrypt.TargetConfig.evercrypt_can_compile_vale && vale then begin
 
-  end else begin
+    if EverCrypt.TargetConfig.evercrypt_can_compile_vec128 then begin
```
msprotz, line 141: I feel like this test here should reference specific CPU flags

also, wants_vale is perhaps legacy and we should consider deprecating it -- it was useful for tests, but frankly, there's just too many flags now

tests/Makefile
```diff
@@ -33,6 +53,15 @@ arm: chacha20-arm-test.test poly1305-arm-test.test blake2-arm-test.test curve255
...
 curve64-rfc.exe: $(patsubst %.c,%.o,$(wildcard rfc7748_src/*.c))
 
+# Note that vec-128-test.exe uses lib/c and not dist/gcc-compatible:
+# this allows to work on and test the vectorized instructions without
+# rebuilding the whole HACL library.
+vec-128-test.exe: ../lib/c/libintvector.h vec-128-test.c
+	cc -I../lib/c -O3 $(CFLAGS) $(CFLAGS_128) vec-128-test.c -o vec-128-test.exe
+
+../lib/c/libintvector.h:
+	touch ../lib/c/libintvector.h
```
msprotz, line 63: ?!!

msprotz (approved):
This looks good to me minus this: I'm not sure @karthikbhargavan and/or @polubelova who both worked on libintvector.h will be very happy with the big debug info. They might prefer to leave that sort of debugging to a branch.

As I mentioned somewhere else it might be good to separate the tests by feature, not by platform, and get rid of the copy-paste for ARM tests, but that can happen later.

Thanks!

beurdouche:
> This looks good to me minus this: I'm not sure @karthikbhargavan and/or @polubelova who both worked on libintvector.h will be very happy with the big debug info. They might prefer to leave that sort of debugging to a branch.

This is not blocking for me, but my preference is also not to have debug inside this file

providers/evercrypt/fst/EverCrypt.Poly1305.fst
Kachoc: Yes indeed, this only depends on Vale. I'm however not completely sure about what to do for the `wants_vale` functions. I added the `deprecated` attribute, but I don't know if I can completely remove all the occurrences.

msprotz: Frankly I think we should leave this to a further cleanup. Remove: wants_vale, StaticConfig, and all of that stuff that predates the nice configuration infrastructure that you instated and materialized in this PR. If you start this now, you'll have to fix a bunch of tests, which is not very fun.

Kachoc:
I agree that it is better if libintvector.h doesn't contain debugging related facilities. However I would like to preserve the debugging code on main and not on a quickly forgotten branch, because they took a bit of time to be written, and actually proved quite useful.
Would it be acceptable if `libintvector.h` contains the following:
```
#if __has_include("libintvector_debug.h")
#include "libintvector_debug.h"
#endif
```
Where `libintvector_debug.h` would be a file only present in `lib/c` which would be manually copied to `gcc-compatible` in case we need to debug the vectorized operations? (from there, I can get the switch to work with the appropriate macros).
I might then add a file in `tests` to use the debugging facilities and make sure they are maintained.

tests/Makefile
Kachoc: Removed that

Kachoc:
I removed the debug functions from `libintvector.h`: will put that in the next PR for PowerPC

msprotz:
@victor-dumitrescu would you mind checking that you are now getting the optimized 128-bit versions of the algorithms with the OCaml bindings?

victor-dumitrescu:
Hi, I think this works as intended. I asked someone with access to a machine to install this as an opam package and indeed 128-bit files are compiled with `-m64 -mzarch -mvx -mzvector -march=z13` and are included and pass the tests in the OCaml API.

For reference, this is the output of the configure script:
```
- ... detected z platform
- ...  supports 128-bit z VSX
-  does not support x64 assembly, disabling Curve64
-  does not support legacy vale stubs
-  does not support 256-bit arithmetic
- ... found ocamlfind in /home/linux1/.opam/tezos_opam_install/bin/ocamlfind
- ... found ctypes in /home/linux1/.opam/tezos_opam_install/lib/ctypes
- ... found ctypes.foreign in /home/linux1/.opam/tezos_opam_install/lib/ctypes
- ... glibc is recent enough for explicit_bzero
```

msprotz:
excellent! and which algorithms are you calling on Z? are you going through the evercrypt layer for those?

victor-dumitrescu:
Do you mean specifically in Tezos? Or in the OCaml API in general?

msprotz:
Sorry I meant in Tezos

victor-dumitrescu:
I think none of the 128-bit algorithms are used either directly or indirectly in Tezos. We use the EverCrypt layer for hashing, HMAC, and Curve25519. Otherwise, we also use Blake2b_256 when possible. We also use NaCl so in theory we could use Poly1305_128 but I don't think that's the case with the current NaCl API.

msprotz:
Ok good to know. Thanks!

Issue #327, opened 2020-07-14
undefined behavior in gcc-compatible detected by ubsan


aep:
ubsan is unhappy about a bunch of things. i'm unsure if these are due to incorrect usage? The C output is a bit difficult to read.

```
3rdparty/hacl-star/dist/gcc-compatible/Hacl_Chacha20Poly1305_32.c:32:23: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior 3rdparty/hacl-star/dist/gcc-compatible/Hacl_Chacha20Poly1305_32.c:32:23 in 
3rdparty/hacl-star/dist/gcc-compatible/Hacl_Chacha20Poly1305_32.c:280:15: runtime error: null pointer passed as argument 2, which is declared to never be null
```

beurdouche:
The code is voluntarily not defensive, there is a precondition that is that you should not be passing a null pointer to that function. I think this should be better documented though.

aep:
i wasnt even are there _is_ documentation for the C api. could you point me to it?

beurdouche:
https://hacl-star.github.io/HaclAEAD.html

msprotz:
@beurdouche I took a look and I couldn't see the non-null precondition here: did you find it somewhere? https://github.com/project-everest/hacl-star/blob/master/code/chacha20poly1305/Hacl.Impl.Chacha20Poly1305.fst#L151

I should have given an update on this: we took a long hard look at this issue back when you reported it, and actually found there is an issue: our model of C, Low\*, allows the user to successfully perform `NULL+0`, on the basis that `NULL` is a zero-length array and that `+0` is pointer arithmetic that is within bounds. Unfortunately, as you accurately found out, this is illegal in C.

Alas, this occurs everywhere an input may have length 0: it is idiomatic, even reasonable, for C callers to pass NULL when they want to have, say, no additional authenticated data. And it is not reasonable to ask C callers to heap-allocate a dummy just to call our functions. (Note: zero-length stack allocations are UB, so callers can't even do that as a dummy for the aad.)

Writing a defensive API would amount to checking whether aad has length zero, heap-allocating a dummy, calling the HACL functions, freeing the dummy, returning the result. This is clearly too much.

[...]

PR #413, opened 2021-03-04
Improvement to Chacha20Poly1305

beurdouche:


beurdouche:
I've been able to test it locally and it passes the tests we have.

polubelova (approved):

polubelova:
No surprise that it passes the tests. The fix is correct as `poly1305_update` returns the initial `acc` value if the input length is 0.
thanks @beurdouche!

msprotz:
Can you add a C test that would make sure this case is caught? It's slightly worrying that this issue wasn't caught earlier -- it'd be good to catch this sort of bugs before they make it over to NSS.

polubelova:
It's related to this one: https://github.com/project-everest/hacl-star/issues/327

msprotz:
the UB is only in low-level C code, but here the spec too is changed, so it seems like there was a spec bug to begin with?

polubelova:
One could just prove the following lemma:
```
val poly1305_padded_lemma:
    r_elem:Poly.felem
  -> text:bytes
  -> acc:Poly.felem -> Lemma
  (requires length text = 0)
  (ensures poly1305_padded r_elem text acc == acc)
```
With that, we won't need to change the spec.

msprotz:
Two questions then.
- Why make the spec more complicated if this is just an implementation matter? The spec should remain as small as possible so that it can be audited.
- Does this mean that the issue from #327 resulted in a mis-compilation that was observed with NSS?

beurdouche:
The spec should have been written this way from the beginning, it is an artifact of someone wanting to reuse the same code too aggressively without branching on `aadlen`. The semantics is if the length of AAD is zero, this function should not have been called at all.

prosecco:
I don’t see a need to change the high-level spec, but this change is innoccuous.
Our specs follow the pseudocode in the RFC, e.g. https://tools.ietf.org/html/rfc7539#section-2.8.1
The check on aad_len allows us to stlightly optimize, but does not change the functionality.
In the low-level code this optimization further allows us to pass in null pointers.
So, this change is essentially about helping out low* code do something safe, it is not a functional change in the spec.
Am I right?

beurdouche:
Correct, prevents the Low* code to do stack allocation and pointer arithmetic at execution for something that is gonna be discarded because of length zero. It doesn't change the functional behavior.

karthikbhargavan (approved):

PR #414, opened 2021-03-05
P256 spec refactor

karthikbhargavan:
This PR refactors the p256 specs to remove a dependency on Lib.Buffer.
This fixes the local Makefile in specs/.
A more comprehensive restructuring of the P256 code to support multiple curves is in the works and will appear in a future PR.


msprotz:
I would try something like:
```
git reset origin/master dist hints
git commit -m "don't push hint & dist changes"
```
or something along those lines (if you want a mergeable PR).

Other than that, looks great, thanks. Can you try removing this file?
```
jonathan@absinthe:~/Code/everest/hacl-star (master) $ cat !$
cat lib/ml/Lib_Memzero0.ml
let memzero _ _ =
  failwith "this is only here because of spurious dependencies from Spec to Lib.Buffer"
```
it was there only because of that dependency. Also: https://github.com/project-everest/hacl-star/blob/master/Makefile#L476 you should be able to remove this while at it

karthikbhargavan:
Thanks Jonathan, I made those changes and re-pushed.

msprotz (approved):
👍

Frosne (approved):
Seems good :)

PR #416, opened 2021-03-08
Vectorized functions for PowerPC64 (little-endian)

Kachoc:
This PR implements vectorized functions for PowerPC64 (little-endian only).
It also adds debugging utilities for the vectorized implementations.

The implementation of the vectorized functions mostly affects `lib/c/libintvector.h` and `dist/configure`.

The debugging utilities come from [this PR](https://github.com/project-everest/hacl-star/pull/410): they were then removed in order not to pollute `libintvector.h`. I moved the utilities to `lib/c/libintvector_debug.h` and updated the `Makefile` so that `libintvector_debug.h` doesn't get copied to the target directories in `dist/`. In case one wants to debug an implementation, it suffices to copy the file to the proper target directory and use some compilation flags (the steps are explained in the comments). The lines added to `libintvector.h` to make this work are kept to the bare minimum.

On a side note, the `tests/test_helpers.h` file doesn't implement the cpu cycles functions for PowerPC64, to compute the number of cpu cycles consumed per processed byte. Computing the cpu cycles is not very difficult, but requires a bit of setup (see the comments in the file). Note that I can add the required implementation it if a reviewer deems it necessary. Also note that `test_helpers.h` still computes the implementations' speed (in seconds).

I also took the opportunity to make minor modifications to code related to SystemZ (nothing important).


Kachoc:
@polubelova @karthikbhargavan you might want to have a look at the code I added to `libintvector.h` to make the debugging utilities work. It's quite small and simple, but I just want to make sure you have no problem with it and it doesn't pollute the file.

dist/c89-compatible/libintvector.h
```diff
@@ -792,6 +780,179 @@ typedef vector128_8 vector128;
...
+// unsigned int, unsigned long long: the compiler complains that the parameter
+// combination is invalid.
+typedef vector unsigned char vector128_8;
+typedef vector unsigned int vector128_32;
+typedef vector unsigned long long vector128_64;
+
+typedef vector128_8 Lib_IntVector_Intrinsics_vec128;
+typedef vector128_8 vector128;
+
+// Small helper to change the endianess of the vector's elements, seen as uint32.
```
msprotz, line 800: typo: endianness

msprotz (approved):
This looks good to me. @Kachoc and I had previously spent a while looking at this together so I'm good with the PR in its current form. High-level questions / remarks:
- do we have a story for CI testing? I am pretty sure that unless we perform some modicum of testing this is going to regress real fast; can Nomadic Labs help with this? this also applies to Z (note: github actions won't work: https://github.com/actions/runner/issues/243)
- perhaps worth asking David if there's any way to work around the memcpys, or maybe it won't affect the performance too much, I don't know
- I don't see anything related to splitting tests according to their feature requirements, and updating the test Makefile accordingly... was this something we thought was worth doing?

Thanks for the PR!

dist/c89-compatible/libintvector.h
Kachoc: Fixed

Kachoc:
> This looks good to me. @Kachoc and I had previously spent a while looking at this together so I'm good with the PR in its current form. High-level questions / remarks:
> 
>     * do we have a story for CI testing? I am pretty sure that unless we perform some modicum of testing this is going to regress real fast; can Nomadic Labs help with this? this also applies to Z (note: github actions won't work: [actions/runner#243](https://github.com/actions/runner/issues/243))
> 
>     * perhaps worth asking David if there's any way to work around the memcpys, or maybe it won't affect the performance too much, I don't know
> 
>     * I don't see anything related to splitting tests according to their feature requirements, and updating the test Makefile accordingly... was this something we thought was worth doing?
> 
> 
> Thanks for the PR!

The tests/Makefile filters the targets which only use vec128/256, etc. but some tests file use scalar *and* vectorized implementations. I can split those files if needs be.

Kachoc:
I split the test files according to the feature requirements. I'm not sure what to do with the arm tests, though. I think a proper way would be to filter them from the `all` target, so that they would be only generated and run with `make arm`, and then I could remove the preprocessing guards inside the files. But I'm not sure if the CI would still test those files, then.

msprotz:
why don't we have `test-c`, `test-128` and `test-256` targets and then the main Makefile (possibly with a shell-script, or something) reads Makefile.config (or config.h) and invokes suitable targets?

protz:
I tried this on my Mac M1 and `make test` seems to work just fine.

msprotz:
Screen Shot 2021-06-10 at 3 16 09 PM


This was the error but I guess you couldn't see it because you didn't have access to the Azure stuff.

msprotz:
The issue was in `make dist/test` i.e. compiling & running not the hand-written C tests, but those written in Low* then extracted to C. It's been fixed on the kremlin side.

msprotz:
@Kachoc can you push a dummy commit? I can't seem to figure out how to restart the drone builds

polubelova (approved):

PR #411, opened 2021-03-02
share bignum25519_51 between curve25519 and ed25519

polubelova:
This PR eliminates some redundancy in the ed25519 code and introduces a static header file for bignum25519_51.

PR #418, opened 2021-03-10
Don't inline point_negate

franziskuskiefer:
I'm not entirely sure why, but in-lining the point negation breaks the Windows build. The function is only declared as extern but compiled (optimized?) out.

> 01F 00000000 UNDEF  notype ()    External     | Hacl_Impl_Ed25519_PointNegate_point_negate

You can also see the error on the `evercrypt-sys` build: https://github.com/franziskuskiefer/evercrypt-rust/runs/2075276459
Not in-lining fixes this for me locally but I'm happy to consider other solutions.

Note that our CI here currently doesn't catch this because it only builds the library and doesn't actually use it.

### DIFF
--- a/code/ed25519/Hacl.Impl.Ed25519.PointNegate.fst
+++ b/code/ed25519/Hacl.Impl.Ed25519.PointNegate.fst
@@ -23,7 +23,6 @@ val point_negate: p:F51.point -> out:F51.point ->
     F51.point_inv_t h1 out /\
     F51.point_eval h1 out == SC.((-x1) % prime, y1, z1, (-t1) % prime)))
 
-[@CInline]
 let point_negate p out =
   push_frame ();
   let zero = create 5ul (u64 0) in


msprotz:
There's a kremlin warning in the build for that:

```
Warning 8: : After inlining, Hacl.EC.Ed25519.point_negate (going into Hacl_EC_Ed25519) calls Hacl.Impl.Ed25519.PointNegate.point_negate (going into Hacl_Ed25519). This is a call across translation units but Hacl.Impl.Ed25519.PointNegate.point_negate has a C "inline" qualifier. The C standard allows removing Hacl.Impl.Ed25519.PointNegate.point_negate from its translation unit (see C11 6.7.3 §5), and CompCert will do it. 
```

Apparently, MSVC will do it too.

msprotz (approved):

msprotz:
Thanks for a good catch!

franziskuskiefer:
> There's a kremlin warning in the build for that:

Ah missed that. But totally makes sense :)

msprotz:
Oh yeah I don't expect anyone to pay attention to that stuff. There's dozens of them. Kremlin generally doesn't remove the inline qualifier (even though kremlin knows it's likely to cause issues) because as you noticed, it doesn't seem to be triggered every single time on mainstream compilers.

polubelova (approved):
thanks @franziskuskiefer!

Let me check whether we need to remove the inline qualifier from other functions (looking at the kremlin output in code/ed25519).

franziskuskiefer:
This was the only one breaking `evercrypt-sys` but there might be more.

polubelova:
I think we can safely remove the inline qualifier from the following functions:
```
Hacl.Bignum25519.reduce_513 
Hacl.Bignum25519.inverse
Hacl.Bignum25519.load_51 
Hacl.Bignum25519.store_51 
Hacl.Impl.Ed25519.PointNegate.point_negate 
Hacl.Impl.Ed25519.PointAdd.point_add
Hacl.Impl.Ed25519.Ladder.point_mul 
Hacl.Impl.Ed25519.PointEqual.point_equal
Hacl.Impl.Ed25519.PointCompress.point_compress 
Hacl.Impl.Ed25519.PointDecompress.point_decompress
```
The performance regression was due to another reason which was fixed by this PR: https://github.com/project-everest/hacl-star/pull/411.
Should I raise another PR that does above?

franziskuskiefer:
Generally the compiler should be intelligent enough to inline these if necessary. Yeah open another PR, I'm happy to approve 🙂

PR #419, opened 2021-03-10
remove cinline from some functions in ed25519

polubelova:
This PR is a follow-up to https://github.com/project-everest/hacl-star/pull/418

### DIFF
--- a/code/ed25519/Hacl.Bignum25519.fst
+++ b/code/ed25519/Hacl.Bignum25519.fst
@@ -92,7 +92,6 @@ let carry51 l cin =
   (l' &. mask_51, l' >>. 51ul)
 
 
-[@CInline]
 let reduce_513 a =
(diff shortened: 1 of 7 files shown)


msprotz (approved):

beurdouche:
I am curious about the performance regression for this, any idea ?

franziskuskiefer (approved):
I don't expect this to have a performance impact (smart compilers and all) but if we want to be 100% certain we'd of course have to test it.

beurdouche:
That would be surprising, don’t we use load/store a lot?

polubelova:
Most of these functions (except point_add and reduce_512) are used once. 
There is no perf diff if the code compiled with clang-9, just checked.

franziskuskiefer:
> That would be surprising, don’t we use load/store a lot?

The compiler inlines if it thinks it makes sense. And probably does every time here, hence no performance difference because the actual result's probably the same.

Issue #420, opened 2021-03-10
Typedef errors building on FreeBSD


web-vertalo:
Can someone provide guidance on how to resol.ve these errors?

```./libintvector.h:33:17: error: cannot combine with previous 'type-name' declaration specifier
typedef __m128i Lib_IntVector_Intrinsics_vec128;
                ^
./config.h:1:41: note: expanded from macro 'Lib_IntVector_Intrinsics_vec128'
#define Lib_IntVector_Intrinsics_vec128 void *
                                        ^
In file included from Hacl_Spec.c:25:
In file included from ./Hacl_Spec.h:33:
./libintvector.h:33:48: error: expected identifier or '('
typedef __m128i Lib_IntVector_Intrinsics_vec128;
                                               ^
./libintvector.h:232:17: error: cannot combine with previous 'type-name' declaration specifier
typedef __m256i Lib_IntVector_Intrinsics_vec256;
                ^
./config.h:2:41: note: expanded from macro 'Lib_IntVector_Intrinsics_vec256'
#define Lib_IntVector_Intrinsics_vec256 void *
                                        ^
In file included from Hacl_Spec.c:25:
In file included from ./Hacl_Spec.h:33:
./libintvector.h:232:48: error: expected identifier or '('
typedef __m256i Lib_IntVector_Intrinsics_vec256;
                                               ^
4 errors generated.
```

msprotz:
What is the output of the configure script on your platform, as executed in `dist/gcc-compatible`, and for the very latest revision on master?

web-vertalo:
Output from, dist/gcc-compatible/configure:
```
 does not support x64 assembly, disabling Curve64
 does not support legacy vale stubs
cc  cannot compile 128-bit vector arithmetic, disabling
 does not support 256-bit arithmetic
... found ocamlfind in /usr/local/bin/ocamlfind
ocamlfind: Package `ctypes' not found
OCaml package ctypes not found
OCaml bindings disabled
```

Evidently $build_target is not set to something sensible.

The specific errors reported are reformatted in the latest version, though essentially unchanged:
```
In file included from ./Hacl_Spec.h:33:
./libintvector.h:232:48: error:
      expected
      identifier
      or '('
  ...;
     ^
```

msprotz:
interesting, because since your toolchain does not seem to support 128-bit vector arithmetic, we should suitably disable both 128-bit and 256-bit vector primitives... @Kachoc do you have a clue?

[...]

PR #421, opened 2021-03-10
Update libintvector.h to guard the code with the flags generated by the configure script

Kachoc:
Everything is in the title: the definitions in libintvector.h are currently guarded by compiler flags, which proved to be incompatible with the flags defined by the configure script as reported here: https://github.com/project-everest/hacl-star/issues/420
This PR fixes the issue.


msprotz:
@web-vertalo could you tell us if this fixes your issue?

Kachoc:
Note that you should be able to build `dist/gcc-compatible` (but only this target - the others need to be regenerated)

beurdouche (changes requested):
I don’t think that works. Those macros don’t exist in libraries that don’t use the configure script.

web-vertalo:
This change fixes my issue.

msprotz:
@beurdouche how does NSS compile on FreeBSD then? these C preprocessor checks (based on @web-vertalo's example) don't guarantee the availability of the vector types

Kachoc:
Yes indeed, forgot about the fact that not all the targets use the `configure`. I suggest adding a hand-written `config.h` file to the targets which don't use `configure`, to define the proper flags (potentially using compiler flags) for those targets. We would also clearly specify which hypotheses we do for those targets.

Kachoc:
I modified the `libintvector.h` code to use the original compiler flags to find the target architecture, but kept the `EVERCRYPT_CAN_COMPILE_VEC128/256` guards.
@web-vertalo does it still compile on your side?
@beurdouche would it be acceptable for you to add `-DEVERCRYPT_CAN_COMPILE_VEC128` to the C compilation flags on your side from now onwards?

web-vertalo:
The include of libintvector.h no longer throws an error.

Attempts to build under this branch with detect_x64 returning false leads to errors like this for vec128 and vec256:

```
In file included from ./Hacl_Chacha20Poly1305_128.h:42:
./Hacl_Poly1305_128.h:43:44: error: unknown type name
      'Lib_IntVector_Intrinsics_vec128'
Hacl_Impl_Poly1305_Field32xN_128_load_acc2(Lib_IntVecto...
                                           ^
./Hacl_Poly1305_128.h:47:3: error: unknown type name
      'Lib_IntVector_Intrinsics_vec128'
  Lib_IntVector_Intrinsics_vec128 *out,
  ^
```

But this is now a non-issue for my setup. The underlying problem was that I had not updated detect_x64 to recognize amd64 in $target_arch as providing support for the vector types. I submitted PR #424 which includes an update to detect_x64.

msprotz:
> Attempts to build under this branch with detect_x64 returning false leads to errors like this for vec128 and vec256:
> 
> ```
> In file included from ./Hacl_Chacha20Poly1305_128.h:42:
> ./Hacl_Poly1305_128.h:43:44: error: unknown type name
>       'Lib_IntVector_Intrinsics_vec128'
> Hacl_Impl_Poly1305_Field32xN_128_load_acc2(Lib_IntVecto...
>                                            ^
> ./Hacl_Poly1305_128.h:47:3: error: unknown type name
>       'Lib_IntVector_Intrinsics_vec128'
>   Lib_IntVector_Intrinsics_vec128 *out,
>   ^
> ```
> 
It's ok, the script is just detecting that (assuming this is a non-x64 platform) the 128 and 256-bit vector instructions are not supported. The errors from the feature tests are not hidden.

@Kachoc let me know if you need help figuring out the build failures

msprotz (approved):
This looks good to me (provided we figure out the build failures).

@beurdouche is it reasonable for the build system of Mozilla to add `-DEVERCRYPT_CAN_COMPILE_VEC128` and `-DEVERCRYPT_CAN_COMPILE_VEC256`? The idea is that this is something to be determined "outside" of the preprocessor. For instance, once we have AVX512, we'll have to run configure-time checks to see if the toolchain is recent enough to even have the required headers to compile these things; and so, it's inevitable in the long run that you'll either have to do these checks on the Mozilla side, too, or alternatively request minimum compiler versions and unconditionally define those. Thoughts?

msprotz:
Looked at this: the configure script needs `-DEVERCRYPT_CAN_COMPILE_128` here: https://github.com/project-everest/hacl-star/blob/master/dist/configure#L104

msprotz:
```
/drone/src/dist/gcc-compatible /drone/src
667s
316... detected ARM platform
667s
317/tmp/testvec128.0Vm4ZAX.c: In function 'main':
667s
318/tmp/testvec128.0Vm4ZAX.c:4:3: error: unknown type name 'uint8_t'; did you mean 'u_int8_t'?
667s
319   uint8_t block[32] = { 0 };
667s
320   ^~~~~~~
667s
321   u_int8_t
667s
322/tmp/testvec128.0Vm4ZAX.c:5:3: error: unknown type name 'Lib_IntVector_Intrinsics_vec128'; did you mean 'Lib_IntVector_Intrinsics_bit_mask64'?
667s
323   Lib_IntVector_Intrinsics_vec128 b1 = Lib_IntVector_Intrinsics_vec128_load32_le(block);
667s
324   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
667s
325   Lib_IntVector_Intrinsics_bit_mask64
667s
326/tmp/testvec128.0Vm4ZAX.c:5:40: warning: implicit declaration of function 'Lib_IntVector_Intrinsics_vec128_load32_le'; did you mean 'Lib_IntVector_Intrinsics_bit_mask64'? [-Wimplicit-function-declaration]
667s
327   Lib_IntVector_Intrinsics_vec128 b1 = Lib_IntVector_Intrinsics_vec128_load32_le(block);
667s
328                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
667s
329                                        Lib_IntVector_Intrinsics_bit_mask64
667s
330/tmp/testvec128.0Vm4ZAX.c:6:3: error: unknown type name 'Lib_IntVector_Intrinsics_vec128'; did you mean 'Lib_IntVector_Intrinsics_bit_mask64'?
667s
331   Lib_IntVector_Intrinsics_vec128 b2 = Lib_IntVector_Intrinsics_vec128_load32_le(block + 16);
667s
332   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
667s
333   Lib_IntVector_Intrinsics_bit_mask64
667s
334/tmp/testvec128.0Vm4ZAX.c:7:3: error: unknown type name 'Lib_IntVector_Intrinsics_vec128'; did you mean 'Lib_IntVector_Intrinsics_bit_mask64'?
667s
335   Lib_IntVector_Intrinsics_vec128 test = Lib_IntVector_Intrinsics_vec128_interleave_high64(b1, b2);
667s
336   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
667s
337   Lib_IntVector_Intrinsics_bit_mask64
667s
338/tmp/testvec128.0Vm4ZAX.c:7:42: warning: implicit declaration of function 'Lib_IntVector_Intrinsics_vec128_interleave_high64'; did you mean 'Lib_IntVector_Intrinsics_bit_mask64'? [-Wimplicit-function-declaration]
667s
339   Lib_IntVector_Intrinsics_vec128 test = Lib_IntVector_Intrinsics_vec128_interleave_high64(b1, b2);
667s
340                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
667s
341                                          Lib_IntVector_Intrinsics_bit_mask64
```

beurdouche:
I have discussed this quickly with Marina again. I think I prefer the style where we have `DISABLE_XYZ` flags so that if those flags are not set, we will compile by default. This is the style we have in NSS and most products I know https://searchfox.org/nss/source/lib/freebl/chacha20poly1305.c#252. In some sense, yes it forces you to update your compiler, but I think it is preferable.

msprotz:
@beurdouche this contradicts your earlier comment in another PR that says "nobody runs the configure script" -- if we implement what you describe, then by default, code will not compile unless you happen to be on x64 with a modern enough compiler to target everything we need. Anything else (ARM, Z, Power, etc.) will cease to work unless users run the configure script.

Another issue I see with your suggestion is that it's not future-proof, i.e. when we add AVX512, your code will break when you pull in the latest HACL* code, and then the burden will be on you to add DISABLE_AVX512 in relevant cases (e.g. for NSS compiling to ARM and all other platforms). If you add a new platform to NSS, you'll also have to add DISABLE_* flags rather than assume it compiles safely by default.

As a verified project, I think it's important to have safe defaults, and state things such as "I assert that my target supports 128-bit vectorized code", rather than rely on implicit assumptions such as "I assert that my target doesn't support 256-bit code and I know that currently 512-bit has not yet landed in HACL so things will work out for the time being".

What do you think?

msprotz:
hi @beurdouche have you had some time to think about this? I had a separate discussion with Son who was also in favor of safe, future-proof defaults -- I've taken a look at this and to restore the existing behavior, all that is needed would be a `config.h`  in your directory that contains:
```
#define EVERCRYPT_CAN_COMPILE_INTRINSICS 1
#define EVERCRYPT_CAN_COMPILE_VALE 1
#define EVERCRYPT_CAN_COMPILE_INLINE_ASM 1
#define EVERCRYPT_CAN_COMPILE_VEC128 1
#define EVERCRYPT_CAN_COMPILE_VEC256 1
```
and this will enable all of the optimizations -- actually, I believe that if you don't have that already, you may be missing out on the optimized add_carry/sub_borrow used for P256, which have already been set to a safe default...

let me know if I can help -- happy to chat about this with you 1:1

thanks,

jonathan

.ci/script.bat
```diff
@@ -1,7 +1,7 @@
 call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\Tools\VsDevCmd.bat" -host_arch=amd64 -arch=amd64
 call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\Tools\VsDevCmd.bat" -test
 cd dist/msvc-compatible
-cl *.c /I ../kremlin/include /I . /I ../kremlin/kremlib/dist/minimal /c || goto :error
+cl *.c /I ../kremlin/include /I . /I ../kremlin/kremlib/dist/minimal /c /DEVERCRYPT_CAN_COMPILE_INTRINSICS /DEVERCRYPT_CAN_COMPILE_VALE /DEVERCRYPT_CAN_COMPILE_VEC128 /DEVERCRYPT_CAN_COMPILE_VEC256 || goto :error
```
franziskuskiefer, line 4: We should probably start using configure for Windows as well at some point. But that's for a different PR.

lib/c/libintvector.h
```diff
@@ -3,28 +3,35 @@
...
+ * a dependency). When comparing traces, note that some instructions are not
+ * compiled in the same order on the different platforms, but it doesn't lead
+ * to a lot of discrepancies in practice. */
 
 #define Lib_IntVector_Intrinsics_bit_mask64(x) -((x) & 1)
 
 #if defined(__x86_64__) || defined(_M_X64)
 
-// The following functions are only available on machines that support Intel AVX
+#if defined(EVERCRYPT_CAN_COMPILE_VEC128)
```
franziskuskiefer, line 34: This said AVX support before and now vec128 support.
Looking at the `detect_uint128` in `configure` that sets `EVERCRYPT_CAN_COMPILE_VEC128` it checks whether `__int128` is available or not. But we are using `__m128i` here, which is available starting with SSE2.
So I'm wondering if the check is correct and the naming is precise (enough).

lib/c/libintvector.h
```diff
@@ -643,30 +656,34 @@ static inline Lib_IntVector_Intrinsics_vec128 Lib_IntVector_Intrinsics_vec128_lo
...
+
+#if defined(EVERCRYPT_CAN_COMPILE_VEC128)
 
 #include <vecintrin.h>
 
-// The main vector 128 type
-// We can't use uint8_t, uint32_t, uint64_t... instead of unsigned char,
-// unsigned int, unsigned long long: the compiler complains that the parameter
-// combination is invalid.
+/* The main vector 128 type
```
franziskuskiefer, line 668: Unrelated but because you changed the comment style here. IIRC `//` comments are only allowed starting C99, i.e. the C89 version shouldn't compile 🤔

franziskuskiefer (commented):
I generally think that this is the right approach. I'm just not sure whether naming and feature checks are on point.

lib/c/libintvector.h
msprotz: I was hoping to make this file compile in C89 mode since that's a property we're trying to preserve?

lib/c/libintvector.h
msprotz: That's my fault I forgot to refresh the snapshots. The diff for `dist/configure` should be correct (I fixed that earlier mistake by Son in one of my commits but forgot to propagate `dist/configure`-->`dist/*/configure`)

.ci/script.bat
msprotz: yeah I don't know how to do that but happy to start talking about it!

msprotz:
Thanks Franziskus! I'll push a refresh of the snapshots which should address your remark. I'm just not sure about the comment stuff... I thought C89-style comments were always the most compatible ones?

lib/c/libintvector.h
franziskuskiefer: Yes, it's a good chance. I was only wondering why no one complained about the c89 dist not actually being c89 compatible 🙂

msprotz: apparently you need -ansi to trigger the rejection of C++-style comments -- we don't use this flag. I was able to confirm that the command-line features -std=c89 -- I seem to recall some other clients of HACL downstream *do* use more pendantic modes, so perhaps it's a good thing to do to try to maintain C89-style comments in any case.

lib/c/libintvector.h
msprotz: everything should be refreshed... thanks for catching this!

franziskuskiefer (approved):
Thanks! Looks good to me now.

prosecco:
I think this is a good change. In HACLxN we were already hitting issues when compiling the code on ARM32 and AVX512 and needed to have special hand-written Makefiles for these cases. We are not handling multiple platforms well right now. Having a uniform configure script that runs first and optimizes for all platforms would be great. It would also be good for this script to generate useful warnings when it detects "non-optimal" configurations.

Ben, for dist/mozilla, how about we generate and check in a config.h file with the settings as they are currently, so you will see no change in behavior. Does that work? We should also change the flags, like Jonathan suggest, to something that is not EVERCRYPT related.

msprotz:
I've implemented the requested changes, namely:
- that the macros do not use the name "EVERCRYPT_" but instead start with HACL_
- provide dist/mozilla/config.h with the following contents:

```
#define HACL_CAN_COMPILE_INTRINSICS 1
#define HACL_CAN_COMPILE_VALE 1
#define HACL_CAN_COMPILE_INLINE_ASM 1
#define HACL_CAN_COMPILE_VEC128 1
#define HACL_CAN_COMPILE_VEC256 1
```

@beurdouche, what do you think of this latest batch of changes?

msprotz:
I also added the following behavior: for `code/*` directories that use the local Makefile, CFLAGS contains all of those flags as "enabled", meaning that a local build always assumes the "best and most recent" toolchain. This is unsafe, and in particular will cause the following problems:
- once AVX512 support lands, local Makefiles will fail with C compiler errors unless your toolchain is very recent and contains the AVX512 headers
- addcarry/subborrow intrinsics will always be unconditionally enabled, even though you may have a buggy GCC
- bzero support is assumed by default

All of these issues are or will be properly handled by the existing configure script, but my understanding is that this is the desired behavior for local development.

@polubelova please let me know if I have understood correctly, thanks

polubelova:
Thanks, @msprotz!
I don't mind fixing local Makefiles. It's just not easy to keep them up to date as they are not under CI and I'm not always aware of the recent changes :)
As for your comment about avx512, we can easily fix it by adding the _vec512.c/h files to BLACKLIST (in the local makefiles).

msprotz:
Perfect! Thanks for your quick response, much appreciated.

beurdouche:
Thanks Jonathan, from the top of my head, I think this is acceptable.

msprotz:
Thanks @beurdouche for your response, much appreciated! I'll merge as soon as I can, but for some reason Travis has been re-enabled and this is preventing the merge from happening. I may just disable it again unless someone knows why it has been re-instated.

msprotz:
Ok Travis is now disabled, @beurdouche can you mark your changes as accepted? This is blocking the merge for now.

msprotz:
Thanks everyone for your reviews and good feedback!

PR #422, opened 2021-03-12
bump rlimit on lemma that is failing with an F* update

nikswamy:
Cf slack discussion:

nik  11:20 AM
Has this lemma been on anyone's instability radar?
Vale.X64.Lemmas.lemma_eq_instr_write_outputs
Some innocuous change is causing it to fail unless I double the rlimit there.

chris.hawblitzel  11:38 AM
It looks like it's been slow but stable.  Doubling the rlmit is probably fine.


Chris-Hawblitzel (approved):

beurdouche (approved):

nikswamy:
thanks!

PR #424, opened 2021-03-19
Freebsd

web-vertalo:
These changes extend support to build the "all" target on FreeBSD 12.

They replace some Lilnux-isms (e.g. /bin/bash) with portable usage, add detection of amd64, increase fuel and rlimit for a failed proof, and work around an issue of exceeding command-line limits in "m obj/*".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,7 +184,7 @@ ifeq ($(shell uname -s),Darwin)
   ifeq (,$(shell which gtime))
     $(error gtime not found; try brew install gnu-time)
   endif
-  TIME := gtime
+  TIME := gtime -q -f '%E'
 else
   ifeq ($(shell uname -s),FreeBSD)
     SED := sed
@@ -193,7 +193,7 @@ else
   else
     SED := sed
     SEDi := -i
-    TIME := /usr/bin/time
+    TIME := /usr/bin/time -q -f '%E'
   endif
 endif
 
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -160,7 +160,7 @@ min-test:
 TO_CLEAN=$(foreach ext,checked checked.lax out cmd err time dump types.vaf krml cmx cmo cmi o d a,-or -iname '*.$(ext)')
 clean:
 	find . -iname '.depend*' $(TO_CLEAN) -delete
-	rm -rf obj/*
+	find obj -depth 1 -name .gitignore -o -delete
```
msprotz, line 163: is this correct? I feel like there's a `-not` missing

msprotz (commented):
@web-vertalo can you say what happens for Vale assembly on FreeBSD? I'd like to see the output of your configure script if that's ok -- is the inline assembly via gcc working out of the box?

Makefile
web-vertalo: Updated to avoid non-portable "-depth n", and change logic to use -not and implicit and instead of -o.

web-vertalo: Output from configure:
```
... detected x64 platform
...  supports compilation of 128-bit and 256-bit AVX/AVX2
...  supports vale
... not using known buggy Apple LLVM build
...  supports our inline ASM
... using a non-buggy GCC
 supports _addcarry_u64
... found ocamlfind in /usr/local/bin/ocamlfind
ocamlfind: Package `ctypes' not found
OCaml package ctypes not found
OCaml bindings disabled
```

web-vertalo: Does this PR need any more input from me?

msprotz:
Our CI system doesn't build PRs from external forks. I've pushed your work to branch protz_freebsd of the main repo where hopefully we'll get a full CI run. I'll merge this one once I get a green (fingers crossed)

PR #429, opened 2021-03-30
Bignum API

polubelova:
This PR introduces the following
- Bignum API: [32-bit](https://github.com/project-everest/hacl-star/blob/139fc5caf6b24f73f955c919754d3b266c73712d/dist/gcc-compatible/Hacl_Bignum32.h) and [64-bit](https://github.com/project-everest/hacl-star/blob/139fc5caf6b24f73f955c919754d3b266c73712d/dist/gcc-compatible/Hacl_Bignum64.h)
- Montgomery API (field arithmetic): [32-bit](https://github.com/project-everest/hacl-star/blob/139fc5caf6b24f73f955c919754d3b266c73712d/dist/gcc-compatible/Hacl_GenericField32.h) and [64-bit](https://github.com/project-everest/hacl-star/blob/139fc5caf6b24f73f955c919754d3b266c73712d/dist/gcc-compatible/Hacl_GenericField64.h)
- Almost Montgomery Multiplication
- Double exponentiation


code/bignum/Hacl.Bignum.SafeAPI.fst
```diff
@@ -13,19 +13,20 @@ module B = LowStar.Buffer
 module HS = FStar.HyperStack
 module LSeq = Lib.Sequence
 
+module E = Hacl.Spec.Exponentiation.Lemmas
```
beurdouche, line 16: `SEL`?

code/bignum/Makefile
```diff
@@ -7,15 +7,16 @@ all: dist/bignum-test.exe
 # Defines rules for producing .checked, .krml, .depend, etc.
 include ../../Makefile.local
 
-CFLAGS += -I ../../../lib/c -I ../../lib/c -march=native -mtune=native -O3
+CFLAGS += -I ../../../lib/c -I ../../lib/c -march=native -mtune=native -O3 -DCOMPILE_INTRINSICS
```
beurdouche, line 10: Is this really needed? If yes, does someone need to update all the build systems for all products currently using HACL* ?

beurdouche (approved):
I went through this pretty thoroughly and it looks very nice.
I generally like the naming conventions much better. There are still places in the internal functions where suffixes like `_f_` or `_r_` could be improved a bit, but overall this is great work !! : )

code/bignum/Makefile
polubelova: As long as they run the configure script, no update is required.
We don't run ./configure for local makefiles, that's why this change is needed.
[This](https://github.com/project-everest/hacl-star/commit/57c68278a4574038956f5b64b53ca8dd42abbd50) commit has changed the default behavior.

beurdouche: Nobody runs the configure file.

beurdouche: So the question is: what will break if this flag is not passed?

polubelova: It will take a fallback implementation or complain that the Hacl_IntTypes_Intrinsics.c file is not found.

prosecco (approved):

PR #432, opened 2021-04-14
bignum api

polubelova:
This PR makes the bignum API more defensive by removing `bBits > 0` from preconditions.


beurdouche (approved):

msprotz (approved):
Thanks Marina. Congratulations on keeping the comments and the code in sync!

PR #433, opened 2021-04-21
Solver tweaks in support of an F* change

nikswamy:
A couple of tweaks to proofs that may be a bit fragile


beurdouche (approved):

polubelova (approved):
Thanks!

PR #434, opened 2021-04-28
JS bindings: functions don't return promises

victor-dumitrescu:
This MR changes the JS API so that it returns the promise of an object containing functions which return immediately, instead of functions which return promises


msprotz (approved):
looks good to me, thanks Victor

denismerigoux (approved):

Issue #436, opened 2021-05-03
inlined bignum functions


franziskuskiefer:
Similarly to #418 and #419 the build breaks on Windows because of removed inline functions.
@polubelova removing inline from all of the bignum functions is probably the easiest approach for now or shall we only remove it from the failing ones?

```
2021-04-28T07:58:12.6980948Z [0m
2021-04-28T07:58:12.6981971Z [0m          libevercrypt.lib(Hacl_RSAPSS.obj) : error LNK2019: unresolved external symbol Hacl_Bignum_Convert_bn_from_bytes_be_uint64 referenced in function Hacl_RSAPSS_rsapss_sign
2021-04-28T07:58:12.6983064Z [0m
2021-04-28T07:58:12.6984030Z [0m          libevercrypt.lib(Hacl_FFDHE.obj) : error LNK2001: unresolved external symbol Hacl_Bignum_Convert_bn_from_bytes_be_uint64
2021-04-28T07:58:12.6984827Z [0m
2021-04-28T07:58:12.6985786Z [0m          libevercrypt.lib(Hacl_RSAPSS.obj) : error LNK2019: unresolved external symbol Hacl_Bignum_Convert_bn_to_bytes_be_uint64 referenced in function Hacl_RSAPSS_rsapss_sign
2021-04-28T07:58:12.6986689Z [0m
2021-04-28T07:58:12.6987480Z [0m          libevercrypt.lib(Hacl_FFDHE.obj) : error LNK2001: unresolved external symbol Hacl_Bignum_Convert_bn_to_bytes_be_uint64
2021-04-28T07:58:12.6988209Z [0m
2021-04-28T07:58:12.6989279Z [0m          libevercrypt.lib(Hacl_GenericField64.obj) : error LNK2019: unresolved external symbol Hacl_Bignum_Lib_bn_get_top_index_u64 referenced in function Hacl_GenericField64_field_init
2021-04-28T07:58:12.6990377Z [0m
2021-04-28T07:58:12.6991106Z [0m          libevercrypt.lib(Hacl_Bignum64.obj) : error LNK2001: unresolved external symbol Hacl_Bignum_Lib_bn_get_top_index_u64
2021-04-28T07:58:12.6991858Z [0m
2021-04-28T07:58:12.6992598Z [0m          libevercrypt.lib(Hacl_Bignum4096.obj) : error LNK2001: unresolved external symbol Hacl_Bignum_Lib_bn_get_top_index_u64
[...]

PR #437, opened 2021-05-03
drop some inlines in the bignum library to fix Windows build

franziskuskiefer:
This should be the minimum to fix the Windows build (also see #436 ).


polubelova (approved):
Thanks a lot!

msprotz (approved):

PR #440, opened 2021-05-10
Make sure dist/mozilla/{Makefile.config,config.h} are build productions

msprotz:
This is because we rely on the property that everything that matches `dist/*/*` is auto-generated... this is leveraged by the nightly CI to remove stale files from the dist/ directories.


dzomo (approved):

Kachoc (approved):

msprotz:
Very honored that the bot @dzomo deigned to review my PR

PR #441, opened 2021-05-12
aarch64 and iOS targets

franziskuskiefer:
This PR adds `aarch64-apple-darwin`, `aarch64-apple-ios`, and `x86_64-apple-ios-simulator` targets to build on arm based macOS, for iOS, and for the x64 iOS simulator.

### DIFF
--- a/dist/configure
+++ b/dist/configure
@@ -199,6 +199,7 @@ show_help () {
   printf "                    * aarch64-none-linux-gnu\n"
   printf "                    * aarch64-apple-darwin\n"
   printf "                    * aarch64-apple-ios\n"
+  printf "                    * x86_64-apple-ios-simulator\n"
   printf "    --disable-bzero Do not use explicit_bzero (binary will work with an old GLIBC)\n"
   printf "    --disable-ocaml Disable OCAML bindings\n"
   printf "\n"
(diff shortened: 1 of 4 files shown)


dist/configure
```diff
@@ -44,6 +44,10 @@ detect_ocaml () {
 }
 
 check_no_bug81300 () {
+  if [[ "$cross_build" == "1" ]]; then
+    # We don't want to build anything with march=native
+    return 0
+  fi
```
msprotz, line 50: I guess we never cross-compile for x64 so that's ok?

msprotz (approved):
That looks solid to me. I particularly appreciate putting those under CI. Thanks!

msprotz:
The CI failure was due to our build machine running out of disk space so I merged master in to trigger a new build.

tahina-pro (approved):

dist/configure
franziskuskiefer: Even if we'd be cross-compiling for x64 (like on an ARM-based Mac), we wouldn't want to use this because it uses `-march=native`. I guess no one is doing this (yet). But maybe we should improve this check to work for cross-compilation.

franziskuskiefer:
Hm, looks like the CI is failing again. But I don't really see an error 🤔

PR #442, opened 2021-05-13
Changes in support of an F* fix removing eta equivalence

nikswamy:
This PR is in anticipation of a fix for an unsoundness landing in F* master soon. @aseemr  and I have been working on it.

F* no longer supports eta equivalence, since it is incompatible with functional extensionality and subtyping. 
On the plus side, F* has become more extensional: functional extensionality is actually provable now, rather than being an axiom. 

This means however, that proofs that were relying on eta equivalence need to be revised. The main change in this PR is to a proof in LoopCombinators that was relying on eta and propositional extensionality in a rather tricky way. The code comment there has an explanation.

I expect the change to F* to land later today or tomorrow. If you approve, I can push the merge button at the right time to synchronize with the F* change.

Thanks!

### DIFF
--- a/code/bignum/Hacl.Impl.Exponentiation.fst
+++ b/code/bignum/Hacl.Impl.Exponentiation.fst
@@ -20,6 +20,7 @@ module BN = Hacl.Bignum
 module SN = Hacl.Spec.Bignum
 module BB = Hacl.Bignum.Base
 module SB = Hacl.Spec.Bignum.Base
+module PT = Hacl.Spec.PrecompTable
 
 #reset-options "--z3rlimit 50 --fuel 0 --ifuel 0"
 
@@ -398,20 +399,16 @@ val table_select_consttime_f:
   (requires fun h ->
     live h table /\ live h acc /\ disjoint table acc)
   (ensures  fun h0 _ h1 -> modifies (loc acc) h0 h1 /\
-    (Math.Lemmas.lemma_mult_le_right (v len) (v j + 2) (v table_len);
-     let res_j = gsub table ((j +! 1ul) *! len) len in
-     as_seq h1 acc == (if v i = v j + 1 then as_seq h0 res_j else as_seq h0 acc)))
+    as_seq h1 acc ==
+    PT.table_select_f (v len) (v table_len) (as_seq h0 table) i (v j) (as_seq h0 acc))
 
 let table_select_consttime_f #a_t len table_len table i j acc =
-  let h0 = ST.get () in
   let c = eq_mask i (BB.size_to_limb (j +! 1ul)) in
-  assert (v c == (if v i = v j + 1 then ones_v a_t else 0));
 
   Math.Lemmas.lemma_mult_le_right (v len) (v j + 2) (v table_len);
   assert (v ((j +! 1ul) *! len) == (v j + 1) * v len);
   let res_j = sub table ((j +! 1ul) *! len) len in
-  map2T len acc (BB.mask_select c) res_j acc;
-  SB.lseq_mask_select_lemma (as_seq h0 res_j) (as_seq h0 acc) c
+  map2T len acc (BB.mask_select c) res_j acc
 
 
 inline_for_extraction noextract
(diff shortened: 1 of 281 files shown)


polubelova (approved):
Thanks!

msprotz (approved):
Thanks for this very technical proof...

A couple questions:
- @aseemr you changed the precondition of C.Loops.while ... is this related? does this mean that preconditions should always be eta-expanded from here on? any guidelines we can communicate to HACL* developers would be helpful
- a lot of the proof steps are done "by tactic"... is this cosmetic, a personal preference, or is there some deep reason for these proofs to be performed with e.g. `l_to_r`?

Thanks!

Jonathan

msprotz:
FYI this build failed because of:

```
[STDERR]/home/everest/hacl-star/lib/Lib.LoopCombinators.fst(102,4-102,10): (Error 228) user tactic failed: `t_trefl: cannot unify (i: a{p i}) and (i: a{q i})` (see also /home/everest/hacl-star/lib/Lib.LoopCombinators.fst(103,45-103,54))
[STDERR]1 error was reported (see above)
[STDERR]make[1]: *** [Makefile:405: obj/Lib.LoopCombinators.fst.checked] Error 1
```

but I assume this will start working once the F* change lands?

aseemr:
Hi Jonathan, yes that eta expansion is related. The way the precondition is written in C.Loops is `fun h -> inv h` and here it was written `h`. To prove these two equal, we earlier had eta equality but not anymore. So I had to expand it here too. I would say no special guidelines as such ... most of the times the preconditions are more than simple `inv`.

The failure is because the proof requires some changes that Nik did in F* `nik_remove_eta`. So we discussed that we will first merge F*, get a green here, and then merge it.

msprotz:
Thanks... so if the function requires `test: ... Stack (fun h -> inv h)` and you pass it an argument of type `.. -> Stack inv`, it's now a typing error, correct?

Will it always be a typing error from here on, or just in certain cases? (If it's the former, it's a surprising enough change that I should send a heads-up to other HACL folks who may not be following such F* changes closely.)

nikswamy:
That particular proof is quite technical. It does require a tactic to rewrite with propositional extensionality, then refinement-type equality, at depth. SMT-based proofs cannot do that. This kind of thing shouldn't come up very often.

aseemr:
It also has to do with the `$` on the argument of `C.Loops.while` ... if that `$` was not there subtyping would kick in and this would work. But with `$`, we try to prove equal which is not possible now in the absence of eta.

msprotz:
Ah I see. Barely anyone uses `$` (@tahina-pro was the one who introduced the `$` here if I recall correctly) so I don't expect this will be a problem for day-to-day uses of HACL*. Thanks for the clarification!

tahina-pro:
I was using `$` to empirically help in second-order unification of the pre and postconditions of the test routine, but I certainly do not claim any expertise on that.

nikswamy:
Weird, I'm seeing failures now in Hacl.Impl.Exponentiation that I wasn't seeing before. Will take a look later this evening.

code/bignum/Hacl.Impl.Exponentiation.fst
```diff
@@ -443,20 +438,19 @@ let table_select_consttime #a_t len table_len table i res =
...
-      //(if v i = v j + 1 then LSeq.sub (as_seq h0 table) (v i * v len) (v len) else as_seq h2 res));
-    assert (inv h3 (v j + 1))
-    )
-//////////////////////////////////////////////////////////
+    Loops.unfold_repeati (v j + 1)
+      (PT.table_select_f (v len) (v table_len) (as_seq h0 table) i) (as_seq h1 res) (v j);
+    table_select_consttime_f len table_len table i j res);
+  PT.table_select_lemma (v len) (v table_len) (as_seq h0 table) i
 
+//////////////////////////////////////////////////////////
```
nikswamy, line 453: Thanks Marina! I would never have figured this out myself. Much appreciated!

PR #444, opened 2021-05-20
Proof regressions for F* changes related to #2288

aseemr:
Proof regressions for F* PR https://github.com/FStarLang/FStar/pull/2288.

Nothing major here. The hacl proofs were just brittle, I checked that the emitted query did not change, there were a few more defns. in the SMT context.

Vale code needs changes because it uses the `PURE` effect directly, so we need to prove the monotonicity of its wps.


polubelova (approved):
Thanks!

beurdouche (approved):

aseemr:
Thanks @polubelova , @beurdouche !

PR #447, opened 2021-05-28
streaming api for ed25519

### LINKED ISSUE #446: Ed25519 and HKDF stack overflow
Ed25519 can't be used on messages of 1MB or larger on Windows and 8MB on macOS/Linux.
The code tries to allocate memory that fits the message on the stack, which doesn't work messages of these sizes.

https://github.com/project-everest/hacl-star/blob/db532a99428c09cc4c474746bc46381fe78c59a0/dist/gcc-compatible/Hacl_Ed25519.c#L1221-L1224

https://github.com/project-everest/hacl-star/blob/db532a99428c09cc4c474746bc46381fe78c59a0/dist/msvc-compatible/Hacl_Ed25519.c#L1221-L1224

For some reason `KRML_CHECK_SIZE` doesn't fail as expected but the memory allocation actually fails with an overflow, which appears to be a bug in kremlin.

> thread 'main' has overflowed its stack

> (signal: 11, SIGSEGV: invalid memory reference

@polubelova how about using a streaming API for the hashing instead? That's probably better than putting the message on the heap.
@protz `KRML_CHECK_SIZE` appears to be for heap allocations rather than stack but is used to guard stack allocations in hacl. We either need another `CHECK_SIZE` or change this to check for stack allocation limits.

polubelova:
This PR fixes #446 for ed25519.


msprotz (approved):
This is great. Thanks for adding the bits in the README!

franziskuskiefer:
Nice, thanks! 👍🏻

karthikbhargavan:
Looks good. Should we do something similar for ecdsa, rsa etc.?

karthikbhargavan (approved):

PR #449, opened 2021-05-31
Add utilities to Lib.Buffer

Kachoc:
This small PR adds several utilities to Lib.Buffer, especially to manipulate (potentially) null buffers and dynamically allocated buffers.


Kachoc:
Tagging @protz for information

code/bignum/Hacl.Bignum.MontArithmetic.fsti
```diff
@@ -21,6 +21,8 @@ module BM = Hacl.Bignum.Montgomery
 
 #set-options "--z3rlimit 50 --fuel 0 --ifuel 0"
 
+val _align_fsti : unit
```
beurdouche, line 24: What is that for?

code/chacha20poly1305/Hacl.Impl.Chacha20Poly1305.fst
```diff
@@ -201,12 +201,12 @@ val aead_decrypt: #w:field_spec -> aead_decrypt_st w
 [@ Meta.Attribute.specialize ]
 let aead_decrypt #w k n aadlen aad mlen m cipher mac =
   push_frame();
-  let h0 = get() in
+  let h0 = ST.get() in
```
beurdouche, line 204: I would much prefer `HS` than `ST` here. But I agree that having the prefix is better.

lib/Lib.Buffer.fsti
```diff
@@ -79,21 +79,59 @@ let const_to_lbuffer #a #len (b:clbuffer a len{CB.qual_of (b <: cbuffer a) == CB
 let const_to_ilbuffer #a #len (b:glbuffer a len)  : r:ilbuffer a len =
   const_to_ibuffer #a b
 
-unfold let null (a: Type0) : lbuffer a (size 0) = B.null #a
+unfold let null (ty : buftype) (a : Type0) : buffer_t ty a =
```
beurdouche, line 82: `ty` should be an implicit.

lib/Lib.Buffer.fsti
```diff
@@ -79,21 +79,59 @@ let const_to_lbuffer #a #len (b:clbuffer a len{CB.qual_of (b <: cbuffer a) == CB
...
+  | CONST -> CB.null a
+
+let g_is_null (#ty : buftype) (#a : Type0) (b : buffer_t ty a) : GTot bool =
+  match ty with
+  | IMMUT -> IB.g_is_null (b <: ibuffer a)
+  | MUT -> B.g_is_null (b <: buffer a)
+  | CONST -> B.g_is_null (CB.as_mbuf (b <: cbuffer a))
 
-//val live: #t:buftype -> #a:Type0 -> h:HS.mem -> b:buffer_t t a -> Type
 let live (#t:buftype) (#a:Type0) (h:HS.mem) (b:buffer_t t a) : Type =
```
beurdouche, line 94: Change this from `t` to `ty` everywhere if you are gonna use this

code/bignum/Hacl.Bignum.MontArithmetic.fsti
Kachoc: There are some issues compiling this file in command-line mode otherwise. To be more precise, it seems that in command-line, the `‘freeable` definition introduced in the .fst [here](https://github.com/project-everest/hacl-star/blob/1ecddaafd4cde2d01677599a1f843619fc4933b1/code/bignum/Hacl.Bignum.MontArithmetic.fsti#L90) doesn't shadow the `Lib.Buffer.freeable` definition soon enough. Introducing this kind of "aligners" solves the issue.

code/chacha20poly1305/Hacl.Impl.Chacha20Poly1305.fst
Kachoc: I can do that but I used `ST` because many people write `ST.get` (some files had a mix of `ST.get` and `get` actually), and  the `module ST = FStar.HyperStack.ST` abbreviation is often present in the files.

lib/Lib.Buffer.fsti
Kachoc: Will update that

code/chacha20poly1305/Hacl.Impl.Chacha20Poly1305.fst
Kachoc: Also, the reason I added this prefix is that I introduced a `Lib.Buffer.get` definition, which shadows those occurrences of `get`

lib/Lib.Buffer.fsti
Kachoc: I actually changed all the `ty` to `t` for the buffer type

lib/Lib.Buffer.fsti
Kachoc: Updated

code/chacha20poly1305/Hacl.Impl.Chacha20Poly1305.fst
beurdouche: Actually, ideally we want to use `Lib.Buffer.get`

beurdouche (approved):

PR #450, opened 2021-06-01
OCaml API revamp and documentation

victor-dumitrescu:
This PR is a major change to the OCaml API. The main points are:
- A full `odoc` based documentation for the entire API, which can be built using `dune build @doc`. The main files containing docstrings are `EverCrypt.mli`, `Hacl.mli`, and `SharedDefs.ml`.
- Switching the default functions from an interface in which the output buffer is passed as an argument to the more idiomatic way of returning the output buffer. This also simplifies the code on the client side because users no longer need to initialise buffers themselves every time. The older functions are still included, either in `Noalloc` submodules or named `*_noalloc` in modules which only contain one such function.
- All arguments now have labels.
- Other changes such as improvements to tests, small fixes, additional runtime checks where they were missing or inconsistent, etc.


msprotz:
Per the HACL* repository rules we need two reviewers, preferably one from INRIA and one from MSR. Sorry Denis, you've been volunteered.

msprotz (changes requested):
This generally looks good to me.
- I like the `NoAlloc` submodule style. Can this be made regular? Right now sometimes you have a `NoAlloc` submodule and sometimes a `_noalloc` suffix on your functions.
- Can you try to work out automatic doc generation as part of this PR? I would absolutely love to make sure we keep this beautiful documentation online and automatically refreshed
- Why is the blake2 test gone? Is it now tested as part of the agile hash API?

The documentation is REALLY good. Much better than what we have for C.

bindings/ocaml/EverCrypt.ml
```diff
@@ -85,30 +65,46 @@ module AEAD = struct
...
-    assert (C.disjoint tag ct);
-    assert (C.disjoint tag ad);
-    assert (C.disjoint ct ad);
-    assert (C.disjoint dt ad);
-    get_result (everCrypt_AEAD_decrypt (!@st)
-                  (C.ctypes_buf iv) (C.size_uint32 iv) (C.ctypes_buf ad) (C.size_uint32 ad)
-                  (C.ctypes_buf ct) (C.size_uint32 ct) (C.ctypes_buf tag) (C.ctypes_buf dt))
+
+  module Noalloc = struct
+    let encrypt ~st:(alg, st) ~iv ~ad ~pt ~ct ~tag : unit result =
```
denismerigoux, line 70: Don't know what's the convention here or at Nomadic Labs, but my personal discipline is to annotate the argument of all functions with their types to improve code readability :)

bindings/ocaml/SharedDefs.ml
```diff
@@ -4,32 +4,39 @@ open Unsigned
...
   let equal = Bytes.equal
+  let make l = Bytes.make l '\x00'
   let disjoint b1 b2 = b1 != b2
   let sub = Bytes.sub
   let z_compare b z = Z.compare (Z.of_bits (Bytes.to_string b)) z
 end
+(** Representation of [Bytes.t] buffers *)
 
-(* VD: temporarely disable, eliminate dependency on bigstring *)
+(* VD: temporarily disable, eliminate dependency on bigstring *)
```
denismerigoux, line 34: Can we remove that commented code now ?

bindings/ocaml/SharedDefs.ml
```diff
@@ -41,7 +48,46 @@ end
 module Hacl_Hash = Hacl_Hash_bindings.Bindings(Hacl_Hash_stubs)
 module Hacl_Spec = Hacl_Spec_bindings.Bindings(Hacl_Spec_stubs)
 
-let max_size_t = Z.of_string "4294967296"
+let pow2 n = Z.(pow ~$2) n
```
denismerigoux, line 51: Never seen that `~$2` syntax, what does it mean?

bindings/ocaml/tests/aead_test.ml
```diff
@@ -103,10 +156,20 @@ let _ =
...
   Printf.printf "RDRAND: %b\n" (has_feature RDRAND);
 
+  test_agile_noalloc chacha20poly1305_test;
   test_agile chacha20poly1305_test;
-  test_nonagile chacha20poly1305_test "Hacl.Chacha20_Poly1305_32" Hacl.Chacha20_Poly1305_32.encrypt Hacl.Chacha20_Poly1305_32.decrypt [];
-  test_nonagile chacha20poly1305_test "Hacl.Chacha20_Poly1305_128" Hacl.Chacha20_Poly1305_128.encrypt Hacl.Chacha20_Poly1305_128.decrypt [VEC128];
-  test_nonagile chacha20poly1305_test "Hacl.Chacha20_Poly1305_256" Hacl.Chacha20_Poly1305_256.encrypt Hacl.Chacha20_Poly1305_256.decrypt [VEC256];
-  test_nonagile chacha20poly1305_test "EverCrypt.Chacha20_Poly1305_256" EverCrypt.Chacha20_Poly1305.encrypt EverCrypt.Chacha20_Poly1305.decrypt [];
 
+  let module Tests = MakeTests (EverCrypt.Chacha20_Poly1305) in
```
denismerigoux, line 162: Using first-class modules here, maybe this breaks backwards-compatibility with really old versions of OCaml ?

bindings/ocaml/tests/dune
```diff
@@ -8,7 +8,6 @@
         hkdf_test
         nacl_test
         drbg_test
-        blake2_test
```
denismerigoux, line 11: Same question as Jonathan, why removing the test?

bindings/ocaml/tests/hash_test.ml
```diff
@@ -12,165 +13,294 @@ type alg =
...
 let test_sha3_512 : Bytes.t hash_test =
   {
     name = "SHA3_512 Test 1";
     alg = SHA3_512;
-    plaintext = Bytes.of_string "\x61\x62\x63";
+    msg = Bytes.of_string "\x61\x62\x63";
     expected = Bytes.of_string "\xb7\x51\x85\x0b\x1a\x57\x16\x8a\x56\x93\xcd\x92\x4b\x6b\x09\x6e\x08\xf6\x21\x82\x74\x44\xf7\x0d\x88\x4f\x5d\x02\x40\xd2\x71\x2e\x10\xe1\x16\xe9\x19\x2a\xf3\xc9\x1a\x7e\xc5\x76\x47\xe3\x93\x40\x57\x34\x0b\x4c\xf4\x08\xd5\xa5\x65\x92\xf8\x27\x4e\xec\x53\xf0"
 }
 
+let test_blake2b : Bytes.t hash_test =
```
denismerigoux, line 89: Oh OK you've added it back here, never mind.

denismerigoux (approved):
Hi Victor, hi Jonathan,

It's the first time I take a look at this code so I'm not aware of all the details but this seems like a great improvement, especially documentation-wise. The labeled argument style for the algorithms is also very good since all arguments have basically the same type. 

This is a lot of code and I wonder if in a future big refactoring we could take advantage and improve of the JSON description of the function's signatures in JS and auto-generate this OCaml code bindings. As far as I could see, all the function's preconditions are asserted but there is an implicit link with the original F* that still has to be checked manually if it changes.

bindings/ocaml/EverCrypt.ml
victor-dumitrescu: Fair point, but since I've annotated the functions in the .mli (in this case `val encrypt : st:t -> iv:bytes -> ad:bytes -> pt:bytes -> ct:bytes -> tag:bytes -> unit Error.result` for example) and by extension in the docs too I thought it's maybe not worth doing it again here?

bindings/ocaml/SharedDefs.ml
victor-dumitrescu: I guess we could. I wanted to keep compatibility with Bigstring in case we at some point decide to use that instead of / in addition to Bytes and I kept this commented code around to give some context to the extra complexity that results from this in the code. Thoughts @msprotz ?

bindings/ocaml/SharedDefs.ml
victor-dumitrescu: It's a Zarith prefix operator for converting from int.

bindings/ocaml/tests/aead_test.ml
victor-dumitrescu: Ah didn't think of that. Seems they were introduced in OCaml 3.12, so I don't think we care too much? (We certainly don't at Nomadic). In any case, they've been in different parts of the codebase from the beginning, like in `SharedFunctors.ml`.

bindings/ocaml/SharedDefs.ml
denismerigoux: Oh you taught me something here, from now on I'll stop struggling with `Z.of_int` :)

bindings/ocaml/EverCrypt.ml
denismerigoux: Yeah you're probably right. And the types are all the same anyway here.

bindings/ocaml/tests/aead_test.ml
denismerigoux: Wouahou I didn't think they were so old, OK I don't think we need to maintain compat below 3.12.

victor-dumitrescu:
Thanks Denis and Jonathan for the reviews! Answering to your more general points:

> Right now sometimes you have a NoAlloc submodule and sometimes a _noalloc suffix on your functions.

Indeed, for modules in which only one function has a "noalloc" version I thought it would be a bit overkill to have a submodule just for it. But perhaps it would be better for consistency?

> Can you try to work out automatic doc generation as part of this PR? 

Yes, I'll try to make that work too.

> Why is the blake2 test gone? Is it now tested as part of the agile hash API?

Yes, I've moved the previous test in `test_hash.ml` and I'm also testing the agile API.

> This is a lot of code and I wonder if in a future big refactoring we could take advantage and improve of the JSON description of the function's signatures in JS and auto-generate this OCaml code bindings. As far as I could see, all the function's preconditions are asserted but there is an implicit link with the original F* that still has to be checked manually if it changes.

Yeah that's something we discussed in the past. Currently we only extract the low-level bindings using KreMLin and the rest is hand-written. It would be great if we could automatically get the preconditions too, leaving only the organisation of the API to be manually written.

bindings/ocaml/EverCrypt.ml
msprotz: I think it's ok because of the mli. If not for the mli, I would have advocated for annotated types.

bindings/ocaml/SharedDefs.ml
msprotz: No strong opinion. I thought we ruled out Bigstring a while back? I'm generally against keeping dead code in comments, since there's always version control. You could leave a comment that says "for a version of this built on `BigString`, see 4c1ee47"

bindings/ocaml/tests/aead_test.ml
msprotz: Clearly, first-class modules are old enough and this is definitely a perfect use-case for them. Does that make you an old-timer Denis?

msprotz:
> Indeed, for modules in which only one function has a "noalloc" version I thought it would be a bit overkill to have a submodule just for it. But perhaps it would be better for consistency?

Please, I think consistency is super important, and for all you know, some of those modules may gain another function in the future, and it'd be good to avoid breaking compatibility.

> Yeah that's something we discussed in the past. Currently we only extract the low-level bindings using KreMLin and the rest is hand-written. It would be great if we could automatically get the preconditions too, leaving only the organisation of the API to be manually written.

The easiest way would probably be to write an F\* meta-program that inspects a series of definitions and dumps JSON metadata to be reused by the bindings. None of the types are carried through extraction so KReMLin really can't help with that unless we undertake a massive refactoring of the whole pipeline.

victor-dumitrescu:
Worth noting that @antoniolocascio, currently an intern at Nomadic, has been working on F* metaprograms for extracting preconditions, which could eventually be used for doing this kind of thing.

victor-dumitrescu:
I've added the automatic generation of docs [here](https://github.com/project-everest/hacl-star/pull/450/commits/dfe5f5f80167ef36a05f02d33000f2cd4b266eb7) following the JS model. I'm not sure however if this will work in CI or not.

victor-dumitrescu:
I've added doc generation [here](https://github.com/project-everest/hacl-star/pull/450/commits/603a05a60859f0fbb1607ac4a6ec25d2e94ac6cf) following the JS example. I think the only thing missing is that `odoc` needs to be installed on the CI machine.

msprotz:
```
[STDERR]Error: Program odoc not found in the tree or in PATH
[STDERR] (context: default)
[STDERR]Hint: opam install odoc
[STDERR]make[1]: *** [Makefile:841: doc-ocaml] Error 1
```

@tahina-pro any chance you could give Victor some pointers on how to make sure odoc is part of the build? thanks!

tahina-pro:
Does this mean that we need to add `odoc` to https://github.com/project-everest/everest/blob/master/opam-packages ?

msprotz:
@victor-dumitrescu can you submit a PR (or push directly) to add odoc there, and then merge master into this branch to trigger a rebuild? thanks!

tahina-pro:
Pushing on Everest will not be enough, since I need to rebuild the underlying CI infrastructure image. Let me take care of those.

msprotz:
Thanks Tahina! Much appreciated

msprotz (approved):
This is really great! Very eager to hear about your work on extracting preconditions. Would be nice to present it at an F* meeting or something. Cheers.

victor-dumitrescu:
Thanks Tahina! I think this PR is good to go then, right?

msprotz:
Looks good to me. Thanks again Tahina!

PR #452, opened 2021-06-11
Bump JS bindings npm package version

denismerigoux:
I also published the corresponding new version of the package here: https://www.npmjs.com/package/hacl-wasm


msprotz (approved):
Is it because you refreshed the package on npm that now contains more algorithms and you want to reflect that with a new version number?

denismerigoux:
Yes it is exactly for this reason

victor-dumitrescu (approved):

PR #455, opened 2021-06-16
Explicitly declare Power9 option in dist/configure

mamonet:
Tuning build configuration for native architecture isn't recommended because if the build machine has newer CPU than the target machine the output binaries will crash with illegal instructions signal when run on older CPU. This patch sets AltiVec and PPC64 as the minimum targeted architecture and set Power9 instruction set as optional. Explicitly using -mpower9-vector flag has performance increase 6% over native CPU tuning on POWER9 processor.

### DIFF
--- a/dist/configure
+++ b/dist/configure
@@ -207,6 +207,7 @@ show_help () {
   printf "                    * x86_64-apple-ios-simulator\n"
   printf "    --disable-bzero Do not use explicit_bzero (binary will work with an old GLIBC)\n"
   printf "    --disable-ocaml Disable OCAML bindings\n"
+  printf "    --enable-power9 Enable Power ISA v3.0 instruction set for PowerPC architecture\n"
   printf "\n"
 }
 
@@ -218,6 +219,7 @@ echo -n > config.h
 # Default arguments
 disable_ocaml=0
 disable_bzero=0
+enable_power9=0
 cross_build=0
 
 # Parse command line arguments.
@@ -227,6 +229,7 @@ while [ $# -gt 0 ]; do
         -target) build_target="$2"; shift ;;
         --disable-ocaml) disable_ocaml=1 ;;
         --disable-bzero) disable_bzero=1 ;;
+        --enable-power9) enable_power9=1 ;;
         --help) show_help; exit 0 ;;
         *) show_help; exit 2 ;;
     esac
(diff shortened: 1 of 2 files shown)


msprotz (commented):
Should this be a new possible value for the `-target` option? I would imagine that just like we're doing `./configure -target aarch64-apple-ios`, we might be able to do `./configure -target power9`?

mamonet:
> Should this be a new possible value for the `-target` option? I would imagine that just like we're doing `./configure -target aarch64-apple-ios`, we might be able to do `./configure -target power9`?

-target option specify the toolchain for build configuration, PPC has the same toolchain for all instruction set architectures for each endianness mode.

msprotz (approved):
Looks good to me, then!

franziskuskiefer:
> > Should this be a new possible value for the `-target` option? I would imagine that just like we're doing `./configure -target aarch64-apple-ios`, we might be able to do `./configure -target power9`?
> 
> -target option specify the toolchain for build configuration, PPC has the same toolchain for all instruction set architectures for each endianness mode.

right, `-target` is usually for cross-compilation. So this looks good to me.

msprotz:
Thanks for the review @franziskuskiefer !

edelsohn:
The CFLAGS options added are not the correct way to enable Power9 support.  One should not enable subsets of processor feature options, such as `-mpower9-vector`.  This can enable some features and omit other, related features.

I don't understand the reason to add `-mpowerpc64` and `-mabi=altivec`.

PPC64LE Linux defaults to Power8, with Power8 VSX enabled (including Altivec).  I'm surprised that any command line options are necessary by default.

For Power9, the `-mcpu=power9` command line option should be used, not individual feature options.

mamonet:
> The CFLAGS options added are not the correct way to enable Power9 support. One should not enable subsets of processor feature options, such as `-mpower9-vector`. This can enable some features and omit other, related features.

I'm aware that `-mpower9-vector` enables certain feature of Power9 subsets but in case of libintvector.h vectorization that feature is what matter, somehow it yields 6% performance increase for 128-bit vectorized Poly1305 MAC over enabling `-mcpu=power9` on POWER9 processors.

> I don't understand the reason to add `-mpowerpc64` and `-mabi=altivec`.
> 
> PPC64LE Linux defaults to Power8, with Power8 VSX enabled (including Altivec). I'm surprised that any command line options are necessary by default.

Right, 64-bit instruction set and altivec support are enabled by default in PPC64LE but what about the big-endian variant?
libintvector.h vectorizing could support both endianness modes and should work on old Power architectures just fine, I think it's ok to keep those flags in case some build the binares in big-endian mode.

edelsohn:
I believe that the options are being used incorrectly.

PPC64 Linux Big Endian is deprecated and end of life.  It doesn't matter.
HACL* on AIX (BE) may be of interest, but that is a separate issue.

PPC64LE Linux is only 64 bit.
`-mpowerpc64` enables 64 bit instructions, but not 64 bit mode.  It is meant for using 64 bit instructions on a 32 bit system, but the OS needs 64 bit support.  This either is unsafe or redundant.
`-m64` enables 64 bit mode, again redundant for PPC64LE.

`-mabi=altivec` enables the Altivec ABI, not Altivec instructions.  It sets some extensions about register usage and intrinsics.  It is enabled by default for PPC64LE Linux.  It is not the correct way to optionally enable Altivec or VSX.

Yes, `-mpower9-vector` enables the subset of Power9 VSX instructions that libintvector.h uses, but it still is not the preferred way to enable those instructions.

In summary, `-mpowerpc64` and `-mabi=altivec` are redundant, and Power9 features should be enabled with `-mcpu=power9`, not enabling the Power9 vector instructions in isolation.

PR #463, opened 2021-07-07
Build fixes for FreeBSD: link in proper ASM implementations

msprotz:
Building on FreeBSD currently results in undefined symbols. Here's why.

- The Makefile finds all .S files of the form *-$(MARCH)-$(VARIANT), assembles them and links them into the final library
- The SystemV ABI files with non-OSX symbol mangling convention are currently named *-linux-x86_64.S
- These files work on Linux and FreeBSD (because I believe the calling convention and symbol convention is the same)
- But on FreeBSD, MARCH=amd64 and VARIANT=freebsd

This patch fixes both MARCH and VARIANT on FreeBSD.


franziskuskiefer (approved):
lgtm

It would be great if the people who care about FreeBSD could set up a CI for it so we know at least when it breaks.

daachi:
@franziskuskiefer I will be working on building out CI for FreeBSD.

tahina-pro:
Should a new snapshot with updated distribution-specific Makefiles also be included in this PR?

tahina-pro:
> Should a new snapshot with updated distribution-specific Makefiles also be included in this PR?

I tried to build and test EverParse with this HACL* PR as is, but of course I failed, since `dist/gcc-compatible/Makefile` was not updated. I would have had to rebuild all of HACL* to make sure to update it properly, but I read the main Makefile and I figured out that copying `dist/Makefile.tmpl` could help. Doing so, I successfully managed to build and test EverParse on a FreeBSD 12.2 VM, following following https://project-everest.github.io/everparse/build.html#build-in-the-source-tree

tahina-pro (approved):

PR #467, opened 2021-07-20
Fixes to gcc-compatible configure script and opam files

victor-dumitrescu:
The main change in this PR is a fix to the `configure` script in cases where `mktemp` fails to create temporary files used to run various tests on the platform on which gcc-compatible / hacl-star-raw is being compiled. This is notably the case for the CI sandbox on opam-repository, which doesn't grant write permissions. In this case, the functions in `configure` which use these tests return true, in turn enabling the compilation of files potentially not available on the platform, which ends in error. The fix makes these functions return false in case of `mktemp` failures.

Other changes include:
* rearranging the opam files to better conform to the opam expected format, since I'm no longer able to use the formatting provided through `opam-publish` because of a bug in the tool
* setting a minimum version for ctypes
* bumping the minimum OCaml versions required from 4.05.0 to 4.08.0

The last change isn't crucial, it's just a convenience for using the Option module, so if anyone thinks we should still support these OCaml versions, this can be accommodated.


msprotz:
Unfortunately while erroring out will result in safe defaults, I'm concerned that the performance will suffer. For instance, the lack of  `uint128` support might induce a big performance hit. Can the temporary file be created in the same directory?

e.g. (sketch)

```
function my_mktemp() {
  local prefix=$1
  local f=$(mktemp /tmp/$prefix.XXXXXXXX)
  if [[ $? == 0 ]]; then
    echo $f
  else
    local i = 0
    while [[ -f test$i.c ]]; do
      i=$(($i+1))
    done
    echo test$i.c
  fi
```

msprotz (approved):
looks good nice and simple fix

Kachoc (approved):

victor-dumitrescu:
Also pulled in the changes from @daachi's #468.

PR #469, opened 2021-07-21
Reenable travis

daachi:
This will turn Travis CI back on for the repo, minus ARM builds which are in drone.io.

We will need to sort out credit usage in order to fully use Travis since builds are currently being rejected due to credits being exhausted.

A follow-on PR will add CI for FreeBSD as well, since we have hacl-star working there now.


daachi:
travis_run_including_freebsd_broken_windows

An inital travis run used up 830 credits. Note that this credit usage includes running CI for FreeBSD. Note also that the Windows build was broken for this run, so the credit usage does not reflect the case where Windows is fully enabled. Not sure if the Windows build is wanted but we'll need to fix paths in script.bat (at least) to get that working again.

In a perfect world, we could review the usage statistics for hacl-star and gain clarity on what's causing builds to be rejected. Reading the [Travis blog post](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) on this topic I'm led to believe that perhaps OSX builds are part of the problem, but need more input on usage to make a meaningful determination.

Usage statistics are available from the Travis UI, user profile -> Settings -> Plan Usage.

Probably need input from @protz and/or @tahina-pro to gather this data.

daachi:
results_of_failed_windows_build

We'll need to fix this to get Windows builds working, if that's wanted. I can figure this out unless someone knows immediately how to fix.

msprotz:
Thanks for all the work! What does this credit consumption mean? Will we run out of credits in a week, a month? Can these be replenished?

Maybe we should disable OSX and Windows since they're already covered by actions and save Travis for more "exotic" scenarios?

franziskuskiefer:
I'd suggest to use Travis only for platforms that aren't supported by GH actions such as FreeBSD to save credits.
Everything in here we could do in actions as well.

msprotz:
Agreed with that, Franziskus. Thanks for chiming in -- there's a bunch of other related PRs and I'd love to have your opinion on those if you haven't commented already.

daachi:
The way the consumption looks to me is that CI would only be able to run on Travis 9 times and then, assuming the 10k credit free plan, that would be all for the month. :joy: 

The credits can be replenished and the best way to determine a budget for the project would be to look at the usage statistics. 

One potential way to to this would be to add some conditionals to the Travis configuration to not have CI running for every single commit.

However, agreeing with Franziskus and assuming there would not be any funds to put towards this, if it would be better to do these jobs with actions, let's do that. I'm happy to build that out if no one is working on it already.

msprotz:
Is there a way to ask politely for a bigger monthly allowance of credits on the basis that we're an open-source project with a big-ish need for CI?

No one is currently actively working on this so please go ahead and feel free to start migrating some of the old travis jobs to GH Actions if they haven't been migrated already. Would be good to decide on a plan on Slack with Franziskus and I at least before you go ahead.

Thanks!

daachi:
I'm not sure about the first question, but will look into it.

I will follow-up with you and Franziskus on Slack about rest.

msprotz:
Any news on this? I'd love to have Travis back but until we sort out the issue of credits, I take it it's not worth it?

daachi:
Yes, unfortunately I don't think it's worth it. We'd only be able to have a
very few runs before the credits were tapped out again.

--Anthony

franziskuskiefer:
Is FreeBSD the main issue you want to address @daachi ? You could do this with GH actions using for example this FreeBSD-vm action https://github.com/marketplace/actions/freebsd-vm

daachi:
@franziskuskiefer yes, we just wanted to have the ability to know if there was a regression on FreeBSD.

franziskuskiefer:
> @franziskuskiefer what do we do with this? is there freebsd CI on hacl-packages?

We don't have CI for this. It's not a tier 2 platform.
If there's someone interested in doing this, it can be done with GH actions (e.g. https://github.com/vmactions/freebsd-vm). I'm happy to review a PR on hacl-packages for this or talk to whoever wants to move FreeBSD support to Tier 2 or higher.

msprotz:
Per the discussion with Franziskus, no immediate action needed on this. If someone wants to reinstate FreeBSD CI, let's do it on hacl-packages, with proper GitHub actions. Thanks.

PR #471, opened 2021-07-27
Adapt uses of sigelt_view to changes in FStar#2291

antoniolocascio:
[FStarLang/FStar#2291](https://github.com/FStarLang/FStar/pull/2291) adds support for mutually recursive let-bindings in `sigelt_view`. To do so, the interface of the`Sg_Let` constructor changes, breaking some uses in HACL*. This PR fixes those uses.


msprotz (approved):
This is really great -- much appreciated! Did you get an Everest green with those changes?

antoniolocascio:
I was able to build Everest, but I think my Everest setup is broken. Tests always fail, even without my changes.

msprotz:
If it's just tests I'm not too worried. Some of them are notoriously broken (i.e. your CPU may not have the right instruction and the test targets don't really account for that.)

msprotz:
As long as it verifies, this is good for me. You may want to ask @R1kM or @Kachoc for an extra review since they know this code

Kachoc (approved):
It looks good to me too. And I have the same position as @msprotz : as long as it verifies, it is good to me.

msprotz:
Merging since CI won't run for external PRs in HACL (even after clicking the "approve" button), and @aseemr checked and got an Everest green.

Thanks again @antoniolocascio !!

PR #474, opened 2021-08-10
Add AESGCM encrypt/decrypt functions with no dynamic checks

Kachoc:
There was previously no easy way of accessing the AESGCM functions outside the agile EverCrypt versions. Due to modularity issues, using those functions also forced people to work in the ST effect, and not the Stack effect. This PR fixes that.


Kachoc:
@msprotz

msprotz (approved):
perfect!

polubelova (approved):
Thanks @Kachoc! 

It would also be nice to regenerate the dist/gcc-compatible files and add a C test for aes-gcm :)

PR #477, opened 2021-09-17
Separate Hacl_Hash into more self-contained files

msprotz:
@victor-dumitrescu there's a fix for one of your recent commits which broke the configure script on OSX... can you take a look? thanks!


dist/gcc-compatible/EverCrypt_HMAC.h
```diff
@@ -38,11 +38,13 @@ extern "C" {
 
 
 #include "Hacl_Kremlib.h"
+#include "Hacl_Hash_SHA1.h"
 #include "Hacl_HMAC.h"
 #include "Hacl_Impl_Blake2_Constants.h"
-#include "Hacl_Hash.h"
+#include "Hacl_Hash_SHA2.h"
```
franziskuskiefer, line 44: Unrelated to this PR but just some food for thought.
I think we'll want to allow selecting algorithms for things like hmac as well in future. This will require some conditionals here on the includes as well as the functions.

franziskuskiefer (approved):
Thanks, lgtm!
The CI is unhappy but that appears to be a merge issue.

We should have caught the OSX issue on CI. And I see
```
./configure: line 121: $file: ambiguous redirect
clang: error: no input files
```
in the logs https://github.com/project-everest/hacl-star/runs/3619648470.
This should have failed the run. Instead `configure` determined that `__int128` is not supported. (But I'd vote for ignoring this issue and focus on making the new system better instead.)

msprotz:
>  This should have failed the run. Instead configure determined that __int128 is not supported. (But I'd vote for ignoring this issue and focus on making the new system better instead.)

Yes this is precisely what my patch to `configure` fixes. It all stems from a non-POSIX option to `mktemp` that prevented the generation of a temporary filename, which sent everything after that off the rails.

@victor-dumitrescu I also see that mktemp on OSX does not support the `p` option, so that's another situation that will fail if /tmp is not writable on a non-GNU system.

@franziskuskiefer agreed for HMAC I, too, would like this to be more modular... let me know if it's urgent or not

msprotz:
Ugh this is really a pain. And changing the name of the files really broke a bunch of our own internal clients, like hand-written C tests.

msprotz:
And now it's the WASM bindings that are causing trouble after the renaming. @denismerigoux any chance you can take a look? you might be able to fix this faster than I can. Thanks.

franziskuskiefer:
> Ugh this is really a pain. And changing the name of the files really broke a bunch of our own internal clients, like hand-written C tests.

This is what I meant by the trouble we'll be running in with these changes. Let's chat next week to plan how we're handling these sort of things.

victor-dumitrescu:
Hi, indeed I scrapped the `--suffix` option for the configure script in the opam package in version 0.4.1, I hadn't noticed that `-p` also fails because I don't think it was ever triggered on opam's OS X CI.

I'm a bit confused about the title and the description of the MR,  these are 2 separate issues right? (splitting Hacl_Hash and fixing the configure script)

franziskuskiefer:
> I'm a bit confused about the title and the description of the MR, these are 2 separate issues right? (splitting Hacl_Hash and fixing the configure script)

Yes, they are separate issues. The configure fix is just to make it work on OSX again. Depending on how long we need to get this merged we might want to pull out the fix and land it separately.

victor-dumitrescu (approved):

msprotz:
Finally. I also took this as an opportunity to fix the error messages in the JS/WASM glue layer to make sure it was easier to debug what went wrong.

PR #481, opened 2021-10-04
Fix intrinsics configuration bug

victor-dumitrescu:
An incorrectly set flag is causing intrinsics not to be used in the compiled code (the `configure` script is setting `HACL_CAN_COMPILE_INTRINSICS` in `config.h `but `lib_intrinsics.h` was looking for `COMPILE_INTRINSICS`)


msprotz:
copy-pasting what I wrote on Slack for reference:

> your PR looks good
> 9:33
> but in configure, the test is trying to compile a file that includes lib_intrinsics.h
> 9:33
> and lib_intrinsics.h now does, with your patch, #ifdef HACL_COMPILE_INTRINSICS so now the test in configure is no longer doing what's is supposed to do
> 9:34
> which is, enable intrinsics, and check whether the correct result is computed (edited) 
> 9:34
> (if not, the intrinsics are buggy because we have a buggy GCC and we should not use them)

the same thing needs to happen with `tests/Makefile` which is also incorrectly defining COMPILE_INTRINSICS via a `-D` flag, this now needs to be `-DHACL_CAN_COMPILE_INTRINSICS`

Thanks!

msprotz (approved):
Thanks!

franziskuskiefer (approved):
lgtm

PR #488, opened 2021-10-18
stabilize proofs for exponentiation

polubelova:


aseemr:
Thanks Marina! And sorry for breaking these proofs. I was about to the file a PR with restart-solver and what not, your changes are much much better : ).

PR #489, opened 2021-10-19
Don't call poly1305_padded without message

franziskuskiefer:


beurdouche:
@polubelova What's your take on being defensive everywhere? That change just by itself would cause Poly to have different expectations from the caller for non-optional inputs than the other primitives. There was a patch for this that was never finished I believe.

franziskuskiefer:
This is not about a defensive API. It's only a hotfix until it gets fixed properly. But messages can be empty and commonly used test vectors such as wycheproof test this (e.g. https://github.com/google/wycheproof/blob/2196000605e45d91097147c9c71f26b72af58003/testvectors/chacha20_poly1305_test.json#L45).
This only makes sure that there's no undefined behavior in hacl.

franziskuskiefer:
Expecting non-null pointer is not a sensible API decision 😉 That's something that has to change anyway.

beurdouche:
I agree, but I don't want to have different assumptions for each primitives and you can pass a non null pointer in the meantime.

polubelova:
Right, we need to finish that patch first, as, for example, having a check for null in `poly1305_update` doesn't mean that `poly1305_padded` is safe to use.

PR #492, opened 2021-11-05
Fix fragile proof in Hacl.Streaming.Functor

msprotz:
- Switch assertions over sequence equality (bad, unreliable, never works)
  to explicit calls to `Seq.lemma_eq_intro` (good, works better,
  somewhat more reliable).
- A couple manual calls to `B.modifies_trans` to guide the reasoning and
  rule out excessive modifies-clause reasoning.


polubelova (approved):
thanks!

nikswamy:
thanks for looking into this and fixing it right away, Jonathan. Much appreciated. 

I guess we'll see how this proof will fare as things churn. With rlimit 300, it's still a bit out of control.

As I mentioned on slack, from a z3 profile, the problem seems to be loc_disjoint reasoning firing too many quantifiers.

nikswamy (approved):

PR #495, opened 2021-11-16
Fix Blake2

Kachoc:
The spec of Blake2 currently does not properly handle the case where the key is non-empty but the data is. This PR fixes that.

Note that this modification in the spec makes it difficult to implement keyed mode for incremental blake2 and streaming blake2. As we currently don't use incremental/streaming blake2 in keyed mode, we removed those implementations (of course, there remains the keyed mode for the "standard" blake2 hash).


msprotz:
Looks like you simply forgot to update the Evercrypt instance of incremental hashes (the agile, multiplexing one)

```
[STDOUT]make[1]: Entering directory '/home/everest/hacl-star'
[STDERR]/home/everest/hacl-star/providers/evercrypt/fst/EverCrypt.Hash.Incremental.fst(104,68-104,72): (Error 189) Expected expression of type "Hacl.Streaming.Blake2.index"; got exp
ression "true" of type "Prims.bool"
[STDERR]1 error was reported (see above)
[STDERR]make[1]: *** [Makefile:408: obj/EverCrypt.Hash.Incremental.fst.checked] Error 1
```

code/streaming/Hacl.Streaming.Blake2.fst
```diff
@@ -733,49 +680,45 @@ let blake2 (a : alg) (m : valid_m_spec a)
...
-           (no_key : bool) (key_size : key_size_t a no_key) :
+           (finish : blake2_finish_st a m) :
   I.block unit =
   I.Block
     I.Erased (* key management *)
     
-    (stateful_blake2 a m) (* state *)
-    (k a no_key key_size) (* key *)
+    (stateful_blake2 a m)     (* state *)
+    (I.stateful_unused unit)  (* key *)
```
msprotz, line 689: why do we still have an argument here? I thought the `key` was dropped everywhere...?

msprotz: ah ok I get it -- blake2 has lost the ability to operate in streaming + key mode, but poly1305 still can operate with a key in incremental mode

code/streaming/Hacl.Streaming.Blake2.fst
```diff
@@ -787,156 +730,76 @@ let blake2 (a : alg) (m : valid_m_spec a)
...
-
-inline_for_extraction noextract
-let optional_key_blake2s (no_key : bool) (key_size : key_size_t Spec.Blake2S no_key) =
-  (I.optional_key () I.Erased (k Spec.Blake2S no_key key_size))
+         Blake2b32.blake2b_update_last Blake2b32.blake2b_finish
 
 inline_for_extraction noextract
-let optional_key_blake2b (no_key : bool) (key_size : key_size_t Spec.Blake2B no_key) =
-  (I.optional_key () I.Erased (k Spec.Blake2B no_key key_size))
+let optional_key_blake2 = I.optional_key () I.Erased (I.stateful_unused unit)
```
msprotz, line 743: why is this optional here? you should just pick mode = `Erased` for blake2 if you're just giving up on the ability to use the key

msprotz: ah ok this is just a helper that probably saves you an inversion... ok

msprotz (approved):
Ok this looks good to me. Thanks for fixing this long-standing issue. Some high-level remarks/questions:
- do you have any C tests to update for the streaming API? if there are no tests, it'd be good to add something brief in tests/c (there are other tests you can replicate for the streaming SHAs)
- I did not understand your comment that says we don't use streaming hash

Thanks!

karthikbhargavan:
Before we approve this, I'd like some more eyes on the F* spec in specs/Spec.Blake2.fst

Some history: a bug in our Blake2 implementation was pointed out by Mike (tankf33der@disroot.org) specifically in the case where the key is non-empty but the data is empty. 

https://datatracker.ietf.org/doc/html/rfc7693#section-3.3

I fixed the spec for Blake2 which was wrong again and then Son and I fixed it one more time.
The relevant spec code is in blake2_update_key (https://github.com/project-everest/hacl-star/blob/2b4917253158c3ae849f56c1f7d93c669305cd77/specs/Spec.Blake2.fst#L463)

I then proved that the Low* code matches this spec and Son propagated the proofs to top-level APIs.
So, everything again relies on the correctness of the F* spec.
Let's carefully check that this spec now correctly reflects the Blake2 RFC.

code/streaming/Hacl.Streaming.Blake2.fst
Kachoc: Yes exactly : I removed the keyed mode for streaming blake2 and incremental blake2 *only*. Which means we still have an argument for the stateful key in the streaming functor (for poly1305), which in the very specific case of blake2 we instantiate with `stateful_unused`.

code/streaming/Hacl.Streaming.Blake2.fst
Kachoc: Yes indeed, just a helper. I could inline it or maybe rename it to make it clearer. Would something like `unit_key` work?

Kachoc: I actually just renamed it to `unit_key`

Kachoc:
> Looks like you simply forgot to update the Evercrypt instance of incremental hashes (the agile, multiplexing one)
> 
> ```
> [STDOUT]make[1]: Entering directory '/home/everest/hacl-star'
> [STDERR]/home/everest/hacl-star/providers/evercrypt/fst/EverCrypt.Hash.Incremental.fst(104,68-104,72): (Error 189) Expected expression of type "Hacl.Streaming.Blake2.index"; got exp
> ression "true" of type "Prims.bool"
> [STDERR]1 error was reported (see above)
> [STDERR]make[1]: *** [Makefile:408: obj/EverCrypt.Hash.Incremental.fst.checked] Error 1
> ```

Thanks for this! I updated the code, let's see if CI passes this time

msprotz:
```
[STDERR]/usr/bin/ld: /home/everest/hacl-star/tests/blake2b-256-test-streaming.c:43: undefined reference to `Hacl_Streaming_Blake2b_256_blake2b_256_with_key_init'
[STDERR]/usr/bin/ld: /home/everest/hacl-star/tests/blake2b-256-test-streaming.c:44: undefined reference to `Hacl_Streaming_Blake2b_256_blake2b_256_with_key_update'
[STDERR]/usr/bin/ld: /home/everest/hacl-star/tests/blake2b-256-test-streaming.c:45: undefined reference to `Hacl_Streaming_Blake2b_256_blake2b_256_with_key_update'
[STDERR]/usr/bin/ld: /home/everest/hacl-star/tests/blake2b-256-test-streaming.c:46: undefined reference to `Hacl_Streaming_Blake2b_256_blake2b_256_with_key_update'
[STDERR]/usr/bin/ld: /home/everest/hacl-star/tests/blake2b-256-test-streaming.c:47: undefined reference to `Hacl_Streaming_Blake2b_256_blake2b_256_with_key_update'
[STDERR]/usr/bin/ld: /home/everest/hacl-star/tests/blake2b-256-test-streaming.c:48: undefined reference to `Hacl_Streaming_Blake2b_256_blake2b_256_with_key_update'
[STDERR]/usr/bin/ld: blake2b-256-test-streaming.o:/home/everest/hacl-star/tests/blake2b-256-test-streaming.c:49: more undefined references to `Hacl_Streaming_Blake2b_256_blake2b_256
_with_key_update' follow
[STDERR]/usr/bin/ld: blake2b-256-test-streaming.o: in function `main':
[STDERR]/home/everest/hacl-star/tests/blake2b-256-test-streaming.c:50: undefined reference to `Hacl_Streaming_Blake2b_256_blake2b_256_with_key_finish'
[STDERR]/usr/bin/ld: /home/everest/hacl-star/tests/blake2b-256-test-streaming.c:53: undefined reference to `Hacl_Streaming_Blake2b_256_blake2b_256_with_key_free'
[STDERR]collect2: error: ld returned 1 exit status
[STDERR]make[2]: *** [Makefile:88: blake2b-256-test-streaming.exe] Error 1
```

and many more. Just run `make` in tests/ to see what's up. Thanks.

msprotz:
I'll try to take a look at the spec and see if I see anything suspicious.

tankf33der:
Hi all, 

This Mike who found the bug, help me to generate C code from dir `dist/c89-compatible` and I will pass it thru my test case [here](https://github.com/tankf33der/monocypher-hacl).

Kachoc:
> Hi all,
> 
> This Mike who found the bug, help me to generate C code from dir `dist/c89-compatible` and I will pass it thru my test case [here](https://github.com/tankf33der/monocypher-hacl).

I pushed the regenerated C files. Do you manage to use them?

tankf33der:
Issue is gone in generated C code.

msprotz:
I reviewed the spec against the RFC (https://datatracker.ietf.org/doc/html/rfc7693). Some thoughts:
- there is not a single comment in the spec; readability would be greatly improved by a few pointers here and there, such as "G function, starting with some helpers", or "F function, which we call the compression function; we split it into three sub-functions because...???"
- some names are close to the spec (`kk`, `ll`, etc.) and this is great! but some others aren't, so again, just a word that e.g. `h` is called `s_iv` in the spec would help
- the spec in HACL differs from the RFC quite a bit and after staring at this for a while, I feel like the control-flow is needlessly complicated and could be amended to stick MUCH closer to the spec, which would significantly strengthen the claim that this spec is faithful vis à vis the RFC. Taking the existing names and conventions, can we aim for something really simple and declarative that simply follows the RFC and prepends a padded key block, if there is a key?

```fstar
val blake2_update:
    a:alg
  -> kk:size_nat{kk <= max_key a}
  -> k:lbytes kk
  -> d:bytes{if kk = 0 then length d <= max_limb a else length d + (size_block a) <= max_limb a}
  -> s:state a ->
  Tot (state a)
let blake2_update a kk k d s =
  let prefix = if kk > 0 then blake2_pad_key a kk k else Seq.empty in
  let d = prefix `Seq.append` d in
  let ll = length d in
  blake2_update_blocks a d s // get rid of the prev argument here
```

Right now the spec is testing for the length of the total message in two different places, and there is a contract that update_key will set to last-block flag correctly and then that the caller of update-key will *also* check the length and skip further updates in case it turns out the data was empty... This is probably correct, but the argument would be stronger with a simpler control-flow like above.

Or am I misunderstanding the RFC?

Thanks,

Jonathan

prosecco:
> I reviewed the spec against the RFC (https://datatracker.ietf.org/doc/html/rfc7693 <https://datatracker.ietf.org/doc/html/rfc7693>). Some thoughts:
> 
> there is not a single comment in the spec; readability would be greatly improved by a few pointers here and there, such as "G function, starting with some helpers", or "F function, which we call the compression function; we split it into three sub-functions because...???”
This is true for all our specs, where the only comments tend to be TODOs.
In cases where the spec is “obvious” the lack of comments results in a succinct mathematical definition, but this is clearly not the case here.
Still, I would put adding comments as a general goal for specs/ in a future pass.
> some names are close to the spec (kk, ll, etc.) and this is great! but some others aren't, so again, just a word that e.g. h is called s_iv in the spec would help
This would be nice yes, and not so hard to do.
> the spec in HACL differs from the RFC quite a bit and after staring at this for a while, I feel like the control-flow is needlessly complicated and could be amended to stick MUCH closer to the spec, which would significantly strengthen the claim that this spec is faithful vis à vis the RFC. Taking the existing names and conventions, can we aim for something really simple and declarative that simply follows the RFC and prepends a padded key block, if there is a key?

Restructing the spec is tempting and I myself did it at least three times. But it does not work as easily with Blake2 as you may imagine and it is likely that you will reintroduce the bug that we just removed.
The issue is that the RFC is truly imperative and a faithful implementation would need to be written in an incremental style.
Essentially the RFC expects that you always have a buffer of data that you’re filling and you only process this buffer when it is full or the data is at an end.
I tried rewriting the spec in that style and it is indeed possible to make it look much closer to the RFC but it requires a large chunk of the Low* code to be rewritten and re-verified.

If someone wants to try this, that is a fun experiment, and I’d be happy to help.
If we can agree that the current spec is “correct”, maybe we should merge it, and add an issue for a more RFC-like version of Blake2?

msprotz:
I'm fine with merging, and the style of specs is a long-standing debate that we're not going to solve on this PR :-). Just sharing my thoughts after taking an in-depth look at this particular specification. (Maybe we should start a discussion amongst ourselves regarding that, actually.)

By the way, I'm happy to also start gathering suggestions for how to author good specs if that helps. But that's less pressing.

karthikbhargavan (approved):

PR #498, opened 2021-11-23
Implementations for add_carry_u64 and sub_borrow_u64 for aarch64

victor-dumitrescu:
This PR introduces implementations for these intrinsics written in C, relying on the fact that uses of `__uint128_t` will be compiled to (reasonably) straightforward assembly that simply performs the u64 additions / subtractions and returns the result and the value of the carry flag, as seen here [here](https://godbolt.org/z/vW37x3zzf).

The result is faster than the fallback portable implementations in Hacl_IntTypes_Intrinsics. As an indication, on my Odroid C4 machine I get these times:

* `uint128_t_add_carry_u64`: 19 us
* `Hacl_IntTypes_Intrinsics_add_carry_u64`: 55 us
* `uint128_t_sub_borrow_u64`: 22 us
* `Hacl_IntTypes_Intrinsics_sub_borrow_u64`: 59 us

This in turn results in better performance on ARM for P-256 verification, which was the main motivation for this work. On the same machine it goes from 13669 us to 7163 us.

The test in `tests/tests/arm8-intrinsics-test.c` takes the measurements shown above and checks the output of these functions against the F* counterparts on randomly generated test vectors (committed in the PR, along with a script to generate them).


dist/gcc-compatible/lib_intrinsics.h
```diff
@@ -18,18 +18,32 @@
 
 #include "Hacl_IntTypes_Intrinsics.h"
 
-#define Lib_IntTypes_Intrinsics_add_carry_u32(x1, x2, x3, x4) \
```
msprotz, line 21: the second one is indeed redundant because I think ARMv8 mandates NEON (I think)

lib/c/arm8_intrinsics.h
```diff
@@ -0,0 +1,16 @@
+
+static inline uint64_t
+uint128_t_add_carry_u64(uint64_t cin, uint64_t x, uint64_t y, uint64_t* r) {
+  __uint128_t res = (__uint128_t)x + cin + y;
```
msprotz, line 4: I guess here you're relying on the fact that + is left-associative to get correct integer promotion rules?

msprotz (commented):
This all looks good to me minus my comment + my remark on Slack, which is that I seem to get 0 cycles/byte:

```
jonathan@armagnac:~/Code/everest/hacl-star/tests (vdum_arm8_intrinsics) $ ./arm8-intrinsics-test.exe 
uint128_t_add_carry_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 7 (0.00us/byte)
bw 22857.14 MB/s
Hacl_IntTypes_Intrinsics_add_carry_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 2 (0.00us/byte)
bw 80000.00 MB/s
uint128_t_sub_borrow_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 2 (0.00us/byte)
bw 80000.00 MB/s
Hacl_IntTypes_Intrinsics_sub_borrow_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 2 (0.00us/byte)
bw 80000.00 MB/s
```

Let me know if you think that's fixable easily or not. I don't want to hold a good PR hostage!

dist/gcc-compatible/lib_intrinsics.h
victor-dumitrescu: Looking around it does seem that aarch64 implies NEON. I think then the only strangeness might come in case an ARM8 processor is running in aarch32 mode, but I think in this case it would register as ARM7 to the configure script, right? I'm not sure how this works though.

lib/c/arm8_intrinsics.h
victor-dumitrescu: Yes. Is that not a good assumption?

victor-dumitrescu:
Looking at your output I was surprised to see the timings since it seems to show the C intrinsics to be slower. I got similar results on an M1 Mac Mini but this seems to be a fluke since the runtimes seem to only depend on the order in which the tests are ran:
```
Hacl_IntTypes_Intrinsics_add_carry_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 8 (0.00us/byte)
bw 20000.00 MB/s
uint128_t_add_carry_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 4 (0.00us/byte)
bw 40000.00 MB/s
Hacl_IntTypes_Intrinsics_sub_borrow_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 4 (0.00us/byte)
bw 40000.00 MB/s
uint128_t_sub_borrow_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 0 (0.00us/byte)
bw      inf MB/s
```

vs

```
int128_t_add_carry_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 8 (0.00us/byte)
bw 20000.00 MB/s
Hacl_IntTypes_Intrinsics_add_carry_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 4 (0.00us/byte)
bw 40000.00 MB/s
uint128_t_sub_borrow_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 4 (0.00us/byte)
bw 40000.00 MB/s
Hacl_IntTypes_Intrinsics_sub_borrow_u64:
cycles for 160000 bytes: 0 (0.00cycles/byte)
time for 160000 bytes: 1 (0.00us/byte)
bw 160000.00 MB/s

```
Do you know why this is the case?

franziskuskiefer:
Well, I'm not sure how useful the cycle output is. There's no implementation to get CPU cycles on ARM https://github.com/project-everest/hacl-star/blob/59758e26d70e1f448bc88a12f33d0d4dff333f79/tests/test_helpers.h#L85

I'd also suggest to add a warmup phase to make the measurements more reliable.

karthikbhargavan:
on arm we measure operation per second and then convert to cycle counts

msprotz:
I thought in any case clock() was not reliable by any means for benchmarking. Landon suggested a good way to benchmark *cycles* here: https://github.com/FStarLang/kremlin/issues/205 so maybe we could use that for M1/OSX, but other than that is there some cycle-count instruction for ARM that would work on all aarch64 targets?

msprotz:
Also the values returned by `clock` are so small that we might be within the margin of error / the round-up. The precision of `clock` really isn't super good.

dist/gcc-compatible/lib_intrinsics.h
msprotz: I would think so, yes. Or, it's a cross-compilation scenario.

lib/c/arm8_intrinsics.h
msprotz: No that's fine, just checking how that's supposed to work, that stuff is notoriously tricky.

victor-dumitrescu:
I tried @franziskuskiefer's suggestion of running a warm-up first but I get the same results. I can confirm however that benchmarking P-256 verification on M1 (through the OCaml API, using core_bench) still shows a big improvement, going from 2449 us to 805 us.

So I propose that we merge this PR as it is and open an issue about improving benchmarking on arm more generally?

karthikbhargavan:
Looks good to me. I assume we still use the HACL fall-backs for ARM32?

victor-dumitrescu:
Yes, we use these if
```
#if !defined(HACL_CAN_COMPILE_INTRINSICS)
[...]
#if (TARGET_ARCHITECTURE == TARGET_ARCHITECTURE_ID_ARM8) \
  && defined(HACL_CAN_COMPILE_VEC128)
```
and use the fallbacks if the second #if is false.

msprotz:
Looks good to me. This is really great, thanks! Wonderful to see so much perf improvement.

msprotz (approved):

lib/c/arm8_intrinsics.h
```diff
@@ -0,0 +1,16 @@
...
+  r[0U] = (uint64_t)res;
+  uint64_t c = res >> 64;
+  return c;
+}
+
+static inline uint64_t
+uint128_t_sub_borrow_u64(uint64_t cin, uint64_t x, uint64_t y, uint64_t* r) {
+  __uint128_t res = (__uint128_t)x - cin - y;
```
franziskuskiefer, line 12: Do we assume `x > y + cin` or is this checked anywhere?

lib/c/arm8_intrinsics.h
```diff
@@ -0,0 +1,16 @@
...
+static inline uint64_t
+uint128_t_add_carry_u64(uint64_t cin, uint64_t x, uint64_t y, uint64_t* r) {
+  __uint128_t res = (__uint128_t)x + cin + y;
+  r[0U] = (uint64_t)res;
+  uint64_t c = res >> 64;
+  return c;
+}
+
+static inline uint64_t
+uint128_t_sub_borrow_u64(uint64_t cin, uint64_t x, uint64_t y, uint64_t* r) {
```
franziskuskiefer, line 11: Not checking the carry is fine I guess for production. But having at least some debug checks that the inputs are safe to use would be nice. (For both functions.)

tests/arm8-intrinsics-test.c
```diff
@@ -0,0 +1,116 @@
...
+      printf("Test failed for %" PRIu64 " and %" PRIu64 " \
+with carry %" PRIu64 ":\n\
+Hacl_IntTypes_Intrinsics_sub_borrow_u64: res %" PRIu64 ", carry %" PRIu64 "\n\
+uint128_t_sub_borrow:                    res %" PRIu64 ", carry %" PRIu64 "\n",
+             a_vectors[i], b_vectors[i], cin_vectors[i], r2, cout2, r, cout);
+      passed = false;
+    }
+  }
+
+  /* currently neither the timings nor the cpucycle counts are reliable */
```
franziskuskiefer, line 57: Can you file an issue for this and add it here for reference?

franziskuskiefer (commented):
Some questions on checking inputs. Otherwise I'm fine with this.

lib/c/arm8_intrinsics.h
victor-dumitrescu: No, but it's expected that `x` can be less than `y + cin` and in that case a borrow is returned, right?

franziskuskiefer: Maybe. I the functions need at least a comment with a contract what exactly it is supposed to do.
Looking at the IntTypes implementation this appears to be the case. But I think we don't cover one case `FStar_UInt64_eq_mask(res, x) & cin`, i.e. when the result is the same as x (smaller than 2^64) and we had a carry in. In this case this function here wouldn't return a carry but should.

victor-dumitrescu: I don't think we're missing this case. Correct me if I'm wrong but the way I interpret `Hacl_IntTypes_Intrinsics_sub_borrow_u64` is the following:
* if res < x then c = 0
* if res > x then c = 1
* if res == x then either
  - y == 0 and cin == 0, so c = 0
  - y == UINT64_MAX and cin == 1, so c = 1
I don't see how res could be equal to x in any other way. If that's the case then I think the function here does correctly a carry in this situation.

franziskuskiefer: hm, I think you're right 😊

franziskuskiefer:
Let's get this in and file follow ups for benchmarking on ARM and documenting these functions.

lib/c/arm8_intrinsics.h
polubelova: you can verify that code directly in F*/Low* :) It supports 64-bit and 128-bit integers.

msprotz: good point Marina, that's actually how the "fallback" implementation works... that would be an excellent intern project, see https://github.com/project-everest/hacl-star/tree/master/code/fallback/

polubelova: It might be a good idea to drop that fallback implementation in favor of a new one if the performance numbers on intel are also better.

victor-dumitrescu:
Superseded by #503, closing. Thanks all for your reviews!

PR #501, opened 2021-12-10
More precise modifies annotations in Vale curve25519

R1kM:
Jason reported issues with Vale's inline assembly when used with old versions of GCC.
This was tracked down to some inline assembly annotations being imprecise: some operands were marked as input/output operands while they were only read inside the code, and should have been marked as input operands only.
See https://lore.kernel.org/wireguard/CAHmME9rFHkydmrYcS83PEf=MJP8brwsdis+7TQ=oLkO4RNTX9g@mail.gmail.com/T/#mc0dce91964a6658441a01c03870a0402625c80a6 for the full details.

The root cause for this was being overly conservative in Vale code with "modifies" annotations. In particular, the register r12 in fsqr and fsqr2 was marked as modified while it is only read. Same for r15 in fmul and fmul2.
This PR changes the Vale annotations to be more precise.
The diff for generated code is consequent as operands in inline assembly get reordered, but this change is actually very minor.
The trusted code responsible for Vale code generation is unchanged.


msprotz (approved):

zx2c4:
Before you merge this, note https://lore.kernel.org/wireguard/3a1be638-4ddd-44c0-87aa-56cd26f4f396@grsecurity.net/

msprotz:
Thanks Jason for the pointer. I was offline last week and saw your messages re. this just this morning. Thanks for all the investigation and detailed bug report!

R1kM:
Thanks Jason for the link to the latest thread. I'm hesitant to add this to the Vale codegen. As mentioned on Slack, the possibility for arguments to be passed in either registers or memory independently is currently out of scope of our Vale model. While this might be okay for these specific functions, supporting "rm" systematically and generally would require significant changes to our current trusted computing base.

zx2c4:
:+1: That makes sense. You might want to update Mathias on that thread, but all good from my side. I'll submit my side of things to LKML this week based on this PR.

parno (approved):
Changes look reasonable to me.  It's interesting how such small changes in the verified code produced large diffs in the `.h` files.  Makes me glad the changes were verified!

Chris-Hawblitzel (approved):

PR #503, opened 2021-12-17
Faster fallback implementations for add_carry_u64 and sub_borrow_u64

victor-dumitrescu:
This PR supersedes #498. It provides the same functionality, but now verified in Low* instead of being hand-written in C, thanks to @polubelova.

Another change is that these implementations are now used on any non-x64 platform that supports uint128, not just ARMv8 as in the previous PR.

The gain in performance for P-256 operations is the same as before, roughly 2x.

I've tried to generalise Marina's proofs for the `*_u64` versions in order to use it for the `*_u32` versions too. Not having much experience with this, let me know if there's a better way of doing this or if it would be better to simply duplicate and adapt the proofs, as I initially did. Thanks!


code/fallback/Hacl.IntTypes.Intrinsics.fst
```diff
@@ -9,8 +9,27 @@ open Lib.Buffer
 
 open FStar.Mul
 
-#set-options "--fuel 0 --ifuel 0 --z3rlimit 100"
+#set-options "--fuel 0 --ifuel 1 --z3rlimit 200"
```
msprotz, line 12: I usually recommend limiting ifuel to a minimum, so maybe here you could use #push-options and #pop-options for the definitions that need ifuel 1? alternatively, keep ifuel 1 and use `let _ = allow_inversion inttype` in the definitions that match on `t`

code/fallback/Hacl.IntTypes.Intrinsics.fst
```diff
@@ -58,10 +83,62 @@ let sub_borrow_st (t:inttype{t = U32 \/ t = U64}) =
...
+  c
+
+val sub_borrow_u32: sub_borrow_st U32
+let sub_borrow_u32 = sub_borrow U32 U64 32ul 64ul
+
+
+(* Fallback versions of add_carry_u64 and sub_borrow_u64 for platforms which
+   don't support uint128.
+   The names Hacl.IntTypes.Intrinsics.add_carry_u64 and sub_borrow_u64 must not
+   be changed because they are hardcoded in KreMLin for extracting wasm code
```
msprotz, line 123: I see this:
```
src/CFlatToWasm.ml:  | CallFunc ("Lib_IntTypes_Intrinsics_add_carry_u64", args) ->
src/CFlatToWasm.ml:      mk_expr env (CallFunc ("Hacl_IntTypes_Intrinsics_add_carry_u64", args))
```
so I guess it's hardcoded?

dist/gcc-compatible/Hacl_IntTypes_Intrinsics.h
```diff
@@ -42,10 +42,18 @@ extern "C" {
 static inline uint32_t
 Hacl_IntTypes_Intrinsics_add_carry_u32(uint32_t cin, uint32_t x, uint32_t y, uint32_t *r)
 {
-  uint32_t res = x + cin + y;
-  uint32_t
-  c = (~FStar_UInt32_gte_mask(res, x) | (FStar_UInt32_eq_mask(res, x) & cin)) & (uint32_t)1U;
-  r[0U] = res;
+  uint64_t res = (uint64_t)x + (uint64_t)cin + (uint64_t)y;
```
msprotz, line 45: is this faster, then?

code/fallback/Hacl.IntTypes.Intrinsics.fst
```diff
@@ -58,10 +83,62 @@ let sub_borrow_st (t:inttype{t = U32 \/ t = U64}) =
...
+   which uses these intrinsics. *)
+inline_for_extraction noextract
```
msprotz, line 125: I thought there *might* have been a use-case for an add-carry for 32-bit platforms without native uint64 but this seems unused here... so the `fallback` version does not rely on a wider type while the u64 version from this file does?

dist/gcc-compatible/Hacl_IntTypes_Intrinsics_128.c
```diff
@@ -0,0 +1,55 @@
+/* MIT License
```
msprotz, line 1: this file seems superfluous... was it leftover in your dist directory?

dist/gcc-compatible/lib_intrinsics.h
```diff
@@ -18,19 +18,33 @@
 
 #include "Hacl_IntTypes_Intrinsics.h"
 
-#define Lib_IntTypes_Intrinsics_add_carry_u32(x1, x2, x3, x4) \
-  (Hacl_IntTypes_Intrinsics_add_carry_u32(x1, x2, x3, x4))
+#if defined(HACL_CAN_COMPILE_VEC128)
```
msprotz, line 21: I don't think this is the right test. Based on the configure test, I would suggest adding after line 504 of the configure script:

```
else
  echo "#define HACL_CAN_COMPILE_UINT128 1" >> config.h
fi
```

and then using that

tests/gen_vectors/gen_uint128_intrinsics_vectors.py
```diff
@@ -0,0 +1,66 @@
+import os
```
msprotz, line 1: can you make sure this test generates boundary conditions, e.g. UINT_MAX, UINT_MAX, along with 0, UINT_MAX, and UINTMAX, 0?

msprotz (approved):
As always, a very nice piece of work. Thanks to you both @polubelova and @victor-dumitrescu. Congratulations!

This seems generally good to me, but I have a few (minor) concerns. Please address those before merging.

msprotz:
Adding Marina as a reviewer to see if she has any thoughts about the details of your proofs.

msprotz:
btw your build is hitting known breakage in karatsuba and streaming (the current regression)

dist/gcc-compatible/Hacl_IntTypes_Intrinsics.h
franziskuskiefer: Some numbers again for both changes would be nice.

franziskuskiefer (approved):
Thanks!

code/fallback/Hacl.IntTypes.Intrinsics.fst
```diff
@@ -25,25 +44,31 @@ let add_carry_st (t:inttype{t = U32 \/ t = U64}) =
...
 
-
 val add_carry_u32: add_carry_st U32
-let add_carry_u32 cin x y r = add_carry #U32 cin x y r
+let add_carry_u32 = add_carry U32 U64 32ul
 
-val add_carry_u64: add_carry_st U64
-let add_carry_u64 cin x y r = add_carry #U64 cin x y r
 
+inline_for_extraction noextract
```
polubelova, line 65: you can replace this function with `uint #t #SEC` from `Lib.IntTypes`

code/fallback/Hacl.IntTypes.Intrinsics.fst
```diff
@@ -25,25 +44,31 @@ let add_carry_st (t:inttype{t = U32 \/ t = U64}) =
...
-  let res = x +. cin +. y in
-  let c = logand (logor (lt_mask res x) (logand (eq_mask res x) cin)) (uint #t 1) in
-  r.(0ul) <- res;
-  logand_lemma (eq_mask res x) cin;
-  logor_lemma (lt_mask res x) (logand (eq_mask res x) cin);
-  logand_mask (logor (lt_mask res x) (logand (eq_mask res x) cin)) (uint #t 1) 1;
+val add_carry:
+  t:inttype{t = U32 \/ t = U64} ->
+  t':inttype{if t = U32 then t' = U64 else t' = U128} ->
+  shift_bits:UInt32.t{if t = U32 then shift_bits == 32ul else shift_bits == 64ul} ->
```
polubelova, line 51: `shift_bits` should be a function:
```
let shift_bits (t:inttype{t = U32 \/ t = U64}) =
  match t with
  | U32 -> 32ul
  | U64 -> 64ul
```

code/fallback/Hacl.IntTypes.Intrinsics.fst
```diff
@@ -58,10 +83,62 @@ let sub_borrow_st (t:inttype{t = U32 \/ t = U64}) =
...
+#push-options "--z3rlimit 600"
 
 inline_for_extraction noextract
-val sub_borrow: #t:inttype{t = U32 \/ t = U64} -> sub_borrow_st t
-let sub_borrow #t cin x y r =
+val sub_borrow:
+  t:inttype{t = U32 \/ t = U64} ->
+  t':inttype{if t = U32 then t' = U64 else t' = U128} ->
+  shift_bits:UInt32.t{if t = U32 then shift_bits == 32ul else shift_bits == 64ul} ->
+  p:UInt32.t{if t = U32 then p == 64ul else p == 128ul} ->
```
polubelova, line 93: similar to `add_carry`, `shift_bits` and `p` should be functions

polubelova (approved):
Thanks @victor-dumitrescu!
I would just duplicate the proofs :)

code/fallback/Hacl.IntTypes.Intrinsics.fst
victor-dumitrescu: I'm not sure I understand, are you saying that there is a requirement for having `add_carry_u32` and `sub_borrow_u32` implementations that don't rely on uint64? If so, I can keep both versions, similar to what I've done for `Hacl.IntTypes.Intrinsics_128.add_carry_u64`, which uses uint128, and `Hacl.IntTypes.Intrinsics.add_carry_u64`, which is just the current fallback version.

dist/gcc-compatible/Hacl_IntTypes_Intrinsics_128.c
victor-dumitrescu: I don't think it's superfluous, I split `Hacl.IntTypes.Intrinsics` into 2 files, the uint128-compatible part and the current fallback implementations.

msprotz: but the definitions are also in the .h file right below https://github.com/project-everest/hacl-star/pull/503/files/93bffad7ac99d060e71a00d99995dcf1bf508a6b#diff-a75ea96722ac2fa74fa24a86706cc737abef098e98a605580837a5b138a209c9 so they appear once as `static inline` in the `.h` and once more in the `.c`? if so, that would be a kremlin bug

code/fallback/Hacl.IntTypes.Intrinsics.fst
msprotz: yes, but maybe I'm thinking about a far-fetched use-case (a 32-bit embedded device or something) and we can re-instate those functions later on if the need arises

code/fallback/Hacl.IntTypes.Intrinsics.fst
victor-dumitrescu: Yes, I remember that I hardcoded this in Kremlin when we wanted to get a wasm version of P-256 and couldn't because of the missing intrinsics. Should we do this another way or it's not worth the hassle?

dist/gcc-compatible/Hacl_IntTypes_Intrinsics_128.c
victor-dumitrescu: Ah yes you're right and indeed it was a leftover file

dist/gcc-compatible/Hacl_IntTypes_Intrinsics.h
victor-dumitrescu: As we discussed in the previous PR, we can't really measure these accurately on ARM at the moment but I can measure the P-256 functions and these are significantly faster. For example, on a small Odroid C4 board I get these speed-ups (in microseconds):
```
P-256/verify 13684 -> 7218
P-256/compressed_to_raw 271 -> 147
P-256/dh_initiator 4240 -> 2247
P-256/valid_pk 4224 -> 2233
```

I wanted to confirm these on an M1 Mac too as I did originally but I don't have access to one right now, I'll report when I will.

victor-dumitrescu: I can confirm these speed-ups on M1 Mac Mini too:
```
P-256/P-256/verify 2542 -> 803
P-256/P-256/compressed_to_raw 50 -> 16
P-256/P-256/dh_initiator 785 -> 247
P-256/P-256/valid_pk 787 -> 248
```

msprotz: 👏

PR #506, opened 2021-12-29
Switch to internal headers to improve quality of dist/

msprotz:
Not going to merge this anytime soon, but putting it up for discussion. See details in https://github.com/FStarLang/kremlin/commit/0990d8cc8189abce8a5e1a38121a5766249fd32e

This is a big change and will need careful regression testing in Everest.


dist/gcc-compatible/Hacl_Chacha20.h
```diff
@@ -39,21 +39,6 @@ extern "C" {
 
 #include "Hacl_Kremlib.h"
 
-extern const uint32_t Hacl_Impl_Chacha20_Vec_chacha20_constants[4U];
```
msprotz, line 42: this is typically the kind of change I really like to see

dist/gcc-compatible/Hacl_Ed25519.h
```diff
@@ -37,31 +37,10 @@ extern "C" {
 #include "kremlin/internal/target.h"
 
 
-#include "Hacl_Kremlib.h"
 #include "Hacl_Streaming_SHA2.h"
+#include "Hacl_Kremlib.h"
 #include "Hacl_Hash_SHA2.h"
 #include "Hacl_Bignum25519_51.h"
-#include "Hacl_Curve25519_51.h"
```
msprotz, line 44: this is also a really nice chunk that goes away

dist/gcc-compatible/Hacl_Hash_Blake2.h
```diff
@@ -47,76 +46,6 @@ extern "C" {
 
 typedef uint8_t Hacl_Impl_Blake2_Core_m_spec;
 
-uint64_t Hacl_Hash_Core_Blake2_update_blake2s_32(uint32_t *s, uint64_t totlen, uint8_t *block);
```
msprotz, line 50: all of these were previously exposed (in spite of being private) because Hacl_Streaming_Blake2 needed to "see" those variants -- now they're only in the internal header

franziskuskiefer:
Thanks! Really nice change. I didn't check all details yet but the changes generally look good and what I would expect.
Another related change would be separating header and source files such that there's an `include` directory that can be used for packaging.

beurdouche (approved):
This looks reasonable to me, thanks. @polubelova and @Frosne are blocking reviews.

polubelova (approved):
This looks good to me, thanks @msprotz!

Frosne (approved):
Looks reasonable! Thanks

Issue #507, opened 2021-12-29
Strange warnings related to OCaml bindings show up during build


msprotz:
```
[STDERR]File "../../.hacl_star.objs/byte/hacl_star.odoc":
[STDERR]Warning: Couldn't find the following modules:
[STDERR]  Ctypes Ctypes_static EverCrypt_AutoConfig2_stubs Hacl_Hash_Base_stubs
[STDERR]  Hacl_Hash_Blake2_stubs Hacl_Hash_MD5_stubs Hacl_Hash_SHA1_stubs
[STDERR]  Hacl_Hash_SHA2_stubs Stdlib Unsigned Z
[STDERR]File "SharedDefs.ml", line 334, characters 73-89:
[STDERR]Warning: Failed to resolve reference unresolvedroot(ecdsa) Couldn't find "ecdsa"
[STDERR]File "SharedDefs.ml", line 331, characters 98-115:
[STDERR]Warning: Failed to resolve reference unresolvedroot(points) Couldn't find "points"
[STDERR]File "SharedDefs.ml", line 334, characters 73-89:
[STDERR]Warning: Failed to resolve reference unresolvedroot(ecdsa) Couldn't find "ecdsa"
[STDERR]File "SharedDefs.ml", line 331, characters 98-115:
[STDERR]Warning: Failed to resolve reference unresolvedroot(points) Couldn't find "points"
[STDERR]File "Hacl.mli", line 363, characters 26-37:
[STDERR]Warning: Failed to resolve reference unresolvedroot(Detached) Couldn't find "Detached"
[STDERR]File "Hacl.mli", line 363, characters 14-21:
[STDERR]Warning: Failed to resolve reference unresolvedroot(Easy) Couldn't find "Easy"
[STDERR]File "Hacl.mli", line 306, characters 0-31:
[STDERR]Warning: Failed to resolve reference unresolvedroot(mac) Couldn't find "mac"
[STDERR]File "EverCrypt.mli", line 65, characters 8-18:
[STDERR]Warning: Failed to resolve reference unresolvedroot(decrypt) Couldn't find "decrypt"
[STDERR]File "EverCrypt.mli", line 64, characters 75-85:
[STDERR]Warning: Failed to resolve reference unresolvedroot(encrypt) Couldn't find "encrypt"
[STDERR]File "EverCrypt.mli", line 64, characters 10-17:
[STDERR]Warning: Failed to resolve reference unresolvedroot(init) Couldn't find "init"
```

[...]

PR #508, opened 2022-01-03
OCaml API: Fixes for odoc documentation

victor-dumitrescu:
Addresses #507 

The warnings reported in the issues above come from using the latest version of odoc (2.0.2). From what I can gather these warnings have a few different origins:
- Some reflected changes in odoc that broke some links, which I've fixed in this PR
- Some are expected, pointing to modules from other packages to which odoc doesn't have access (see https://github.com/ocaml/odoc/issues/794). This doesn't cause any issue and can just be ignored.
- Some seem to indicate unresolved references to sections in other pages, which I think happen because these references go in the opposite direction in the dependency graph to that of their underlying modules. However, in the generated documentation these references are valid so I think these warnings can also be ignored.

If you'd rather not have these warning in the CI output at all, we can look at how to silence them.


msprotz (approved):
This is great and I am happy to have the warnings -- it means we can catch them and fix them. Thanks for the fix!

Kachoc (approved):

PR #514, opened 2022-01-19
remove hacl-star-raw before installing

franziskuskiefer:
Running `make` twice fails because we try to re-install `hacl-star-raw`. With this change `hacl-star-raw` is removed before it's being installed again.


victor-dumitrescu (approved):

msprotz (approved):
👍 thanks!

franziskuskiefer:
Looks like the CI is broken?

msprotz:
unrelated to you I need to submit my PR to remove the benchmarks... coming up today

PR #515, opened 2022-01-19
Remove C++ benchmarks

msprotz:
This PR is to remove the C++ benchmarks from the repository. I've looped in various stakeholders to approve this change -- if you agree, please approve this PR.

Here's the rationale:
- we have not used these benchmarks since 2019 when we submitted the EverCrypt paper
- the maintenance cost is very high:
  - no one "owns" these benchmarks -- they were originally written by Christoph Wintersteiger but he's now moved on to some other projects
  - they regularly break the build
  - they have a foreign build system with a CMakeLists.txt that includes every file by name, so whenever something changes in build/packaging, this is yet another unrelated place that you have to go fix -- everyone (me included) forgets about these and this is extra added tax on every badly-needed cleanup that we attempt
- they caused e.g. linking errors on Windows so they're already disabled on Windows
- none of us are C++ gurus, so it's yet another language / build system / philosophy that we have to understand, and none of us are invested in maintaining this
- these benchmarks don't use state-of-the-art measurements like e.g. kbench9000
- Marina has been maintaining a set of benchmarks in a separate repository that has been used for recent papers

The only thing these benchmarks have brought us since 2019 is more maintenance burden. The current breakage is the straw that broke my camel's back, so I'm now advocating for their removal.

Please speak up if you have any objections. Thanks!


parno (approved):
@msprotz makes a compelling case.  Seems reasonable to me.

tahina-pro:
Thank you Jonathan, I fully agree that maintaining these benchmarks has become practically less and less feasible over time.
For reference, could you or @polubelova please tell (here or in the docs) the location of Marina's new benchmark repository? Thank you in advance.

Chris-Hawblitzel (approved):

R1kM:
I also agree with removing these benchmarks, feel free to merge

beurdouche:
I am fine with this as long as this is ok with Marina and Natalia 👍

Frosne:
It seems to be fine with me!

polubelova:
Looks good to me, thanks!

For the haclxn paper, we used SUPERCOP and KBENCH9000 for benchmarking.
The results are available in the following repo: https://github.com/polubelova/hacl-star-benchmarks

PR #516, opened 2022-01-24
Describe what ecp256dh_i does.

msprotz:
Thanks @Frosne!


msprotz:
This is just a hotfix for two specific comments, but I would like to do a pass over this module to:
- have consistency for the order of arguments (destination argument is sometimes first, sometimes second)
- have consistency regarding the naming of functions and arguments
- general fixup of comments, introducing early on all three representations: compressed, not compressed, and uncompressed (are there better names?!)

@franziskuskiefer and @Frosne I'd be happy to hear your thoughts on this... thanks!

code/ecdsap256/Hacl.P256.fsti
```diff
@@ -332,7 +332,12 @@ val compression_not_compressed_form: b: lbuffer uint8 (size 64) -> result: notCo
     )
 
 
-[@ (Comment " Input: a point buffer (internal representation: uint8[64]), \n result: a point in not compressed form (uint8[33]).")]
+[@ (Comment "Convert from internal representation to 33-byte compressed form.
```
franziskuskiefer, line 335: Because the compressed form is rarely used it might be useful to add what the first byte is (0x02 or 0x03 depending Y)? Not too important though.

code/ecdsap256/Hacl.P256.fsti
```diff
@@ -348,10 +353,15 @@ val compression_compressed_form: b: lbuffer uint8 (size 64) -> result: compresse
...
 
 
-[@ (Comment " Input: result: uint8[64], \n scalar: uint8[32].
-  \n Output: bool, where True stands for the correct key generation. 
-  \n False means that an error has occurred (possibly that the result respresents point at infinity). 
-  ")]
+[@ (Comment "Convert a private key into an uncompressed public key.
+
+  Input: `scalar`, the private key, of type `uint8[32]`.
+  Output: `result`, the public key, of type `uint8[64]`.
```
franziskuskiefer, line 359: So this sounds like this isn't prefixed with the format (0x04)?
This should be made consistent with the compressed form. But for the comment we should mention what's in these 64 bytes.

msprotz: Then say this is the concatenation of `r`and `s`?

code/ecdsap256/Hacl.P256.fsti
msprotz: There's some terminology discussion here: https://github.com/project-everest/hacl-star/blob/master/dist/gcc-compatible/Hacl_P256.h#L228 earlier in the file, so I'm assuming this has been introduced already?

Realistically we need to do a big overhaul of this file, and re-think documentation, APIs and naming conventions. In the meanwhile, I'm basically trying to fix up the most glaring omissions because I have some folks struggling with the API.

Let me know if you would like a reminder of what the compressed form is, in this specific function, or if the earlier explanation is ok.

code/ecdsap256/Hacl.P256.fsti
msprotz: This file introduces a terminology of "compressed (33), not compressed (65) and internal also called uncompress (64)". See comment earlier in the file. 

The OCaml API has a different take, see here: https://hacl-star.github.io/ocaml_doc/hacl-star/Hacl_star/Hacl/P256/index.html#points

I could try to align this comment to what is done in the OCaml API, and then maybe fix up the earlier introduction of terminology? Would that help? I'm trying to steer away from a complete overhaul of the documentation of this file (even though I'm quite tempted...).

code/ecdsap256/Hacl.P256.fsti
franziskuskiefer: Ah, I missed the terminology. I guess it's fine for now.

code/ecdsap256/Hacl.P256.fsti
franziskuskiefer: Officially the 65 bytes (including the 0x04) is the uncompressed form (compared to the compressed form) - see RFC 5480 or SEC1. So I think using "not compressed" is a little confusing. The 64-byte version (concatenation of `r` and `s`) is usually referred to as raw. But there's no official term as it's not part of any standard afaik.
(The API really shouldn't work on the raw values, but I think that's a more general issue and not for this comment.)

msprotz: I updated the documentation to refer to `raw`, `compressed` and `uncompressed`, of size 64, 33 and 65 respectively. This should hopefully clarify.

msprotz:
I did a general pass. Hopefully this should be enough for now.

msprotz:
Please only review Hacl_P256.h -- I haven't pushed changes to the rest of the files, CI will take care of that. Thanks.

dist/gcc-compatible/Hacl_P256.h
```diff
@@ -100,7 +125,19 @@ Hacl_P256_ecdsa_sign_p256_sha512(
 );
 
 /*
- Input: result buffer: uint8[64], 
+P256 signature WITHOUT hashing.
+
+This is NOT RECOMMENDED. Please use of the combined hash-and-sign functions
```
franziskuskiefer, line 130:
```suggestion
This is NOT RECOMMENDED. Please use the combined hash-and-sign functions
```

dist/gcc-compatible/Hacl_P256.h
```diff
@@ -249,30 +312,48 @@ bool Hacl_P256_decompression_not_compressed_form(uint8_t *b, uint8_t *result);
...
+  `scalar` and `result` MUST NOT overlap.
 */
 bool Hacl_P256_ecp256dh_i(uint8_t *result, uint8_t *scalar);
 
 /*
- 
+ECDH key agreement.
+
+This function takes a 32-byte secret key, another party's 64-byte raw public
+key, and computeds the 64-byte ECDH shared key.
```
franziskuskiefer, line 355:
```suggestion
key, and computes the 64-byte ECDH shared key.
```

dist/gcc-compatible/Hacl_P256.h
```diff
@@ -249,30 +312,48 @@ bool Hacl_P256_decompression_not_compressed_form(uint8_t *b, uint8_t *result);
...
+
+  `scalar` and `result` MUST NOT overlap.
 */
 bool Hacl_P256_ecp256dh_i(uint8_t *result, uint8_t *scalar);
 
 /*
- 
+ECDH key agreement.
+
+This function takes a 32-byte secret key, another party's 64-byte raw public
```
franziskuskiefer, line 354: Public keys must be validated before use. I don't think that's happening in here so the comment should tell users to first validate the inputs.

dist/gcc-compatible/Hacl_P256.h
```diff
@@ -120,6 +157,11 @@ Hacl_P256_ecdsa_sign_p256_without_hash(
   uint8_t *k
 );
 
+
+/****************/
+/* Verification */
```
franziskuskiefer, line 162: Same comment as for ecdh. Public keys need to be checked before use.

franziskuskiefer (changes requested):
Thanks, a couple nits

msprotz:
Thanks Franziskus, I've pushed an update to add the two request comments. Thanks for fixing the typos.

code/ecdsap256/Hacl.P256.fsti
```diff
@@ -452,6 +458,9 @@ val ecp256dh_i:
 This function takes a 32-byte secret key, another party's 64-byte raw public
 key, and computeds the 64-byte ECDH shared key.
 
+The user MUST validate keys with `is_more_than_zero_less_than_order` and
```
polubelova, line 461: we could simply expose one function called `is_public_key_valid` :)

msprotz: thanks Marina, I've pushed further changes that update the names in the API to be more informative -- I have essentially chosen @victor-dumitrescu's naming conventions

franziskuskiefer (approved):
Thanks!

beurdouche:
This was all discussed with Natalia who owns that code I imagine?

karthikbhargavan:
@Frosne is designated as a reviewer for this PR? Natalia, can you take a look if the documentation added here is correct?

msprotz:
The initial version of this PR was following discussions with Natalia on Slack, hence my original comment thanking her for the explanations :-). 

Since then, there have been several requests for improvements or further changes from the reviewers, which I have implemented.

@Frosne is also a review. Any chance you can comment on the most recent changes?

Frosne:
> The initial version of this PR was following discussions with Natalia on Slack, hence my original comment thanking her for the explanations :-).
> 
> Since then, there have been several requests for improvements or further changes from the reviewers, which I have implemented.
> 
> @Frosne is also a reviewer. Any chance you can comment on the most recent changes?

Sure :)

code/ecdsap256/Hacl.P256.fsti
```diff
@@ -19,10 +19,30 @@ open Spec.ECDSAP256.Definition
...
+between various point representations, and ECDH key agreement.
+
+*******************************************************************************/
+
+/**************/
+/* Signatures */
+/**************/
+
+/*
+  The standard strongly encourages the user to hash the input before signing
```
Frosne, line 37: I am not sure about "strongly encourages". Usually, the hash of the message is signed, true, but I think "strongly encouraged" gives an impression of an attack if it is not the hash

code/ecdsap256/Hacl.P256.fsti
```diff
@@ -19,10 +19,30 @@ open Spec.ECDSAP256.Definition
...
+  it. Therefore, we recommend using one of the signature variants below.
+*/";
+
+Comment "Combined SHA2-256 and P256 signature function.
```
Frosne, line 41: Maybe "A message is first signed using SHA2-256, and then signed"?

code/ecdsap256/Hacl.P256.fsti
```diff
@@ -114,7 +138,19 @@ val ecdsa_sign_p256_sha512: result: lbuffer uint8 (size 64)
   )
 
 
-[@ (Comment " Input: result buffer: uint8[64], \n m buffer: uint8 [mLen], \n priv(ate)Key: uint8[32], \n k (nonce): uint32[32]. 
+[@ (Comment "P256 signature WITHOUT hashing.
+
+This is NOT RECOMMENDED. Please use of the combined hash-and-sign functions
```
Frosne, line 143: It's not about it. It's more that we expect to have a hash here. This is the function used by Tezos iirc.

code/ecdsap256/Hacl.P256.fsti
```diff
@@ -147,7 +183,18 @@ val ecdsa_sign_p256_without_hash: result: lbuffer uint8 (size 64)
...
 
 
-[@ (Comment " The input of the function is considered to be public, 
+[@@ CPrologue "
+/****************/
+/* Verification */
+/****************/
+
+/*
+  The user MUST validate the public key before calling any of the
```
Frosne, line 192: The verification function already validates the key itself.

code/ecdsap256/Hacl.P256.fsti
```diff
@@ -258,16 +311,42 @@ val verify_q:
       )
     )
 
+[@@ Comment "Validate a private key.
+
+Input: scalar: uint8[32].
+  \n Output: bool, where true stands for the scalar to be more than 0 and less than order."]
+val validate_private_key: x: lbuffer uint8 (size 32) -> Stack bool
```
Frosne, line 318: Any reason why you need a private key validation?

code/ecdsap256/Hacl.P256.fsti
```diff
@@ -258,16 +311,42 @@ val verify_q:
...
+[@@ CPrologue "
+/*****************************************/
+/* Point representations and conversions */
+/*****************************************/
+
+/*
+  Elliptic curve points have 2 32-byte coordinates (x, y) and can be represented in 3 ways:
+
+  - \"raw\" form (64 bytes): the concatenation of the 2 coordinates, also known as \"internal\"
+  - \"compressed\" form (33 bytes): the first byte is either \x02 or \x03, followed by x
```
Frosne, line 336: I would have put that the first byte is equal to 0x2 + (y % 2); otherwise, it sounds a bit random, wdyt?

code/ecdsap256/Hacl.P256.fsti
```diff
@@ -258,16 +311,42 @@ val verify_q:
...
+  - \"uncompressed\" form (65 bytes): the first byte is always \04, followed by the \"raw\" form
+
+  For all of the conversation functions below, the input and output MUST NOT overlap.
+*/
+";
+ (Comment "Convert 65-byte uncompressed to raw.
+
+This function effectively strips the first byte of the uncompressed representation.
```
Frosne, line 344: It also checks that the representation is correct.

code/ecdsap256/Hacl.P256.fsti
Frosne: I think it could be needed only for ecdsa

code/ecdsap256/Hacl.P256.fsti
msprotz: Can you go ahead and suggest an alternate phrasing? Don't hesitate to commit on the branch.

msprotz: I've reworded to keep closer to the standard.

code/ecdsap256/Hacl.P256.fsti
msprotz: Reworded too.

code/ecdsap256/Hacl.P256.fsti
msprotz: I rewrote.

code/ecdsap256/Hacl.P256.fsti
msprotz: changed

code/ecdsap256/Hacl.P256.fsti
msprotz: it was in the API to begin with, just with a surprising name; it is also exposed as a private key validation function by the OCaml bindings (see https://hacl-star.github.io/ocaml_doc/hacl-star/Hacl_star/Hacl/P256/index.html)

code/ecdsap256/Hacl.P256.fsti
msprotz: not sure the details of the representation matter to the clients... I was just aligning myself to what the OCaml API describes. Unless you're implementing P256, do you need to know what `0x02` and `0x03` correspond to for the initial byte? I feel like clients don't need to know about internal representation details.

code/ecdsap256/Hacl.P256.fsti
msprotz: What does it mean for the representation to be correct? That the `x` coordinate does not exceed the order of the group? Does it perform a full call to `validate_public_key`? It would be good to be precise here. I've added a comment but if you could clarify what you mean that'd be great.

msprotz:
@Frosne: thanks for your comments. I've addressed all of them, except "checks that the representation is correct"...

My suggestion: "The function errors out if the first byte is incorrect, or if the resulting point is invalid, i.e. its `x` coordinate exceeds the order of the group."

If you like this, please go ahead, commit and merge. Otherwise, I'll pick this up again tomorrow morning. Thanks.

code/ecdsap256/Hacl.P256.fsti
franziskuskiefer: Technical details on the first byte are probably not necessary. But usually this is described as the sign byte or even/odd. Maybe some wording like that would help?

dist/gcc-compatible/Hacl_P256.h
```diff
@@ -218,16 +272,37 @@ Hacl_P256_ecdsa_verif_without_hash(
...
+
+/*****************************************/
+/* Point representations and conversions */
+/*****************************************/
+
+/*
+  Elliptic curve points have 2 32-byte coordinates (x, y) and can be represented in 3 ways:
+
+  - "raw" form (64 bytes): the concatenation of the 2 coordinates, also known as "internal"
+  - "compressed" form (33 bytes): the first byte is either  or , followed by x
```
victor-dumitrescu, line 295: Missing the byte value here, gets removed from the comment during extraction? Same in the C file.

code/ecdsap256/Hacl.P256.fsti
Frosne: What about mentioning https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186-draft.pdf (D.2 Point Compression section)?

code/ecdsap256/Hacl.P256.fsti
Frosne: We call the representation to be correct if the first byte is 0x4/0x2 or 0x3. Probably, it's not so important

dist/gcc-compatible/Hacl_P256.h
msprotz: nice catch Victor, thanks

code/ecdsap256/Hacl.P256.fsti
msprotz: Ok, but looking at the implementation of decompressionCompressedForm (now known as `compressed_to_raw`), there is also a check `lessThanPrime`... should that be mentioned as one of the possible reasons for failure, too?

code/ecdsap256/Hacl.P256.fsti
msprotz: To me, as a person who's not an expert on P256, I find the NIST standard quite unhelpful. I feel like you're already an expert of P256 and then you know where the standard is, or you're a client and you don't want to hear about those details. I'll try to rephrase to include sign bit.

Frosne: Ok :)

Frosne (approved):
It's fine with me. I think we discussed and solved all the possible issues.

Impressive contribution, guys! Thanks a lot! 
Viva la P256.

Issue #511, opened 2022-01-12
Build error on certain s390x machines


victor-dumitrescu:
The `configure` script currently passes the following compilation flags for `*_128.c` files on s390x: `-m64 -mzarch -mvx -mzvector -march=z15"`. The `-march=z15` flag causes compilation to fail on older machines (e.g. z14).

The solution seems to be passing `-march=native` instead, but I'm opening this issue in case there are any objections / alternative solutions.

CC @Kachoc @msprotz

Kachoc:
We implemented support for IBMz following IBM's demand. From what I remember we used `-march=z15` because we wanted the best performance (even though I don't think this should have any impact), some of the intrinsics we use only work on some recent configs, and IBM only needed support for z15. Following this, I'm ok with changing this to `-march=native`.

victor-dumitrescu:
The change from `z15` to `native` was made in #518.

Closing for now, but might need to revise this choice later depending on priorities.

PR #518, opened 2022-01-25
Add missing s390x intrinsics

victor-dumitrescu:
After the merge of #314 the build broke on s390x and PowerPC due to some required intrinsics on these platform not being available in `lib/c/libintvector.h`. This PR adds these intrinsics for s390x, along with tests for each one:

- `Lib_IntVector_Intrinsics_vec128_load32_be`
- `Lib_IntVector_Intrinsics_vec128_store32_be`
- `Lib_IntVector_Intrinsics_vec128_shift_right32`
- prototype for `Lib_IntVector_Intrinsics_vec128_xor`, required by gcc on s390x

The tests for SHA2 multibuffer now pass on s390x and the opam package can be installed.

Another change, previously discussed in #511, is switching the `-march` flag from `z15` to `native`, which allows the code to be built on older z architectures.

The build remains broken on PowerPC.

Thanks @Kachoc for your help with this PR!


Kachoc (approved):
Thanks a lot for this!
Could you just add the proper debugging functions in `lib/c/libintvector_debug.h`? This file is used to generate execution traces (it can happen - and actually happened - that the unit tests for the intrinsics in `tests/` are not exhaustive enough, in which case we need to debug the execution of the primitives themselves).

franziskuskiefer (approved):
Thanks! lgtm

PR #519, opened 2022-01-27
Modifying the way has_include is used

Frosne:
#### Summary
The PR modifies the has_include ('config.h') like in order to support the compilers where 'has_include' itself is not defined.

#### Motivation
We would like to support gcc4.4 and gcc4.8 compilers that fail because of the has_include. 

This PR…
-->
- [x] Modifies the has_include to have a wrapper around it checking that has_include is supported


msprotz (approved):
Thanks Natalia, this looks good. I have two questions.

Can you please explain a little bit more about intrin.h? Which toolchains is it needed on? What functions does it provide that were not in scope already? We've been successfully building on Windows via GitHub actions so it'd be great to document why we need another inclusion, when CI says we're fine.

Second request: the other obvious option is to always require config.h, and to always require that people who integrate the library create a (possibly-empty) config.h. It's less flexible, but has less macro-checking. Can you say a little more about why you prefer to define features via compiler flags?

None of these are issues -- it's just that we have been going back and forth on these things a lot, so if there's a good reason to pick one solution over the other, I want to document for the future. Thanks!

lib/c/evercrypt_targetconfig.h
```diff
@@ -18,11 +18,13 @@
 #define TARGET_ARCHITECTURE_ID_SYSTEMZ 5
 #define TARGET_ARCHITECTURE_ID_POWERPC64 6
 
+#if defined(__has_include)
 #if __has_include("config.h")
 #include "config.h"
 #else
 #define TARGET_ARCHITECTURE TARGET_ARCHITECTURE_ID_UNKNOWN
```
franziskuskiefer, line 25: So `TARGET_ARCHITECTURE` isn't defined at all when `__has_include` is not available? We might want this defined in any case where we don't have the `config.h`.

lib/c/libintvector.h
```diff
@@ -25,6 +27,10 @@
 #include <tmmintrin.h>
 #include <smmintrin.h>
 
+#if defined(_MSC_VER)
+  #include <intrin.h>
```
franziskuskiefer, line 31: As @msprotz mentioned, some information on this would be good. Why is this needed now and what wasn't working before?

franziskuskiefer (changes requested):
Thanks. Two nits

lib/c/evercrypt_targetconfig.h
msprotz: I'm wondering if we should just get rid of TARGET_ARCHITECTURE. It doesn't seem to provide much added value and brings us further away from our goal of thinking exclusively in terms of features (rather than in terms of specific targets)

franziskuskiefer: I wouldn't mind getting rid of it. But this PR changes when and how it gets added. I think we should keep the logic here and then think about removing it.

msprotz: agreed that this should be a separate PR... it's just that we don't use that thing anywhere else (as far as I can tell) except for hacl_vec128_not_avx () and we could refactor that code to not depend on it...

Frosne:
> Thanks Natalia, this looks good. I have two questions.
> 
> Can you please explain a little bit more about intrin.h? Which toolchains is it needed on? What functions does it provide that were not in scope already? We've been successfully building on Windows via GitHub actions so it'd be great to document why we need another inclusion, when CI says we're fine.

Small helpful link: https://dev.to/yumetodo/list-of-mscver-and-mscfullver-8nd.

It seems that our windows build requires it. Otherwise, our build returns the following exception: `_mm_cmpgt_epi64': intrinsic function not declared. ` The problem gets solves once we add intrin.h file. 

We use 191125547 MSC_VER(sion), HACL* CI runs using 192930139 Version.

lib/c/libintvector.h
msprotz: I think I just remembered: intrin.h breaks the build when using cygwin-packaged, native mingw compilers. We should have CI for that toolchain, too.

Frosne: Would it then be better to have smth like: 
```

#ifdef _MSC_VER
#if _MSC_VER > version
 
 #include <intrin.h> 
#endif
#endif
```

Frosne: Or will mingw be already filtered out by the `#ifdef _MSC_VER` clause?

msprotz: I'm not sure -- I think you would need to check. Basically, I'm fine with any solution, as long as it doesn't break the mingw-based setup that several of us currently rely on.

@tahina-pro do we still have windows CI anywhere in our system that could check Natalia's patch? Or alternatively, could you tell us quickly if `#include <intrin.h>` works with x86_64-mingw64-gcc?

tahina-pro: Right now we have no working Windows CI on our side. We only have a nightly `./everest make test` on Windows, which has been failing for months now...
Re. intrin.h with x86_64-mingw64-gcc, I managed to compile a dummy C file containing just that include, but I guess that won't tell us much. I don't fully understand here, but I think the header in question might be https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/crt/intrin.h and a related discussion might be https://mingw-w64-public.narkive.com/9uy03D8V/what-is-the-purpose-of-intrin-h

franziskuskiefer:
> It seems that our windows build requires it. Otherwise, our build returns the following exception: `_mm_cmpgt_epi64': intrinsic function not declared. ` The problem gets solves once we add intrin.h file.

Another issue here is that `_mm_cmpgt_epi64` is SSE4.2 but is used when `HACL_CAN_COMPILE_VEC128` is set. But `HACL_CAN_COMPILE_VEC128` doesn't actually check for SSE4.2 but only for SSE2 and SSE3

lib/c/libintvector.h
msprotz: thanks Tahina... so `intrin.h` appears to define MSVC-like macros for mingw... at this point I'm thoroughly confused why we need that header

maybe to re-center the discussion, @frosne, you can provide us with the compilation error that this include fixes? thanks!

Frosne: It looks like it's indeed required by mingw

franziskuskiefer: I would suggest pulling this change out of this PR and handling it separately as it appears to need clarification. The `__has_include` is uncontroversial and could go in.

Looking at the [official documentation for MSVC](https://docs.microsoft.com/en-us/cpp/intrinsics/x86-intrinsics-list?view=msvc-170) `intrin.h` should be the only one needed. The others (`tmmintrin.h` and `smmintrin`) appear to be outdated.

MinGW supports `intrin.h` as well such that it should work for both environment.
We will have a Windows CI in a couple weeks to test this on and I might be able to tell you more next week. But I have to set up a Windows environment first.
If someone else has a Windows setup, they could give this a go with `intrin.h` only.

msprotz (approved):
All looks good to me, thanks!

msprotz:
@franziskuskiefer this PR now has the intrin.h bit removed and only concerns __has_include

beurdouche (approved):

franziskuskiefer (approved):
Thanks!

PR #530, opened 2022-02-22
Renaming variables named `as`

aseemr:
F* PR https://github.com/FStarLang/FStar/pull/2464 makes `as` a keyword in F*. This PR renames the uses of `as` variable name.

I may not have been able to keep the variable names convention intact, so please feel free to change the names that I have chosen as you see fit (of course to something other than `as` : ) ).


polubelova (approved):
Thanks!

aseemr:
Thanks Marina, Benjamin!

msprotz (approved):
Thanks Aseem for the heads-up, and no worries, if someone has a strong opinion about variable naming, they can always submit a followup PR later on...

PR #531, opened 2022-02-22
Modification of libintvector to include intrin.h

Frosne:
#### Summary

This PR changes the way the intrinsics are included in the libintvector.h file. 

#### Motivation
(Otherwise, our build system returns error C2164: '_mm_cmpgt_epi64': intrinsic function not declared)

### DIFF
--- a/lib/c/libintvector.h
+++ b/lib/c/libintvector.h
@@ -7,9 +7,8 @@
  * properly brought into scope. Users can either run the configure script, or
  * write a config.h themselves and put it under version control. */
 #if __has_include("config.h")
- #include "config.h"
+#include "config.h"
 #endif
-
 
 /* # DEBUGGING:
  * ============
@@ -937,7 +936,6 @@ typedef vector128_8 vector128;
 // If libintvector_debug.h exists, use it to debug the current implementations.
 // Note that some flags must be enabled for the debugging to be effective:
 // see libintvector_debug.h for more details.
-
 #if __has_include("libintvector_debug.h")
 #include "libintvector_debug.h"
 #endif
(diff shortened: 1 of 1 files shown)


msprotz:
Thanks for the separate PR and the compilation error. So, the Intel manual says this intrinsic is found in nmmintrin.h

https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_cmpgt_epi64&ig_expand=1038

I do have nmmintrin.h on both my Linux and OSX boxes... this is useful to see what headers are available

```
jonathan@absinthe:/tmp $ find $(gcc-11 --print-sysroot) -iname '*intrin.h' | xargs basename | sort | uniq
```

Two questions:
- Is this header available on Windows? Would be great to find the "right" header so that we can settle on the standard way of including this rather than having an ifdef for Windows, which apparently is only needed for some versions of MSVC.
- Which Visual Studio are you working with? We have had 2019 and 2022 on CI and both worked alright with the current set of header inclusions.

msprotz:
@franziskuskiefer have you noticed any missing headers in hacl-packages for the Windows CI regressions? @Frosne can you tell us which version of Visual Studio this happens with?

If we need intrin.h, then let's figure out which setup requires it, and then maybe add this particular CI configuration to hacl-packages so that it's covered by CI.

Frosne:
I think I managed to make it work for us by adding the intrin.h file as a part of a makefile, so I don't insist on addressing this problem.

Frosne:
I would still prefer to have it opened till I check that with the new code (SHA3.Streaming) sha3 will keep working.

franziskuskiefer:
> @franziskuskiefer have you noticed any missing headers in hacl-packages for the Windows CI regressions? @Frosne can you tell us which version of Visual Studio this happens with?

Nope, this appears to work fine on our Windows CI.

msprotz:
@Frosne have you integrated SHA3? are things working out alright?

karthikbhargavan:
If this is working now, maybe we can close this?

PR #532, opened 2022-02-28
Modification of SHA3 code

Frosne:
#### Summary

The PR contains a small modification of the SHA3 code. 

#### Motivation 
The extracted code did not have a constant length allocation of a buffer. Instead, the length of the buffer to allocate depended on the SHA3 variable. It caused a problem when running the code for windows (or anything that does not support it). 

Additionally I have modified the inlining philosophy of the file. I removed the extract keyword from the two functions I would like to be able to use in our code. 

#### Updates
this fixes https://github.com/project-everest/hacl-star/issues/540


polubelova (commented):
could you please also check in the code for `dist/gcc-compatible`?

code/sha3/Hacl.Impl.SHA3.fst
```diff
@@ -377,6 +377,28 @@ let storeState rateInBytes s res =
       copy res (sub block 0ul rateInBytes)
     )
 
+(* 
+   Lemma_equivalence_state_rate proves that it is equivalent
```
polubelova, line 381: it's `lemma_equiv_alloc_bigger_state`, no?

Frosne:
@polubelova the build system is not gonna regenerate the code itself in a day? Or you would like to compare the modifications?

code/sha3/Hacl.Impl.SHA3.fst
Frosne: :) Sure

polubelova:
Yes, it will do that, but I want to see the diff :)

Frosne:
Ok, gonna do.

code/sha3/Hacl.Impl.SHA3.fst
```diff
@@ -377,6 +377,28 @@ let storeState rateInBytes s res =
...
+   to allocate a bigger temporary buffer and then use a part of it 
+   compared to allocation of the exact size buffer
+*)
+
```
s-zanella, line 385: This lemma is overly specific. Can't it be omitted or inlined in the definition of `absorb_next` and `absorb_last`? If not, can you at least generalize it by dropping the refinement in the parameter and generalizing the length and type?

Frosne: I will try! Thanks for the suggestion.

s-zanella:
I second @polubelova's suggestion to check in the extracted code to see the diff. Besides the extraction of `absorb_next` and `absorb_last` and avoiding the variable-length allocation this change should be transparent.

msprotz (commented):
I took a quick look at this, and the function seems *more* unsafe than the previous version, because now a C caller can pass in a rate that is > 200 by mistake and the C code won't error out. Is there a defensive version somewhere that fails if a C caller does that?

I'm also concerned about 200 that seems like an arbitrary bound. Why not perform a heap allocation instead? You would get rid of the arbitrary limit *and* of the `alloca`

Thanks!

Frosne:
200 here is a number that is bigger than any possible rate (https://en.wikipedia.org/wiki/SHA-3, Instances section, there is a table. The table has rates in bits, the code uses the same, but in bytes).

Frosne:
But I can indeed introduce a check inside absorb_last/absorb_inner functions, such that we accept any input and fail if it's less than 200.

msprotz:
There are several tables in the article, which one are you referring to?

If you are adamant that 200 is a well-chosen constant and suitable for all use-cases, then the function needs to be defensive. Right now, I'm afraid the stack can easily be blown *in HACL code*; should this ever lead to a security vulnerability, the blame would fall on us. (You could add a comment saying "all bets are off if you use a rate > 200", but few people read comments, and a defensive API would be superior.)

The return type could change to bool to indicate success, for instance.

Frosne:
I am slightly worried though that adding a check to each absorb call will decrease the speed of the code.

code/sha3/Hacl.Impl.SHA3.fst
```diff
@@ -412,25 +433,63 @@ val absorb_last:
...
+      absorb_next s rateInBytes;
+      lemma_equiv_alloc_bigger_state #200 rateInBytes)
+
+val compare_more_zero_lte_b: a: size_t -> b: size_t -> Tot (r: bool {
+  r == not (v a > 0 && v a <= v b)})
+
+let compare_more_zero_lte_b a b = 
+  let a_is_zero = eq a (size 0) in 
+  let a_lte_200 = gt a b in 
+  a_is_zero || a_lte_200
```
s-zanella, line 461:
```suggestion
  let a_gt_b = gt a b in 
  a_is_zero || a_gt_b
```

code/sha3/Hacl.Impl.SHA3.fst
```diff
@@ -412,25 +433,63 @@ val absorb_last:
...
       lastBlock.(rem) <- byte_to_uint8 delimitedSuffix;
       loadState rateInBytes lastBlock s;
       if not ((delimitedSuffix &. byte 0x80) =. byte 0) &&
          (size_to_UInt32 rem = size_to_UInt32 (rateInBytes -. 1ul))
       then state_permute s;
-      absorb_next s rateInBytes)
+      absorb_next s rateInBytes;
+      lemma_equiv_alloc_bigger_state #200 rateInBytes)
+
+val compare_more_zero_lte_b: a: size_t -> b: size_t -> Tot (r: bool {
```
s-zanella, line 455: I'm not sure that introducing this auxiliary function helps. I'd rather inline the check.
If you prefer to keep it, it would need to be marked as `inline_for_extraction noextract` and the name could be improved (`positive_and_lte`?).

code/sha3/Hacl.Impl.SHA3.fst
```diff
@@ -412,25 +433,63 @@ val absorb_last:
...
+
+val absorb_last:
+    delimitedSuffix:byte_t
+  -> rateInBytes:size_t
+  -> rem:size_t
+  -> input:lbuffer uint8 rem
+  -> s:state
+  -> Stack bool
+    (requires fun h -> live h s /\ live h input /\ disjoint s input)
+    (ensures  fun h0 flag h1 ->
```
s-zanella, line 472: `flag` -> `success`?
In my mind, a _flag_ is raised when there is an error.

s-zanella (changes requested):
Can you please re-generate the gcc-compatible code once more?
You were worried that the extra runtime checks would impact performance. Can you compare performance before/after this PR?

code/sha3/Hacl.SHA3.fst
```diff
@@ -148,3 +150,29 @@ let cshake256_frodo input_len input cstm output_len output =
   absorb s 136ul input_len input (byte 0x04);
   squeeze s 136ul output_len output;
   pop_frame ()
+
+[@ (Comment "Keccak instantiation with the rate equal to 1088 bits ")]
+val keccak_1088:
```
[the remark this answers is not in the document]
Frosne: IIRC Tezos is using exactly this version of keccak together with SHAs. In this case, the only modification they need to do is to use this function instead of keccak (and remove the rate). The extracted keccak function has a different API - it returns bool instead. So, it was just a way to simplify their life adapting the changes :)

msprotz:
hi @Frosne, what is the status of this PR? Santiago had a question:

> Can you compare performance before/after this PR?

Do you have some stats?

Thanks!

msprotz:
@Frosne I am also noticing that this line is incorrect: https://github.com/project-everest/hacl-star/pull/532/files#diff-2362ae7382c03775602df49a51177c268f05d03b7825dcf292b3fa18f92e3e53R13

you need `$(shell uname)`, not `$(uname)`. Thanks.

msprotz:
To summarize... the following needs to be fixed for this PR to move forward:
- Santiago's comment needs to be addressed
- https://github.com/project-everest/hacl-star/pull/532#issuecomment-1169187922 needs to be addressed
- the *second* point of https://github.com/project-everest/hacl-star/issues/540#issuecomment-1075440160 needs to be addressed

Thanks!

msprotz:
I took another look at this: to fix the build on OSX, it might be the case that in addition to point 2, you need this change (untested) as well:

```diff
diff --git a/Makefile b/Makefile
index adb3d4528b..4ac7ca9cf2 100644
--- a/Makefile
+++ b/Makefile
@@ -980,7 +980,7 @@ dist/mozilla/Makefile.basic: \
 dist/mozilla/Makefile.basic: VALE_ASMS := $(filter dist/vale/curve25519-%,$(VALE_ASMS))
 dist/mozilla/Makefile.basic: HAND_WRITTEN_OPTIONAL_FILES =
 dist/mozilla/Makefile.basic: HAND_WRITTEN_H_FILES := $(filter %/libintvector.h,$(HAND_WRITTEN_H_FILES))
-dist/mozilla/Makefile.basic: HAND_WRITTEN_FILES =
+dist/mozilla/Makefile.basic: HAND_WRITTEN_FILES = $(filter %/Lib_Memzero0.c,$(HAND_WRITTEN_FILES))
 dist/mozilla/Makefile.basic: TARGETCONFIG_FLAGS =
 dist/mozilla/Makefile.basic: HAND_WRITTEN_LIB_FLAGS =
 dist/mozilla/Makefile.basic: TARGET_H_INCLUDE = -add-early-include '<stdbool.h>'
```

msprotz:
Will you resubmit this? I'm actually quite keen on getting rid of that VLA. @polubelova and I chatted about this, and if we put Hacl.Impl.SHA3 on the right-hand side of the -bundle argument, then `absorb` won't be exposed to C clients so there won't be a concern that a C client may blow the stack by passing an incorrect argument.

msprotz:
Also, #649 will solve the build break issue so this will be one less thing to fix here.

PR #533, opened 2022-02-28
Added CHANGES.md

Frosne:
#### Summary

After the discussion about HACL* governance, we decided that it may be nice to have to track the modifications.

I am extremely open to any suggestions if we want to modify the file I added etc.


msprotz (approved):
I am broadly in favor of this, but we need consensus to make sure no pull request is merged without an entry in Changes.md. Others?

karthikbhargavan:
Questions:
- What is the suggested format of Changes.md? Should we add a pointer to the commit?
- Do we add a line for each change, e.g. does an rlimit change in some file, or some Makefile edit merit a line?
- How do we mark security-relevant changes and bugfixes? Should they be marked out in some way, e.g. by linking to an advisory? 

Larger open-source projects only mark changes during releases (https://github.com/openssl/openssl/blob/master/CHANGES.md)
So we need to answer this for ourselves.

msprotz:
@Frosne any thoughts on this?

@karthikbhargavan I think a link to the PR would be even better, example:

```
- #533 Start tracking changes in Changes.md (Natalia Kulatova)
```

We explicitly exclude maintenance: build fixes, dist changes, stabilizing proofs, etc.

Security-relevant changes should probably go into their own section. We "cut" the list of changes for each release and then start new section, as done in e.g. https://github.com/ocaml/ocaml/blob/trunk/Changes

Frosne:
@msprotz, I agrew with you about the format of the message. 
On the other hand, I still think that stabilizing proofs might be good to mention in the code (if the modification was big and touched several functions, for instance).

msprotz:
Fair enough if it's a big change that touches several files, maybe worth giving others a heads-up. Let's do some categories then:

```
# Current release

## New features
- #123 Added AVX-2048 support for SHA-5

## Bug fixes
- #234 Fix SHA-5 for empty input

## Proof improvements
- #345 Rewrite SHA-5 to use repeat_blocks_multi_loop_i to improve proof stability

**Comment by msprotz** [comment]
We talked about this at a recent maintainer discussion. Here's what we wrote in the meeting notes:

> Not everything should go in CHANGES.md, only user-facing changes (breaking APIs, adding new features, perf improvements of the executable code, bumping versions of F*/Z3, ...)

@karthikbhargavan seems to have the most formed thoughts on this, so @karthikbhargavan will create a [pull request template](https://github.blog/2016-02-17-issue-and-pull-request-templates/) that will provide guidelines on how to format Changes.md and how to add new entries.

File: CHANGES.md [hunk 1]
```diff
@@ -0,0 +1,28 @@
...
+
+
+## 2022-12-01
+
+### Changed
+
+- Added a template to CHANGES.md
+
+
+## 2022-02-28
```
polubelova, line 24: Should we start from 2022-12-01? Otherwise, it looks like nothing happened between February and December

polubelova (approved):

PR #535, opened 2022-03-09
Clean up local makefiles

pnmadelaine:


msprotz (approved):
Ok, this seems generally like a good idea... why is the realpath important?

I would like someone else to check this who actually uses those local makefiles. Can you ask @polubelova for a review?

Thanks!!

polubelova (approved):
I've checked it locally, it works as expected. Thanks!

@pnmadelaine why are you working with a forked repo of hacl-star and not with the main one?

pnmadelaine:
@polubelova only out of habit I guess, I'm used to the fork/pr workflow

polubelova:
@pnmadelaine CI doesn't run external PRs, so you might end up pushing these changes to your branch of the main hacl-star repo

pnmadelaine:
@msprotz I answered with a comment on the changes, which may not be the most readable option.

There were some makefiles where, for instance, `-I../../lib/c` *and* `-I../../../lib/c` were passed to gcc.
This was a workaround the fact that gcc is called by the makefile from `./` but also from `./dist`.
Using an absolute path for HACL_HOME should guarantee we don't run in situations like this in the future, and allow replacing both flags with a single instance of `-I$(HACL_HOME)/lib/c`.

msprotz:
@pnmadelaine I don't see the comment on the changes, where is it? 

Note that force-pushing makes it hard to follow what's going on...
- If your comment was a GitHub comment, then it's probably lost on the previous history that got re-written
- If your comment was added in the source code via an additional commit, I can't tell what happened on the pull request because every single change was a rewrite of the history, so I see a complete diff instead of incremental changes when you push

pnmadelaine:
@msprotz sorry for the confusion, my previous message in this discussion is essentially a copy of the original comment

PR #543, opened 2022-03-25
unroll first iteration of bn-mul

polubelova:
This PR unrolls the first iteration of schoolbook multiplication to avoid extra additions with zeros. 
It also makes the post-condition of `bn_field_init` more accurate.

### DIFF
--- a/code/bignum/Hacl.Bignum.Karatsuba.fst
+++ b/code/bignum/Hacl.Bignum.Karatsuba.fst
@@ -194,7 +194,7 @@ let rec karatsuba () = karatsuba_open karastuba
 
 inline_for_extraction noextract
 let bn_karatsuba_mul_st (t:limb_t) =
-    len:size_t{4 * v len <= max_size_t}
+    len:size_t{0 < v len /\ 4 * v len <= max_size_t}
   -> a:lbignum t len
   -> b:lbignum t len
   -> tmp:lbignum t (4ul *! len)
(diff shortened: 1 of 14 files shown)


msprotz (commented):
Does this generate more efficient code? It seems like it's adding a bunch of C code, are C compilers happier with this?

polubelova:
I need to double-check it with @victor-dumitrescu as my generated asm code is quite different from his one. 

Here are some performance numbers for a 256-bit schoolbook multiplication used in p256 and compiled with gcc-9:
```
p256-mul: 49.35 cycles
bignum-mul: 41.32 cycles
bignum-mul-unroll1: 40.98 cycles
```
for `bignum-mul-unroll1`, gcc generates less `add` and `adc` instructions than for `bignum-mul`(4 and 4, respectively), and clang outputs the same asm code.

Anyway, addition is a cheap operation, so we may not need to add such "optimization".

Frosne:
That's pretty surprising about the p256 mult. Probably there's some
inlining missing there? Or compiler was cleverer in the bignum case :/

polubelova:
@Frosne you could try to run this [test](https://github.com/project-everest/hacl-star/blob/vdum_multiplication_benchmarks/tests/multiplication-test.c) on your laptop (yon need to run first `make` in `hacl-star/dist/gcc-compatible` and then run `make multiplication-test.exe` in `hacl-star/tests`). No surprise if you get another numbers.

Frosne:
Sure!

msprotz:
@Frosne any news on this? how does this interact with your other PR?

karthikbhargavan:
Looks like this is an easy win, so I am for it.

polubelova:
It would be great to see the performance numbers for P256-master, P256-MR #563 and P256-MR #543 (if we rebase MR #543 on top of MR #563, so that P256 uses the above "optimization" for schoolbook multiplication). If they are good, we can merge these MRs. @Frosne: could you please post them here?

polubelova:
Closing this PR as it doesn't improve the performance of `bn_mul` compared with the most recent version from `main`.

PR #545, opened 2022-03-31
document API for Ed25519

polubelova:
This PR also removes extra condition on a message size (cf. #447).


code/ed25519/Hacl.Impl.SHA512.ModQ.fst
```diff
@@ -1,31 +1,32 @@
...
   -> input:lbuffer uint8 len ->
   Stack unit
     (requires fun h ->
       live h hash /\ live h prefix /\ live h input /\
       disjoint input hash /\ disjoint prefix hash)
     (ensures fun h0 _ h1 -> modifies (loc hash) h0 h1 /\
       as_seq h1 hash == Spec.Agile.Hash.hash Spec.Hash.Definitions.SHA2_512
-           (concat #uint8 #32 #(v len) (as_seq h0 prefix) (as_seq h0 input))
-    )
+        (Seq.append (as_seq h0 prefix) (as_seq h0 input)))
```
msprotz, line 29: Is this where the restriction on the extra 64 bytes is lifted? Because you use `Seq.append` which can reason on > uint32_max long sequences...?

msprotz (approved):
This looks pretty good overall. Thanks for the comments that's really excellent!

I'm wondering where the magic happens and the extra 64 bytes of padding disappear. Can you confirm or infirm my guess? I'm curious what happened...

Thanks!

Jonathan

code/ed25519/Hacl.Impl.SHA512.ModQ.fst
polubelova: yes, that's correct

karthikbhargavan (approved):
Looks good, thanks.

PR #547, opened 2022-04-14
Large overhaul of the JS API for HACL-WASM

msprotz:
I am opening this PR to start the review process since this is likely going to be a big one. At a high-level, this PR does the following.
- Simplify and clean up the JS bindings. There were a few things that weren't super smooth, so I've simplified them. Notably:
  - I separate the parsing of size expressions from their evaluation.
  - The "remember the index" strategy used here: https://github.com/project-everest/hacl-star/blob/master/bindings/js/api.js#L238 was confusing and error-prone, and I've removed it in favor of a combination of `map` and `filter`
  - Large amounts of comments notably for how the lengths are automatically determined
- Extend the grammar of size expressions to support multiplication and division
- Allow support for "ghost" length variables; it is now ok to have an argument list such as `{ type: buffer, size: len*2, kind: input }, { type: buffer, size: len, kind: output }`; the binding code will invert `len*2` to determine that `len` is the first buffer's length divided by two, and will use that to allocate the output buffer
  - Side note: this happens for `mod_precomp` -- at run-time, the length is stored deep inside the `ctx` struct, so the C code *does* know what the length is, but from the binding's perspective, we don't want to add a grammar of lengths that allows selecting fields from a struct... because this would be incredibly ad-hoc
- Support for new buffer types, denoted `buffer(uint32)` and `buffer(uint64)`
  - This requires adding a layer of abstraction via the `array_type` and `cell_size` wrappers to make sure all length computations are correct
- Support for layouts. This is the big item in this PR.
  - krml now dumps in a file called layouts.json the layout of Low* structs as compiled to wasm.
  - layouts.json contains size, field offsets, types... it's essentially a run-time type description
  - krml has been extended with support for heap allocations in a very degenerate fashion: allocations are accumulated at the end of the memory; each heap block starts with a header that describes its size; nothing is ever freed (PR forthcoming)
  - for those functions that we know return a heap-allocated value that
    - contains no cycles
    - contains no sharing
    - and where the entirety of the heap allocations performed by the function are reachable via the return value (no global state),
    then we use `heapReadLayout` to traverse the heap object graph returned by the function and map it to a "native" JS object
  - then, if a function needs to *receive* a struct, we use `stackWriteLayout` to lay out this object on the stack

For instance, the following layout, as dumped by krml:

```
Hacl_Bignum_MontArithmetic_bn_mont_ctx_u64
{
  size: 20,
  fields: [
    [ 'len', [Array] ],
    [ 'n', [Array] ],
    [ 'mu', [Array] ],
    [ 'r2', [Array] ]
  ]
}
```

when passed to `heapReadLayout`, returns the following JS object:

```
{
  len: 1,
  n: BigUint64Array(1) [ 67n ],
  mu: 1101298153654301589n,
  r2: BigUint64Array(1) [ 21n ]
}
```

in the sample test I have added to test_api.js

There are extra files because I've put too much stuff under version control. This will get cleaned up later.


msprotz:
I forgot to mention that this also makes `api.js` loadable from a browser context without having to resort to packaging npm modules with webpack.

bindings/js/api.js
```diff
@@ -1,76 +1,150 @@
...
+  if (arg.indexOf("-") >= 0)
+    return parseOp(arg, "-");
+  if (arg.indexOf("*") >= 0)
+    return parseOp(arg, "*");
+  if (arg.indexOf("/") >= 0)
+    return parseOp(arg, "/");
+  return [ arg ];
+};
+
+var evalSize = function(arg, args_int32s) {
```
denismerigoux, line 49: So at this point you're implementing a mini-compiler in Javascript. i was hoping we could avoid that but it seems inevitable...

bindings/js/api.js
```diff
@@ -1,76 +1,150 @@
...
+      let x = parseInt(const_);
+      if (args_int32s[var_] % x != 0)
+        throw new Error("Argument whose length is "+arg+" is not a multiple of "+x);
+      return args_int32s[var_] / x;
+    default:
+      return args_int32s[var_];
+  }
+};
+
+var invertSize = function(arg, total) {
```
denismerigoux, line 68: Given `total = len<op>size`, this returns `len` by inverting the formula?

bindings/js/api.js
```diff
@@ -1,76 +1,150 @@
...
       });
       if (func_obj.return === undefined) {
         throw Error("please provide a 'return' field for " + obj_name + " in api.json");
       }
-    });
-  });
+    };
+  };
 };
-validateJSON(api_json);
```
denismerigoux, line 73: I suspect you moved this validation call somewhere else ?

bindings/js/api.js
```diff
@@ -144,114 +216,334 @@ var HaclWasm = (function() {
...
-  var parseSize = function(arg, var_lengths) {
-    if (arg.indexOf("+") >= 0) {
-      return evalSizeWithOp(arg, "+", var_lengths);
-    } else if (arg.indexOf("-") >= 0) {
-      return evalSizeWithOp(arg, "-", var_lengths);
-    } else {
-      return var_lengths[arg];
+  var heapReadInt = (typ, ptr) => {
+    switch (typ) {
+      case "A8":
```
denismerigoux, line 305: Why `A8` for the name and not `U8` or `I8`?

bindings/js/api.js
```diff
@@ -144,114 +216,334 @@ var HaclWasm = (function() {
...
+          heapWriteBlockFast(data[1][0], dst, v);
+          break;
+        }
+        // pass-through
+      default:
+        throw new Error("Not implemented: "+tag);
+    }
+  };
+
+  var heapReadLayout = (layout, ptr) => {
```
denismerigoux, line 409: So this reads the JSON emitted in `layout.json` by Karamel right ?

dist/wasm/README
```diff
@@ -0,0 +1,24 @@
+How to run the WASM output of KreMLin?
+
+# With v8, from the console
```
denismerigoux, line 3: Why d8 and not node.js ?

dist/wasm/loader.js
```diff
@@ -61,20 +61,23 @@ function p32(n) {
   return p8((n >>> 24) & 255) + p8((n >>> 16) & 255) + p8((n >>> 8) & 255) + p8(n & 255);
 }
 
-function dump(mem, size) {
```
denismerigoux, line 64: This comes from karamel and https://github.com/FStarLang/karamel/pull/251 right?

denismerigoux (approved):

bindings/js/api.js
msprotz: yes it's kind of annoying, but the only other thing I can think of is using `eval` and that's even more unpleasant...

bindings/js/api.js
msprotz: yes, I moved it because it's now a promise being resolved here: https://github.com/project-everest/hacl-star/pull/547/files#diff-ef3f015ccaf90df83df03c4a14fb05bde98677501f5524461f7684d415bcebb1R619

in the case of a web-based setting, you use `fetch` to retrieve the JSON off the network, so it's not immediately available

dist/wasm/README
msprotz: I guess this is kind of outdated, node is what I'm using now... years ago node was on an old version of v8 and didn't have wasm enabled by default... I'll fix it

dist/wasm/loader.js
msprotz: correct, loader.js gets copied and put under version control in the repository

dist/wasm/README
msprotz: fixed on the kremlin side (this README is auto-generated by kremlin and I'm not even sure we normally keep it in the regular build)

bindings/js/api.js
msprotz: fixed

msprotz:
Notable new features over the past few days:
- allow dependent lengths, for instance here: https://github.com/project-everest/hacl-star/blob/protz_wasm_bignum/bindings/js/api.json#L3059
- full bindings for HACL's Bignum/BignumMontgomery64 CC @polubelova for info
- caught one missing multiplication by cell size that was being hidden due to 8-byte stack alignment for buffer(uint64) and my test bignums being only one limb in length

msprotz:
Spurious failures on drone (error: RPC failed; curl 56 LibreSSL SSL_read: SSL_ERROR_SYSCALL, errno 104) but otherwise green, as in [05a989f](https://github.com/project-everest/hacl-star/pull/547/commits/05a989f570bda61b354946651c4360c572429830)

The policy for the HACL* repo is to have two reviews per PR, one Microsoft, one INRIA. I'm not sure anyone at Microsoft has expertise to review this (@tahina-pro? @aseemr? @s-zanella?), so unless someone steps up, I'll bypass the restriction and merge. Thanks.

(And I'll submit followup PRs for the rest of the changes I have lined up.)

bindings/js/api.js
```diff
@@ -1,76 +1,207 @@
...
+
+var parseOp = (arg, op) => {
+  let [ var_, const_ ] = arg.split(op);
+  return [ "Op", var_, op, const_ ];
+};
+
+var parseApp = (arg) => {
+  let i = arg.indexOf("(");
+  let [ m, f ] = arg.substr(0, i).split(".");
+  let x = arg.substr(i+1, arg.length - 1 - i - 1);
```
W95Psp, line 51: Why isn't `parseInt` called here once instead of multiple times later in `evalSize` and `invertSize`?
Also, shouldn't `x` be checked to be a number (and throw an error if that's not the case)?

bindings/js/api.js
```diff
@@ -1,76 +1,207 @@
...
+
+var parseSize = (arg) => {
+  if (arg.indexOf("+") >= 0)
+    return parseOp(arg, "+");
+  if (arg.indexOf("-") >= 0)
+    return parseOp(arg, "-");
+  if (arg.indexOf("*") >= 0)
+    return parseOp(arg, "*");
+  if (arg.indexOf("/") >= 0)
+    return parseOp(arg, "/");
```
W95Psp, line 63: I think it would be more clear to use `arg.contains(op)` instead of `arg.indexOf(op) >= 0` here

bindings/js/api.js
```diff
@@ -79,31 +210,30 @@ var HaclWasm = (function() {
...
           return {
             buf: b,
             name: shell.my_modules[i]
           };
         }));
       }).then(function(scope) {
-          Module = scope;
-          isInitialized = true;
+        Module = scope;
+        isInitialized = true;
```
W95Psp, line 235: Maybe using `await/async` would be more clear here, something like:
```js
var loadWasm = async () => {
	if(!isInitialized){
	    Module = await loader.link(
		my_imports,
		(await modules_promise()).map((buf,i) => ({
		    buf,
		    name: shell.my_modules[i]
		}))
	    );
	    isInitialized = true;
	}
    };
```

bindings/js/api.js
```diff
@@ -144,114 +276,346 @@ var HaclWasm = (function() {
...
+      default:
+        throw new Error("Unknown array type: "+type);
+    }
+  }
+
+  var check_array_type = function(type, candidate, length, name) {
     if ((typeof(candidate) !== "object") ||
       (candidate.length !== length) ||
-      (candidate.constructor !== Uint8Array)
+      (candidate.constructor !== array_type(type))
```
W95Psp, line 308: `typeof candidate !== "object" || candidate.constructor !== array_type(type)` could be simplified into `!(candiate instanceof array_type(type))` here

bindings/js/api.js
```diff
@@ -1,76 +1,207 @@
...
-            throw Error("incorrect 'size' field value (" + arg.size + ")for argument #" + i + " of " + obj_name + " in api.json");
-          }
+      func_obj.args.forEach(function(arg, i) {
+        if (arg.type.startsWith("buffer") && typeof arg.size === "string" && arg.kind === "output" &&
+          !length_args_available.includes(arg.size) &&
+          !length_args_available.includes(arg.size.substring(0, arg.size.indexOf("+"))) &&
+          !length_args_available.includes(arg.size.substring(0, arg.size.indexOf("-"))) &&
+          !length_args_available.includes(arg.size.substring(0, arg.size.indexOf("/"))) &&
+          !length_args_available.includes(arg.size.substring(0, arg.size.indexOf("*"))) &&
+          !(arg.size.indexOf("(") >= 0)
```
W95Psp, line 192: Why is `arg.size` parsed manually like this here, why not call `parseSize` directly?

bindings/js/api.js
```diff
@@ -144,114 +276,346 @@ var HaclWasm = (function() {
...
+      case "buffer":
+        return 1;
+      case "buffer(uint32)":
+        return 4;
+      case "buffer(uint64)":
+        return 8;
+      default:
+        throw new Error("Unknown array type: "+type);
+    }
+  }
```
W95Psp, line 303: Why not `let cell_size = type => array_type(type).BYTES_PER_ELEMENT`?

bindings/js/api.js
```diff
@@ -260,67 +624,101 @@ var HaclWasm = (function() {
...
+    if ("kind" in proto.return && proto.return.kind === "layout") {
+      // Heap-allocated value
+      if (proto.return.type.startsWith("buffer")) {
+        let r = heapReadBuffer(proto.return.type, call_return);
+        memory[Module.Karamel.mem.buffer.byteLength/4-1] = 0;
+        return r;
+      } else {
+        let r = heapReadLayout(proto.return.type, call_return);
+        memory[Module.Karamel.mem.buffer.byteLength/4-1] = 0;
+        return r;
```
W95Psp, line 666: Refactoring like below could improve readability maybe
```js
let read = proto.return.type.startsWith("buffer") ? heapReadBuffer : heapReadLayout;
let r = read(proto.return.type, call_return);
memory[Module.Karamel.mem.buffer.byteLength/4-1] = 0;
return r;
```

bindings/js/api.js
msprotz: see https://github.com/project-everest/hacl-star/blob/protz_wasm_bignum/bindings/js/api.json#L3059 for an example of the `App` case... it might be `f(x)` where both `f` and `x` are variables, not concrete numbers

bindings/js/api.js
msprotz: yeah this is just a leftover of how things were done before... I think I have a TODO for this

I can fix it

bindings/js/api.js
msprotz: but `contains` is not a method of the String prototype...?

W95Psp: Oops, sorry, I always mistake, it's `includes`, not `contains`...

W95Psp (approved):

franziskuskiefer:
@msprotz Looks like this broke the build on arm. I'm seeing a lot of the following errors
```
...
  0x   0: Zero-sized arrays are not supported in C and in WASM either. See WasmSupport.fst
  0x  51: Overflow in memzero; see WasmSupport.fst
...
  0x 86a: Abort: sha256_update was meant to be hand-written and provided at link-time, but contains an I64 and therefore cannot be called from WASM.
```

msprotz:
Do you mean the build on a Mac M1?

msprotz:
If you have a complete build log somewhere to send me (even privately) that'd be super great. Thanks.

PR #549, opened 2022-04-18
Attempt to fix nightly

msprotz:
Apparently the hints are no longer valid for `Hacl.Impl.P256.PointDouble.fst`. Without hints, this file generates a z3 query that alone takes nearly 40 minutes on my machine. This results in a nightly build that simply times out (we have a 1h limit). With hints, the file takes 1min to verify.

This seems like a case of a proof that could be fleshed out in a little more detail, but in the meanwhile, I hope this fixes nightly. I'll manually kick a nightly build after this is merged.

### DIFF
--- a/hints/Hacl.Impl.P256.PointDouble.fst.hints
+++ b/hints/Hacl.Impl.P256.PointDouble.fst.hints
@@ -13,7 +13,7 @@
         "refinement_interpretation_Tm_refine_0c4926e33a4dc1c024a6bb2210141ee1"
       ],
       0,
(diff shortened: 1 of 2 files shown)


polubelova (approved):
Thanks!

aseemr (approved):
Thanks Jonathan! Would it be worth making the options for nightly same as that of the regular CI build, given that those are the ones that we usually test as a check-in gate?

msprotz:
Yeah I wonder too. Let me start a discussion on Slack.

PR #552, opened 2022-05-12
Allow high-level HACL-WASM API to only load a subset of the WASM files

msprotz:
This leverages the latest changes in krml.

Note that to test these changes, one has to first do `cp ../../dist/wasm/* .` then do `node test2.js` and/or `http-server .` to navigate to test.html


msprotz:
Would love a review from @w95psp... thanks!

msprotz:
For some reason I can't seem to be able to request a review from you... weird

PR #554, opened 2022-05-12
 Allow high-level HACL-WASM API to only load a subset of the WASM files

msprotz:
This leverages the latest changes in krml.

Note that to test these changes, one has to first do `cp ../../dist/wasm/* .` then do `node test2.js` and/or `http-server .` to navigate to test.html

**Note:** for some reason, `git push` succeeded when I didn't intend for this branch to be merged? I think what happened is that now that we've lost CI, GitHub thought everything was green and didn't prevent me from pushing. I tried to prevent that from happening again.


msprotz:
Would love a review from @W95Psp... thanks!

msprotz:
Note that this PR also reduces the size of many WASM files because there was an unwanted sharing of a hash table that happened in krml (see https://github.com/FStarLang/karamel/pull/261#discussion_r862217073). So, this is a nice side-effect. The total size for all of HACL*/EverCrypt in WASM is now 1.1M

bindings/js/api.js
```diff
@@ -7,18 +7,31 @@ if (typeof module !== 'undefined') {
...
   var shell = this;
   var api_promise = fetch("api.json").then(r => r.json());
   var layouts_promise = fetch("layouts.json").then(r => r.json());
-  var modules_promise = Promise.all(my_modules.map(m => fetch(m + ".wasm")))
-    .then(response => Promise.all(response.map(r => r.arrayBuffer())));
+}
+
+// We now allow the user to pass a custom list of modules if they want to do
+// their own, more lightweight packaging.
+function getModulesPromise(modules) {
```
W95Psp, line 20: I think using a default parameter would be more clear here, i.e. having `function getModulePromise(modules = shell.my_modules)` and thus simplifying `modules || my_modules` to just `modules`

bindings/js/api.js
```diff
@@ -7,18 +7,31 @@ if (typeof module !== 'undefined') {
...
+        buf: new Uint8Array(source),
+        name: m
+      });
+    }));
+  else
+    return Promise.all((modules || my_modules)
+      .map(name =>
+        fetch(name + ".wasm")
+        .then(r => r.arrayBuffer())
+        .then(buf => ({ buf, name }))));
```
W95Psp, line 34: Why not factorize the code a bit by extracting the "in browser / in module" behavior, for instance like that:
```js
function getModulesPromise(modules = shell.my_modules) {
    const readModule = async path =>
        typeof module !== 'undefined'
            ? new Uint8Array(await fs.readFile(path.resolve(__dirname, './' + path)))
            : (await fetch(path)).arrayBuffer();
    return Promise.all(modules.map(async name =>
        ({ buf: await readModule(name + ".wasm"), name })
    ));
}
```

bindings/js/test.html
```diff
@@ -0,0 +1,90 @@
...
+        // API demonstration.
+
+        // Test helpers
+        // ------------
+
+        function buf2hex(buffer) {
+          return Array.prototype.map.call(new Uint8Array(buffer), function(x) {
+            return ('00' + x.toString(16)).slice(-2);
+          }).join('');
+        }
```
W95Psp, line 30: I think using spread syntax (`[...buffer].map(function (x) ...)`) would be more readable here, instead of `Array.prototype.map.call(...)`.

Also, why not using fat arrows & template strings? For instance:
```js
const buf2hex = buffer => [...buffer].map(x => `00${x.toString(16)}`.slice(-2)).join('')
```

bindings/js/api.js
```diff
@@ -7,18 +7,31 @@ if (typeof module !== 'undefined') {
...
+  if (typeof module !== 'undefined')
+    return Promise.resolve((modules || shell.my_modules).map(m => {
+      var source = fs.readFileSync(path.resolve(__dirname, './' + m + ".wasm"));
+      return ({
+        buf: new Uint8Array(source),
+        name: m
+      });
+    }));
+  else
+    return Promise.all((modules || my_modules)
```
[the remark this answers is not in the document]
msprotz: Well, it's tricky. If we load via node, we do `var shell = require("./shell.js")` which exports `my_modules`. If we load in the browser, we just do `<script src="shell.js">` which imports `my_modules` in the global scope (`window`).

Maybe `shell.js` could also define a variable called `shell` and set its `my_modules` property and then the code would be uniform...?

bindings/js/test.html
msprotz: @W95Psp I think you are just pointing out my advanced age and the fact that I have not caught up with the latest JS syntax innovations. Thanks for dragging me into the future ;-)

bindings/js/api.js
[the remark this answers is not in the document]
msprotz: Actually I already define `var shell = this` in the browser case so this should work too and allow the default argument... thanks!

bindings/js/test.html
msprotz: I took into account your other comments but this one doesn't work because ArrayBuffer is not iterable and so spread syntax does not work for it. I'm marking this as resolved, but please feel free to suggest something else. Thanks.

W95Psp: Ah, yeah, there is this thing of views over buffers, I did not read carefuly enough the original code. I think ``const buf2hex = buffer => [...new Uint8Array(buffer)].map(x => `00${x.toString(16)}`.slice(-2)).join('')`` would work then!

W95Psp (approved):

bindings/js/test.html
msprotz: That worked, thanks!

bindings/js/test2.js
```diff
@@ -1,29 +1,32 @@
...
 
-function buf2hex(buffer) {
-  return Array.prototype.map.call(new Uint8Array(buffer), function(x) {
-    return ('00' + x.toString(16)).slice(-2);
-  }).join('');
-}
+// Test helpers
+// ------------
+
+const buf2hex = buffer => [...new Uint8Array(buffer)].map(x => `00${x.toString(16)}`.slice(-2)).join('');
```
denismerigoux, line 13: Shouldn't this be provided by the Kremlin modules instead of re-implementing it in each test ?

denismerigoux (approved):

bindings/js/test2.js
msprotz: yes, I'll do that in the next PR

msprotz:
@W95Psp your review expired, any chance you can renew it? thanks!

W95Psp (approved):

PR #555, opened 2022-05-20
Precise header inclusion

msprotz:
This PR removes the blanket inclusion of `libintvector.h` from every single file in HACL*; instead, we now:
- include libintvector.h in the C file for well-abstracted algorithms that use libintvector.h internally, but do not rely on vector types in their public header (e.g. chacha20_256)
- include libintvector.h in the H file for algorithms whose public API relies on vector types (e.g. poly1305_256)

I also removed two distributions that are obsolete to simplify the build and Makefile.


msprotz:
Note that due to transitive header inclusion, we don't need to include libintvector.h in files that include a vectorized algorithm but are not using vectorized code themselves (e.g., streaming_poly1305_256).

msprotz:
Screen Shot 2022-05-20 at 12 45 34 PM

franziskuskiefer (approved):
Thanks!
Unfortunately there are still too many other issues to easily test this with gcc4.8 and see how far it gets us (as you mentioned there are still some issues). But its a nice cleanup in any case so let's get it in.

polubelova (approved):
looks good to me, thanks!

PR #557, opened 2022-05-23
Nix flake

W95Psp:
Hi,
This PR adds support for building Everest with Nix (on which we've been working with @pnmadelaine), via a Nix expression that lives in https://github.com/project-everest/everest-nix. This is useful for the (upcoming) CI.

Note that dependencies to specific branches of F* or KaRaMeL can be specified by replacing `master` by any other git reference in the field `fstar-src.url` and `karamel-src.url`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ dist/wasm/*.wast
 doc/_build
 config.h
 *.smt2
+flake.lock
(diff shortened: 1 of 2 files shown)


karthikbhargavan (approved):
Thanks. This CI is still experimental, but we have high hopes that we can transition to it.

msprotz (approved):

PR #558, opened 2022-05-24
Add support for Blake2b/Blake2s optimized versions in EverCrypt.Hash

protz:
There remains one item to be done, namely implement the copy function to fatally fail if someone started hashing with blake2 *then* called autoconfig only to discover that the vectorized version is enabled.

### DIFF
--- a/providers/evercrypt/fst/EverCrypt.Hash.fst
+++ b/providers/evercrypt/fst/EverCrypt.Hash.fst
@@ -37,20 +37,52 @@ let string_of_alg =
 let uint32_p = B.buffer uint_32
 let uint64_p = B.buffer uint_64
 
-inline_for_extraction noextract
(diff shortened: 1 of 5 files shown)


providers/evercrypt/fst/EverCrypt.Hash.fst
```diff
@@ -549,7 +692,7 @@ let copy #a s_src s_dst =
       let s_dst: state SHA2_512 = s_dst in
       let p_dst = SHA2_512_s?.p !*s_dst in
       B.blit p_src 0ul p_dst 0ul 8ul
-  | Blake2S_s p_src ->
+  (*| Blake2S_s p_src ->
```
protz, line 695: this is the part that I need to fix -- @karthikbhargavan I think you said there was no hope of copying a blake2b/32 state into a blake2b/256 state?

PR #560, opened 2022-05-26
github action to update hints and dist

pnmadelaine:
This PR adds a nightly github action to retrieve binary files generated by the [Hydra CI](https://everest-ci.paris.inria.fr).
This github action doesn't trigger a new build, it simply downloads build products from the latest succesful build on Hydra.

### DIFF
--- a/.github/workflows/buildproducts.yml
+++ b/.github/workflows/buildproducts.yml
@@ -5,7 +5,7 @@ on:
     - cron: '0 2 * * *'
 
 jobs:
-  update:
+  hints-and-dist:
     runs-on: ubuntu-latest
     steps:
       - name: checkout
--- a/.github/workflows/buildproducts.yml
+++ b/.github/workflows/buildproducts.yml
@@ -1,0 +1,26 @@
+name: update hints and dist
+
+on:
+  schedule:
+    - cron: '0 2 * * *'
+  workflow_dispatch:
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+      - name: download new files
+        run: |
+          git rm -r hints dist/*/*
+          export url=https://everest-ci.paris.inria.fr/job/everest/branch-master/hacl-build-products.x86_64-linux/latest/download
+          diff <(curl -L ${url}/3) <(git rev-parse HEAD)
+          curl -L ${url}/1 | tar -xv
+          curl -L ${url}/2 | tar -xv
+      - name: commit changes
+        uses: EndBug/add-and-commit@v9
+        with:
+          author_name: Project Everest
+          author_email: project@everest
+          message: "[CI] update hints and dist"
--- a/.github/workflows/buildproducts.yml
+++ b/.github/workflows/buildproducts.yml
@@ -18,8 +18,8 @@ jobs:
           baseUrl="https://everest-ci.paris.inria.fr"
           latestFinishedEval=$(curl -sLH 'Content-Type: application/json' "$baseUrl/jobset/hacl-star/branch-master/latest-eval")
 
-          rev=$(echo "$latestFinishedEval" | jq '.flake | split("/") | last')
-          id=$(echo "$latestFinishedEval" | jq '.id')
+          rev=$(echo "$latestFinishedEval" | jq -r '.flake | split("/") | last')
+          id=$(echo "$latestFinishedEval" | jq -r '.id')
 
           [[ "$rev" == "$(git rev-parse HEAD)" ]] || {
               echo "The latest evaluation on the CI doesn't correspond to the latest commit."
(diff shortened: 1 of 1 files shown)


msprotz:
@pnmadelaine this seems quite great, I am in favor of adding this

pnmadelaine:
@msprotz I tested it on my local fork of hacl and the extraction is failing, I'll try to debug this tomorrow.

pnmadelaine:
There seems to be a bug in hydra when distributing several files from the same build, thus I split the build of hints and dist to work around the issue. This github action is now working as expected.

W95Psp (approved):

msprotz:
HACL* is built twice, just to produce different tarballs?

pnmadelaine:
@msprotz I meant to say *derivation* instead of *build*, a derivation is a unit of build in Nix. hacl is built only once its own derivation, then there are two derivations that depend on hacl to build dist and hints separately instead of one derivation building the two.

pnmadelaine:
I just had a talk with @R1kM, the conclusion of which is that it could be desirable that the github action checked that the downloaded hints and dist do match the current head of master. otherwise we can imagine some situations when the github action commits stale hints or dist.

R1kM:
Regarding @pnmadelaine's comment, the case I have in mind is if the CI refreshes dist according to master, but a PR is merged in the meantime adding/changing some files. This is likely to break the invariant that dist/ always corresponds to the F* code in HACL.

pnmadelaine:
I did a pretty big update to the setup I propose, these are the main steps:
1. [project-everest/everest-nix](https://https://github.com/project-everest/everest-nix) has a github action that runs every day at midnight to update and lock the lastest versions of fstar, karamel and hacl. this potentially triggers a build in the CI.
2. If the build is succesful, he [Hydra CI](https://everest-ci.paris.inria.fr/project/everest) builds a derivation containing two tarballs (hints and dist) and a file with the hacl revision thery are built from
3. everey day, two hours later, this github action downloads these files, compares HEAD against the published revision and if they match it merges hints and dist into master

The main reason why this is a bit convoluted, is because the Hydra API only offers a way to access the latest succesful build, and to my knowledge there is no easy way to access the corresponding git revision, hence the check in case the latest succesful build is stale because the latest build failed.

msprotz:
Can you do something like `git rev-parse HEAD > REVISION` as part of the build to be able to easily compare?

Just as a data point: the convoluted scenario where we have a conflict never happened in our case, or if it did, then the application of the hints failed because the diff was stale, and this means that someone just merged a recent PR that was green, and that the hint regeneration will work next time.

.github/workflows/buildproducts.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+      - name: download new files
+        run: |
+          git rm -r hints dist/*/*
+          url=https://everest-ci.paris.inria.fr/job/everest/branch-master/hacl-build-products.x86_64-linux/latest/download
+          diff <(curl -L ${url}/3) <(git rev-parse HEAD) # fails if the tarballs are stale
+          curl -L ${url}/1 | tar -xv
+          curl -L ${url}/2 | tar -xv
```
W95Psp, line 19:
```suggestion
          # Clean up hints and dists
          git rm -r hints dist/*/*

          STATUS_SUCCEEDED=0 # from https://github.com/NixOS/hydra/blob/cf9f38e43fd81f9298e3f2ff50c8a6ee0acc3af0/hydra-api.yaml#L927-L941
          baseUrl="https://everest-ci.paris.inria.fr"
          latestFinishedEval=$(curl -sLH 'Content-Type: application/json' "$baseUrl/jobset/hacl-star/branch-master/latest-eval")
          
          rev=$(echo "$latestFinishedEval" | jq '.flake | split("/") | last')
          id=$(echo "$latestFinishedEval" | jq '.id')
          
          [[ "$rev" == "$(git rev-parse HEAD)" ]] || {
              echo "The latest evaluation on the CI doesn't correspond to the latest commit."
              exit 1
          }
          
          buildUrl="$baseUrl/eval/$id/job/hacl-build-products.x86_64-linux"
          buildInfo=$(curl -sLH 'Content-Type: application/json' "$buildUrl")
          
          # Check wether the build was built correctly
          [[ "$(echo "$buildInfo" | jq '.buildstatus')" == "$STATUS_SUCCEEDED" ]] || {
              echo "The latest evaluation wasn't successful. Aborting."
              exit 2
          }
          
          installTar () {
              curl -sL "$buildUrl/download-by-type/file/$1" | tar -xv
          }
          
          installTar "hints"
          installTar "dist"
```

At the price of a bit more convoluted bash script, we can actually ask Hydra on which commit revision something was built.
The script here fetches the information about the last finished evaluation, checks wether it succeeded, and then downloads and untar `hints` and `dist`.

W95Psp (changes requested):

.github/workflows/buildproducts.yml
pnmadelaine: I do prefer this solution, even if the github action is more complex, it's more self-contained than my version and should cause less problems.

msprotz:
Looks good to me!

msprotz (approved):

W95Psp (approved):

PR #561, opened 2022-05-27
Remove usage of __has_include

msprotz:
@franziskuskiefer reported that our usage of `__has_include` was yet one more oddity that made support across compiler versions difficult.

I believe this *used* to be a an issue but with a recent change from Natalia, there is now a preprocessor guard that doesn't trigger __has_include unless we know we have it.

But simplifying this logic is good, and so I would like to remove our usage of __has_include. (There is one left for debugging purposes in libintvector.h that I believe is innocuous.)

There would need to be further changes, for instance:
- fixing the local Makefiles to copy a dummy development config.h to enable all the features
- making sure no dist relies on the absence of config.h (I know Mozilla has one now, but maybe election guard needs one...?)

Therefore, I'm leaving this as a draft PR for further discussion.

### DIFF
--- a/dist/c89-compatible/Lib_Memzero0.c
+++ b/dist/c89-compatible/Lib_Memzero0.c
@@ -1,8 +1,4 @@
-#if defined(__has_include)
-#if __has_include("config.h")
 #include "config.h"
-#endif
-#endif
 
 #ifdef _WIN32
 #include <windows.h>
--- a/dist/c89-compatible/evercrypt_targetconfig.h
+++ b/dist/c89-compatible/evercrypt_targetconfig.h
@@ -18,13 +18,7 @@
 #define TARGET_ARCHITECTURE_ID_SYSTEMZ 5
 #define TARGET_ARCHITECTURE_ID_POWERPC64 6
 
-#if defined(__has_include)
-#if __has_include("config.h")
 #include "config.h"
-#else
-#define TARGET_ARCHITECTURE TARGET_ARCHITECTURE_ID_UNKNOWN
-#endif
-#endif
 
 // Those functions are called on non-x64 platforms for which the feature detection
 // is not covered by vale's CPUID support; therefore, we hand-write in C ourselves.
--- a/dist/c89-compatible/lib_intrinsics.h
+++ b/dist/c89-compatible/lib_intrinsics.h
@@ -2,11 +2,7 @@
 
 #include <sys/types.h>
 
-#if defined(__has_include)
-#if __has_include("config.h")
 #include "config.h"
-#endif
-#endif
 
 #if defined(HACL_CAN_COMPILE_INTRINSICS)
 #if defined(_MSC_VER)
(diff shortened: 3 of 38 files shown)


msprotz:
Unsurprisingly, this breaks windows where indeed, our CI does not generate a config.h (it should!).

franziskuskiefer:
> I believe this _used_ to be a an issue but with a recent change from Natalia, there is now a preprocessor guard that doesn't trigger __has_include unless we know we have it.

`__has_include` doesn't exist in older GCCs. So any GCC 4.8 build for example will not pick up the `config.h`.

> Unsurprisingly, this breaks windows where indeed, our CI does not generate a config.h (it should!).

I'd suggest removing the CI. There's no point in trying to keep this broken build system around.

msprotz:
I think there is still some benefit to this, namely that we check that the C code generated by krml is syntactically correct for MSVC, so I'd like to keep this if we can. It's better than nothing, until we have more comprehensive Windows CI.

franziskuskiefer (approved):
Thanks!
It appears to work fine on Windows as well: https://github.com/cryspen/hacl-packages/actions/runs/2581290156

msprotz:
Thanks, I needed a couple more tweaks, hopefully this time is the right time.

msprotz:
So now `make` will error out if you don't have config.h and Makefile.config. You can still write these by hand (Mozilla does), but of course the suggestion is to run `./configure` before invoking make.

The CI scripts have been updated to reflect this.

franziskuskiefer (commented):
lgtm, but some changes in gcc-compatible look unrelated?

Frosne:
If you don't mind, I will first check that the modification does not break anything in our code :) 
I should be back with a definitive answer on Monday. In any case, I support any code simplifications.

msprotz:
Thanks, if this looks good for you on Monday, please go ahead and merge (I'll be off on vacation).

Frosne:
as promised, I am back. 

It seems that everything works, but I would like to take more time to look at it. 
I will be away as well, until 13th of July. As soon as I am back, it will be the first thing I will address :)

msprotz:
@Frosne can you confirm this looks good for you? I think we've pretty much committed to always having config.h now (even if it's hand-written), so __has_include can go I think.

Frosne:
IIUC, we (dist/mozilla) also should have a config.h file? Could we have it empty?

Frosne:
I would prefer not to have any predefined values there, because we do it in our build system.

msprotz:
Ok. Until now there *was* a config.h under version control for you guys, in dist/mozilla.

Are you saying you're ok to remove dist/mozilla/config.h?

msprotz:
Sorry, here: https://github.com/hacl-star/hacl-star/blob/main/dist/config.mozilla.h

Frosne:
We do not use it, so yes, I think it is ok :)

PR #562, opened 2022-05-31
OCaml API: revise upper bounds checks which use pow

victor-dumitrescu:
While profiling the memory usage of a piece of code which uses the incremental EverCrypt hash interface, @polubelova noticed that the runtime checks in the OCaml API, specifically those comparing the size of a buffer to a large power of 2, result in excessive allocations.

~~The original intention was to encode the F* preconditions as runtime checks, but I think in this case the cost outweighs the benefits, especially since the large upper bound being tested here are unlikely to be reached in many real-world use cases. I therefore propose to remove them.~~

As suggested in this PR, I've hoisted computations of the relevant powers of 2 and replaced `Z.pow` with `Z.shift_left`. I've also reduced the checked bounds to be no larger than the types that represent them. This solves the excessive allocations problem.

### DIFF
--- a/bindings/ocaml/EverCrypt.ml
+++ b/bindings/ocaml/EverCrypt.ml
@@ -138,7 +138,7 @@ module Hash = struct
   open EverCrypt_Hash
   module Noalloc = struct
     let hash ~alg ~msg ~digest =
-      check_max_input_len alg (C.size msg);
+      check_max_buffer_len (C.size msg);
       assert (C.size digest = digest_len alg);
       assert (C.disjoint digest msg);
       everCrypt_Hash_hash (alg_definition alg) (C.ctypes_buf digest) (C.ctypes_buf msg) (C.size_uint32 msg)
@@ -155,10 +155,10 @@ module Hash = struct
     let incr_len = ref Z.zero in
     Gc.finalise everCrypt_Hash_Incremental_free st;
     (alg, incr_len, st)
-  let update ~st:(alg, incr_len, t) ~msg =
-    check_max_input_len alg (C.size msg);
+  let update ~st:(_alg, incr_len, t) ~msg =
+    check_max_buffer_len (C.size msg);
     incr_len := Z.add !incr_len (Z.of_int (C.size msg));
-    assert (Z.lt !incr_len (max_input_len alg));
+    assert (Z.lt !incr_len max_uint64);
     everCrypt_Hash_Incremental_update t (C.ctypes_buf msg) (C.size_uint32 msg)
   let finish ~st:(alg, incr_len, t) =
     let digest = C.make (digest_len alg) in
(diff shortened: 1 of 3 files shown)


polubelova (approved):
Thanks!

franziskuskiefer:
I have to look at the actual code. But shifting could indeed be used to check the length without allocating memory. It would be nice if we could keep the checks.

msprotz:
@polubelova defensive APIs in C is another matter, and potentially a fairly involved piece of work (but I agree it'd be nice to have).

Regarding the specific issue of a slow left-shift: there are multiple things you can do here without disabling safety checks.
- Stop recomputing `pow2 64` on every call; can you define `let pow2_64 = pow2 64` at the top-level and use that?
- More efficient `pow2` using `Z.shift_left`

If none of these work according to your benchmarks, then we can maybe contemplate having an "insane" mode where some checks are disabled but not by default to keep the API safe in all cases.

Can you investigate a little bit more this pow2 issue? Unfortunately I don't have access to you benchmarks so I don't know how to reproduce. Thanks!

victor-dumitrescu:
Makes sense, I can try rewriting the checks with `shift_left`. Marina also pointed out that some of these checks aren't doing anything in practice (see comment above) and could be rewritten with lower bounds than in the specs.

msprotz (approved):
If this gives you the performance improvements you need, I am happy with this 👍  thanks!

franziskuskiefer (approved):
lgtm, thanks!

victor-dumitrescu:
I was running some benchmarks and thought it might be useful to document the improvement that we got from this fix, specifically for Blake2b. Lower memory usage overall + for very small digests it makes a noticeable running time difference too:

- before:
```
│ Hacl/Blake2b_256.hash/0              │     433.39ns │  37.00w │ 0.14e-3 │
│ Hacl/Blake2b_256.hash/100            │     439.87ns │  37.00w │ 0.14e-3 │
│ Hacl/Blake2b_256.hash/1000           │   1_651.76ns │  37.00w │ 0.14e-3 │
│ Hacl/Blake2b_256.hash/10000          │  13_904.17ns │  37.00w │ 0.14e-3 │
│ Hacl/Blake2b_256.hash/100000         │ 136_611.97ns │  37.00w │ 0.14e-3 │
│ Hacl/Blake2b_32.hash/0               │     597.80ns │  37.00w │ 0.14e-3 │
│ Hacl/Blake2b_32.hash/100             │     597.11ns │  37.00w │ 0.14e-3 │
│ Hacl/Blake2b_32.hash/1000            │   2_903.03ns │  37.00w │ 0.14e-3 │
│ Hacl/Blake2b_32.hash/10000           │  25_962.63ns │  37.00w │ 0.14e-3 │
│ Hacl/Blake2b_32.hash/100000          │ 252_820.96ns │  37.00w │ 0.13e-3 │
```

- after:
```
│ Hacl/Blake2b_256.hash/0              │     303.42ns │  31.00w │ 0.12e-3 │
│ Hacl/Blake2b_256.hash/100            │     307.70ns │  31.00w │ 0.12e-3 │
│ Hacl/Blake2b_256.hash/1000           │   1_565.54ns │  31.00w │ 0.12e-3 │
│ Hacl/Blake2b_256.hash/10000          │  14_128.73ns │  31.00w │ 0.12e-3 │
│ Hacl/Blake2b_256.hash/100000         │ 138_708.93ns │  31.00w │ 0.12e-3 │
│ Hacl/Blake2b_32.hash/0               │     461.02ns │  31.00w │ 0.12e-3 │
│ Hacl/Blake2b_32.hash/100             │     465.45ns │  31.00w │ 0.12e-3 │
│ Hacl/Blake2b_32.hash/1000            │   2_747.43ns │  31.00w │ 0.12e-3 │
│ Hacl/Blake2b_32.hash/10000           │  25_470.37ns │  31.00w │ 0.12e-3 │
│ Hacl/Blake2b_32.hash/100000          │ 250_708.27ns │  31.00w │ 0.10e-3 │
```

PR #563, opened 2022-06-01
Modification of the multiplication algorithm used for P256

Frosne:
#### General
This PR replaces the multiplication and squaring primitives used for the P256 curve with the Bignum-based ones. 

#### Motivation
The multiplication and squaring functions were using long calc calculations that were decreasing the verification speed of the files. Having that the bignum multiplication and squaring primitives compute the same thing, we decided to replace these two functions. We do not expect a big decrease in performance. 

The mul function was replaced by bn_mul, the square function was replaced by bn_sqr function. Finally, the PR introduces the lemmas required for the interconnection of the existing code and the bignum code.


msprotz:
@Frosne any updates on @karthikbhargavan's question? Thanks!

Frosne:
Yea, I did put it in the first message. 

The main motivation of the modifications are to simplify the code by reusing the functionality already implemented for the generic bignum. Here we were aiming in replacing the functions mul/sq with the same functions from the bignum library. 

As a big plus, these functions in the P256 library was causing much longer verification time than the ones in Bignum library, so we have double win :)

karthikbhargavan:
Thanks, these reasons make sense. I am all for reducing redundancy and verification times.

Can we measure the performance impact of this change? 

Are there other functions from Bignum we could also use, in addition to mul and sq?

Frosne:
I think Marina and Victor have already done that! See here: (https://github.com/project-everest/hacl-star/pull/543)

Unfortunately, I don't think that anything else could be replaced..

msprotz:
@Frosne thanks for adding the motivations in the initial message. FYI, I don't think anyone gets a notification if the original message in the thread gets modified.

If @polubelova @victor-dumitrescu are ok with this PR, then I'll happily merge it (after you fix conflicts). Thanks!

@polubelova @victor-dumitrescu I'm requesting a review from you both to make sure we're on the same page for P256. Thanks!

victor-dumitrescu:
Hi @Frosne, could you please include the C code in the PR as well? It would be easier for me to review and benchmark. Thanks!

Frosne:
Sure :)

Frosne:
Regenerated the */*P256.{c, h} files

polubelova:
@Frosne: did you try to apply [this patch](https://github.com/project-everest/hacl-star/commit/6b8c0137b104fcea0f9fc09b6280e7a2a44d65dd) from @sonmarcho to unroll loops for bignum multiplication and squaring in P256?

Frosne:
Nope, 
the C code reflects only the modifications done by this particular PR :)

msprotz (approved):
The Changes.md file is really not super informative. How about deleting it, or writing a meaningful entry, such as:
- replaced hand-written custom multiplication with generic primitives from the existing bignum directory

polubelova (approved):

victor-dumitrescu:
I benchmarked the code now in `main` against the latest OCaml release (0.5.0) and the P-256 functions in `main` seem to be 5-10% slower.

franziskuskiefer:
I can confirm that the P256 performance on x64 regressed significantly.

New code on main
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
P256_SHA256_ECDSA_Sign       170069 ns       170073 ns         4076
P256_SHA256_ECDSA_Verify     169703 ns       169707 ns         4099
P256_ECDH                    170106 ns       170110 ns         4129
```

Old code
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
P256_SHA256_ECDSA_Sign       147520 ns       147524 ns         4752
P256_SHA256_ECDSA_Verify     142291 ns       142293 ns         4893
P256_ECDH                    143269 ns       143272 ns         4882
```

Frosne:
You would like to revert?

franziskuskiefer:
With #614 a lot of the P256 code will change again. I guess the question is whether this regression is acceptable until that's moving. I'd lean towards reverting but the @hacl-star/maintainers will have to decide what to do.

karthikbhargavan:
This is unexpected since our previous benchmarks indicated there should be a *speedup*. Let us try to figure out why this slowdown and whether there is an easy fix, and if not we will have to revert.

victor-dumitrescu:
Perhaps a build issue then? Was the code that you benchmarked built differently than `dist/gcc-compatible`?

polubelova:
It seems that the performance regression comes from removing `inline` for `Hacl_Bignum_Addition_bn_add_eq_len_u64` (cf. PR #620). I'll submit a follow-up PR with the fix.

PR #564, opened 2022-06-02
Replace mono with dotnet

pnmadelaine:


tahina-pro:
Thank you PNM and Aymeric! I'm now starting an Everest branch with this PR and FStarLang/FStar#2513

R1kM:
Quick summary of the changes: https://github.com/FStarLang/FStar/pull/2513 changes the F* ecosystem to use .NET instead of mono for the F# examples. This is a breaking change for HACL* since the Vale tool, implemented in F#, currently runs using mono.

This PR changes the top-level Makefile and the SConstruct file to replace calls to mono by calls to dotnet. This was tested with .NET 6.0. While doing this change, we initially observed errors when trying to call Vale. I'm no expert about the .NET ecosystem so I might be completely misunderstanding, but messages on StackOverflow seemed to indicate that this was because dotnet does not include by default a runtime system while mono did. The solution to this issue is apparently to include a trivial {projectname}.runtimeconfig.json file in the same folder as a {projectname}.exe F# executable. We added such a trivial config file in the repo, and tweaked the get_vale script to copy-paste it for the two Vale executables (Vale and ImportFStarTypes). There might be a better way to do this, for instance by including F# project files directly in the Vale repo, and shipping them with the releases, @Chris-Hawblitzel might be more knowledgeable than me about this.
We also extended the Makefile to warn if these config files are not present with the vale executables, for instance if Vale was directly installed from the sources.

For existing HACL* developers, this switch of tools is a breaking change. The transition is however smooth, the required steps are the following:
1. Install dotnet (on Ubuntu, by running `sudo snap install dotnet-sdk`)
2. Run ./tools/get_vale.sh

tahina-pro (approved):
I have green CI on my Everest branch: https://github.com/project-everest/everest/actions/runs/2430831431
So, on my side, I am good to go. Once this PR is merged, my next step is to merge that Everest branch into master. In particular, I removed the logic to fetch vale from the Everest script, calling `tools/get_vale.sh` from HACL* instead. I'll also do the same thing for F* CI when merging 2513.
Thank you again!

Chris-Hawblitzel (approved):

PR #565, opened 2022-06-02
Support for returning uint64 across the WASM/JS boundary

msprotz:
This is the companion PR to FStarLang/karamel#267 --  it enables returning uint64s as a multi-value, and the "nice" API layer on top of it is aware of that calling convention and uses a JS bignum to represent a 64-bit integer adequately.

There are some extra fixes to be reviewed by @denismerigoux ... namely, the `int32` return value was renamed to `uint32`, because as far as I know, all of the functions that were bound by the API return *unsigned* values... and the api.js code was changed accordingly to interpret those as unsigned via a logical-shift (`>>> 0`), otherwise, a HACL function that returns `0xffffffff` would be returning `-1` in JS -- can you confirm that, as far as you remember, there was no usage of *signed* integers in the code for which you wrote bindings?


msprotz (commented):
Note that this finally allows exposing the complete Bignum64 API!

denismerigoux (approved):
Nope, no signed integers returned in my memory. The `>>> 0` is horrible but indeed you need it...

W95Psp:
Sure, sorry, I completely forgot about this PR! It looks good to me!

W95Psp (approved):

PR #566, opened 2022-06-06
Changes for F* PR#2466

aseemr:
Changes for F* PR https://github.com/FStarLang/FStar/pull/2603.

### DIFF
--- a/code/meta/AUTHORS.md
+++ b/code/meta/AUTHORS.md
@@ -1,1 +1,1 @@
-This code was authored by Jonathan Protzenko.
+This code was authored by Jonathan Protzenko (MSR).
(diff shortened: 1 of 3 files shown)


msprotz (approved):
Thanks for fixing this Aseem, much appreciated!

Does this mean an `fvar` may now appear either underneath Tv_FVar or Tv_UInst?

R1kM (approved):

aseemr:
Hi Jonathan, yes, e.g. the full syntax for `List.length` is `List.length<u> #a l`, where `u` is the universe of the type `a`. When reflecting on it, earlier we were removing the universe and returning it just as an fvar. Now, it will be reflected as a `Tv_UInst`, and the clients may throw away the universes if they want.

PR #569, opened 2022-06-13
Replace `/usr/bin/time` with `which time`

pnmadelaine:
`/usr/bin/time` is not available on all Linux distributions.
This change uses `which` to locate the `time` command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -202,7 +202,7 @@ else ifeq ($(shell uname -s),FreeBSD)
     TIME := /usr/bin/time
 else
   SED := sed -i
-  TIME := /usr/bin/time -q -f '%E'
+  TIME := $(shell which time) -q -f '%E'
 endif
 
 ifneq ($(OS),Windows_NT)


R1kM (approved):

pnmadelaine:
I don't know if that assumption is reasonnable, but I figure the assumption was already made about `/usr/bin/time`, is there a reason why that would be more reasonnable?

msprotz:
> on some distributions (notably arch), which won't find the time command even when it's installed as /usr/bin/time.

?!! Why?!

pnmadelaine:
I don't know why, but we just tested it.

msprotz:
🤯

pnmadelaine:
Should I go ahead and merge then?

msprotz:
It sounds like merging would break zsh users, which is the default shell on OSX nowadays. 

How about `whereis -b -q time`?

msprotz:
Ok, `-q` is not posix. How about `whereis -b time | cut -d ' ' -f 2-`?

pnmadelaine:
> It sounds like merging would break zsh users

doesn't `$(shell ...)` in a Makefile call to bash anyway?

msprotz:
Only if the Makefile has SHELL=/bin/bash... does it?

R1kM:
What about using `command -v time` instead of `which` ? From some posts on StackOverflow (e.g., https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script), it seems to be a POSIX-compliant solution.

franziskuskiefer:
Throwing in another option 😄  `where time | tail -n1`

msprotz:
https://github.com/project-everest/hacl-star/blob/master/Makefile#L219

pnmadelaine:
> https://github.com/project-everest/hacl-star/blob/master/Makefile#L219

well, no problem then?

msprotz:
I think we're good.

@R1kM :  `command -v time` returns the shell built-in

So yeah you can go ahead and merge. Thanks

PR #571, opened 2022-06-15
More rlimit + ifuel

R1kM:
Similar to #570. The fact that these bumps in ifuel are needed for Meta only are quite surprising.
@msprotz had a hypothesis that proofs might succeed in Hacl.Impl.HPKE without the bump might be due to hint files helping with the proof. I managed to verify the entirety of Hacl.Impl.HPKE locally without hints, which seems to suggest the reason is to be found elsewhere...

### DIFF
--- a/code/hpke/Hacl.Impl.HPKE.fst
+++ b/code/hpke/Hacl.Impl.HPKE.fst
@@ -665,7 +665,7 @@ val encap:
      )
 
 #restart-solver
-#push-options "--z3rlimit 500 --z3refresh"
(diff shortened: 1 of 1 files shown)


pnmadelaine (approved):

beurdouche (commented):
600 is quite high, this proof should be stabilized a different way...

R1kM:
@beurdouche I would tend to agree, unfortunately, the spec itself already requires very high rlimits (up to 1000 in some cases). Furthermore, the flakiness seems to only occur in the meta-generated version (post inlining) of this function, which makes it really hard to debug.

PR #572, opened 2022-07-01
Hints refresh

nikswamy:
Hints haven't been refreshed in HACL* for a while, I think.

With an upcoming change to F* tactics engine, we noticed that some proofs were taking very long without hints both using F* master and with our upcoming patch. 

So here are newly generated hints ... thanks @aseemr!


msprotz (approved):
I believe @pnmadelaine and @W95Psp are close to completion on the nightly build, but in the meanwhile this is very helpful. Thanks!

msprotz:
Need two approvals to merge this. @pnmadelaine, @aseemr, @W95Psp, @tahina-pro ?

R1kM (approved):

PR #573, opened 2022-07-01
Refresh dist/ using FStarLang/karamel#271

### LINKED ISSUE #523: SHA2 vec128 depends on vec256 header
This might have been introduced with the internal headers.
The vec128 SHA2 implementation now includes `Hacl_SHA2_Vec256.h`. There is no `Hacl_SHA2_Vec128.h` that would define the types required such as `K____uint8_t__K____uint8_t__K____uint8_t___uint8_t_` (which could also get a better name 😬 ).

https://github.com/project-everest/hacl-star/blob/c4496a13bd85970b6b15a061540adf926fdefc93/dist/gcc-compatible/Hacl_SHA2_Vec128.c#L27

@msprotz thoughts how this could get fixed? Since the typedefs aren't specific to the vectorised versions I'd suggest to pull them out into a common header.

msprotz:
This fixes #523


dist/gcc-compatible/Hacl_Impl_Poly1305_Field32xN.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <string.h>
+#include "krml/internal/types.h"
+#include "krml/lowstar_endianness.h"
+#include "krml/internal/target.h"
+
+
+
+#include "evercrypt_targetconfig.h"
+typedef void *Hacl_Impl_Poly1305_Field32xN_fmul_precomp_r_pre;
```
polubelova, line 40: I think [`fmul_precomp_r_pre`](https://github.com/project-everest/hacl-star/blob/7307fe02cc0d9778b9ddd4e54630655d2e36ccbf/code/poly1305/Hacl.Impl.Poly1305.Field32xN.fst#L81)  should be marked as `inline_for_extraction noextract`

dist/gcc-compatible/Hacl_P256.c
```diff
@@ -592,11 +592,19 @@ static void montgomery_multiplication_buffer(uint64_t *a, uint64_t *b, uint64_t
   uint64_t t[8U] = { 0U };
   uint64_t round2[8U] = { 0U };
   uint64_t round4[8U] = { 0U };
+  uint32_t size_0 = (uint32_t)0U;
```
polubelova, line 595: should we mark `size_i` with `[@inline_let]`?

dist/gcc-compatible/Hacl_Streaming_Functor.h
```diff
@@ -37,25 +37,17 @@ extern "C" {
 
 
 #include "evercrypt_targetconfig.h"
-typedef uint8_t EverCrypt_Helpers_uint8_t;
+typedef void *Hacl_Streaming_Functor_invariant_s;
```
polubelova, line 40: should we mark these definitions (`invariant_s`, `freeable`, `preserves_freeable`, `invariant`) as `inline_for_extraction noextract`?

providers/evercrypt/EverCrypt.Specs.fsti
```diff
@@ -1,52 +1,52 @@
...
-
-let aead_create_pre = fun _ -> False
-let aead_create_post = fun _ _ _ -> True
-let aead_encrypt_pre = fun _ -> False
-let aead_encrypt_post = fun _ _ _ -> True
-let aead_decrypt_pre = fun _ -> False
-let aead_decrypt_post = fun _ _ _ -> True
-let aead_free_pre = fun _ -> False
-let aead_free_post = fun _ _ _ -> True
+inline_for_extraction noextract let curve_x25519_pre = fun _ -> False
```
polubelova, line 3: out of curiosity, why do you need this file?

franziskuskiefer (approved):
Looks like this fixes the issue.
It also pulls out `Hacl_Bignum_MontArithmetic_bn_mont_ctx_u64_s` in `Hacl_Bignum.h`, which isn't really needed. But it's ok to have there.
So lgtm

msprotz:
@polubelova all good comments, thanks, I'll address them and re-submit a PR.

For the EverCrypt.Specs: it's only there because of super-legacy code that is kept alive for miTLS. We need to ditch it but need to figure out how to retain a migration plan for miTLS... which is a big discussion. Hence why this thing remains (it was supposed to be a temporary placeholder!).

dist/gcc-compatible/Hacl_Impl_Poly1305_Field32xN.h
msprotz: good catch this file is now gone... for some reason I was looking at the output of `git diff` *before* I ran git add on the new files, so I didn't catch these new added files... thanks for spotting this!

dist/gcc-compatible/Hacl_P256.c
msprotz: done

dist/gcc-compatible/Hacl_Streaming_Functor.h
msprotz: same reason, I added suitable flags except for those that were generating new files... this is now fixed and this file is gone, too

providers/evercrypt/EverCrypt.Specs.fsti
msprotz: answered on the main thread

polubelova (approved):
Looks good to me. Thanks Jonathan!

msprotz:
https://github.com/FStarLang/karamel/pull/271 has an Everest green so I'm going ahead and merging this

PR #577, opened 2022-07-19
define fpow once for curve25519 and ed25519

polubelova:
This PR uses a modular exponentiation and its properties from `lib/` for both `curve25519` and `ed25519`. 
This is just a small clean-up of specifications and proofs.


beurdouche (approved):
👍

karthikbhargavan (approved):
Nice

msprotz (approved):
Great!

@R1kM this may be useful to fix that stack overflow in the spec tests for HPKE...? I feel like this is exactly the lemma you need

polubelova:
As for P256, I think one could replace [`modp_inv2_pow`](https://github.com/project-everest/hacl-star/blob/master/specs/ecdsap256/Spec.P256.Lemmas.fst#L151) with [`modp_inv2`](https://github.com/project-everest/hacl-star/blob/master/specs/ecdsap256/Spec.P256.Lemmas.fst#L145) in [`_norm`](https://github.com/project-everest/hacl-star/blob/master/specs/ecdsap256/Spec.P256.fst#L84).

- there is no need to compute modular reduction in [`modp_inv_prime`](https://github.com/project-everest/hacl-star/blob/master/specs/ecdsap256/Spec.P256.Lemmas.fst#L139), as the result of `exp` is already less than `prime`.
- `pow x (prime256 - 2) % prime256` could indeed lead to a stack overflow, as it first computes a large number `x^(prime256-2)` and then performs a modular reduction.

Frosne:
Indeed!

It will be very nice if someone could replace it and reverify if needed (:

PR #578, opened 2022-07-20
hints and dist regeneration

pnmadelaine:
Since master is protected and requires two reviews to push, there seems to be only two solutions to merge hints and dist:
-use a bot with admin permissions on the repo to override the protection
-automatically generate and manually merge PRs

This implements the second solution, and updates hints and dist weekly instead of daily to reduce the inconvenience.

Following the suggestion from @polubelova I also added a check to avoid trivial updates.


polubelova (approved):
Thanks a lot!

msprotz (approved):
Looks good to me, thans!

msprotz:
Can you confirm somewhere in the build someone runs rm -rf dist/*/* ? This is the important bit in case some change results in a file being deleted

pnmadelaine:
@msprotz this is done in the Nix build

PR #582, opened 2022-07-26
Tentative improvements to the code-gen

msprotz:
https://github.com/FStarLang/karamel/pull/274 contains a potential improvement to remove the amount of unused variables

This is not necessarily a slam dunk. I'll comment on specific bits of the diff to point out potential issues.

### DIFF
--- a/dist/gcc-compatible/EverCrypt_CTR.c
+++ b/dist/gcc-compatible/EverCrypt_CTR.c
@@ -104,8 +104,8 @@ EverCrypt_CTR_create_in(
           uint8_t *ek = KRML_HOST_CALLOC((uint32_t)304U, sizeof (uint8_t));
           uint8_t *keys_b = ek;
           uint8_t *hkeys_b = ek + (uint32_t)176U;
-          uint64_t scrut = aes128_key_expansion(k, keys_b);
-          uint64_t scrut0 = aes128_keyhash_init(keys_b, hkeys_b);
+          (void)aes128_key_expansion(k, keys_b);
+          (void)aes128_keyhash_init(keys_b, hkeys_b);
           uint8_t *iv_ = KRML_HOST_CALLOC((uint32_t)16U, sizeof (uint8_t));
           memcpy(iv_, iv, iv_len * sizeof (uint8_t));
           KRML_CHECK_SIZE(sizeof (EverCrypt_CTR_state_s), (uint32_t)1U);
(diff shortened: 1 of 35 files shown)


dist/gcc-compatible/EverCrypt_CTR.c
```diff
@@ -104,8 +104,8 @@ EverCrypt_CTR_create_in(
           uint8_t *ek = KRML_HOST_CALLOC((uint32_t)304U, sizeof (uint8_t));
           uint8_t *keys_b = ek;
           uint8_t *hkeys_b = ek + (uint32_t)176U;
-          uint64_t scrut = aes128_key_expansion(k, keys_b);
-          uint64_t scrut0 = aes128_keyhash_init(keys_b, hkeys_b);
+          (void)aes128_key_expansion(k, keys_b);
```
msprotz, line 107: this is great, and the kind of change that motivated this patch

dist/gcc-compatible/EverCrypt_DRBG.c
```diff
@@ -1011,10 +1011,9 @@ EverCrypt_DRBG_generate_sha1(
   {
     x1 = KRML_EABORT(Hacl_HMAC_DRBG_state, "unreachable (pattern matches are exhaustive in F*)");
   }
-  bool b;
   if (x1.reseed_counter[0U] > Hacl_HMAC_DRBG_reseed_interval)
   {
-    b = false;
+
```
msprotz, line 1016: interesting side-effect of this patch, but why not -- need to improve code-gen to skip empty if blocks

dist/gcc-compatible/EverCrypt_Hash.c
```diff
@@ -402,7 +401,7 @@ void EverCrypt_Hash_init(EverCrypt_Hash_state_s *s)
     r1[1U] = iv5;
     r1[2U] = iv6;
     r1[3U] = iv7;
-    FStar_UInt128_uint128 uu____1 = FStar_UInt128_uint64_to_uint128((uint64_t)0U);
+    (void)FStar_UInt128_uint64_to_uint128((uint64_t)0U);
```
msprotz, line 404: just weird! I think this patch makes things worse, but for this specific instance, I've patched krml to "know" that this function is pure, so we can easily remove this quirk

msprotz: (in a future iteration)

dist/gcc-compatible/EverCrypt_Hash.c
```diff
@@ -441,7 +440,7 @@ void EverCrypt_Hash_update_multi_256(uint32_t *s, uint8_t *blocks, uint32_t n)
   if (has_shaext && has_sse)
   {
     uint64_t n1 = (uint64_t)n;
-    uint64_t scrut = sha256_update(s, blocks, n1, k224_256);
+    sha256_update(s, blocks, n1, k224_256);
```
msprotz, line 443: why does this one *not* have the void cast?

dist/gcc-compatible/EverCrypt_Hash.c
```diff
@@ -1249,7 +1242,7 @@ EverCrypt_Hash_Incremental_init(Hacl_Streaming_Functor_state_s___EverCrypt_Hash_
   Hacl_Streaming_Functor_state_s___EverCrypt_Hash_state_s____ scrut = *s;
   uint8_t *buf = scrut.buf;
   EverCrypt_Hash_state_s *block_state = scrut.block_state;
-  Spec_Hash_Definitions_hash_alg i = EverCrypt_Hash_alg_of_state(block_state);
+  (void)EverCrypt_Hash_alg_of_state(block_state);
```
msprotz, line 1245: this is the worst: krml doesn't have an inter-procedural analysis so doesn't know that this function is "pure" (for what definition of pure?) and therefore "voids" away the result, but this is definitely strange -- need to work on source code to get rid of this

PR #589, opened 2022-08-11
fix memory leak when creating keys for rsa-pss

### LINKED ISSUE #588: Memory leaks in RSA PSS
RSA PSS leaks memory in error cases.

`Hacl_RSAPSS_new_rsapss_load_pkey` and `Hacl_RSAPSS_new_rsapss_load_skey` return `NULL` in the error case without freeing `pkey`/`skey`.

https://github.com/project-everest/hacl-star/blob/1b7df371e8fd968824842c7634a4171b1332651e/dist/gcc-compatible/Hacl_RSAPSS.c#L634-L638

https://github.com/project-everest/hacl-star/blob/1b7df371e8fd968824842c7634a4171b1332651e/dist/gcc-compatible/Hacl_RSAPSS.c#L721-L725

polubelova:
This PR fixes #588. 

Thanks @franziskuskiefer for reporting the issue.


msprotz (approved):
super fast patch, thanks Marina

PR #590, opened 2022-08-15
Replace =. with =

msprotz:
=. evaluates to FStar.UInt32.eq which F*'s normalizer won't reduce --
there is no need to use the special operator, instead, it suffices to
use =, the polymorphic equality operator, and all is well


franziskuskiefer (approved):
nice! That's an easy fix 😄

msprotz:
FYI I am looking at what it would take to fix the root problem rather than fixing the use-site. But this will be a separate PR.

polubelova (approved):
Looks great, thanks!

PR #591, opened 2022-08-15
Refresh dist/gcc-compatible with some code-gen improvements

msprotz:
I recently pushed a series of fixes to krml: from the changelog

```
    Improve code-generation for arrays of vectors.
    
    - krml is now aware that HACL's vec128_zero, vec256_zero and so on are
      zeroes and as such that a for-loop is not needed for compiling
      allocations, and a zero-initializer suffices
    - an if-statement that is about to compiled as an ifdef can be safely
      ignored for the sake of the buffer-hoisting phase
    - improved buffer hoisting: if the initial value is a closed term, do
      not bother with a copy assignment and lift the array declaration
      wholesale
```

This refreshes dist/ accordingly.

### DIFF
--- a/dist/gcc-compatible/Hacl_HMAC_Blake2s_128.c
+++ b/dist/gcc-compatible/Hacl_HMAC_Blake2s_128.c
@@ -82,9 +82,7 @@ Hacl_HMAC_Blake2s_128_compute_blake2s_128(
     uint8_t yi = key_block[i];
     opad[i] = xi ^ yi;
   }
-  Lib_IntVector_Intrinsics_vec128 s[4U];
-  for (uint32_t _i = 0U; _i < (uint32_t)4U; ++_i)
-    s[_i] = Lib_IntVector_Intrinsics_vec128_zero;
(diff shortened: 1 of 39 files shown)


dist/gcc-compatible/Hacl_HMAC_Blake2s_128.c
```diff
@@ -166,25 +164,15 @@ Hacl_HMAC_Blake2s_128_compute_blake2s_128(
...
-    uint64_t
-    ev2 =
-      Hacl_Hash_Blake2s_128_update_last_blake2s_128(s0,
-        ev1,
-        (uint64_t)(uint32_t)64U,
-        hash1,
-        (uint32_t)32U);
-    ev11 = ev2;
-  }
+  uint64_t ev1 = Hacl_Hash_Blake2s_128_update_multi_blake2s_128(s0, ev0, opad, (uint32_t)1U);
```
msprotz, line 167: this is from my earlier fix to hmac, not specific to the krml change

polubelova (approved):
Looks good, thanks a lot!

PR #592, opened 2022-08-15
Make sure =. reduces to =

msprotz:
This should prevent issues like https://github.com/project-everest/hacl-star/pull/590 from happening again in the future.

I regenerated dist/ with this and it doesn't improve anything currently.

### DIFF
--- a/lib/Lib.IntTypes.fst
+++ b/lib/Lib.IntTypes.fst
@@ -1065,18 +1065,7 @@ let mod_lemma #t a b =
   | S64 -> Int.pow2_values 64
 
 let eq #t x y =
-  match t with
-  | U1   -> UInt8.eq x y
-  | U8   -> UInt8.eq x y
-  | U16  -> UInt16.eq x y
-  | U32  -> UInt32.eq x y
-  | U64  -> UInt64.eq x y
-  | U128 -> UInt128.eq x y
-  | S8   -> Int8.eq x y
-  | S16  -> Int16.eq x y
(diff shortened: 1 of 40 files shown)


msprotz:
(There is currently a diff in dist/ because this is on top of #591 but once #591 is merged the diff will reduce to the one change in lib/)

lib/Lib.IntTypes.fst
```diff
@@ -1065,18 +1065,7 @@ let mod_lemma #t a b =
...
-  | U16  -> UInt16.eq x y
-  | U32  -> UInt32.eq x y
-  | U64  -> UInt64.eq x y
-  | U128 -> UInt128.eq x y
-  | S8   -> Int8.eq x y
-  | S16  -> Int16.eq x y
-  | S32  -> Int32.eq x y
-  | S64  -> Int64.eq x y
-  | S128 -> Int128.eq x y
+  x = y
```
msprotz, line 1068: this is the only important change, the rest is not relevant

msprotz:
@karthikbhargavan I don't think I'm breaking any of the important properties of Lib.IntTypes with this change, but please do take a look

polubelova (approved):
Thanks!

This change doesn't break properties of `Lib.IntTypes` as `eq` is defined on public integers.

karthikbhargavan (approved):
Looks good. As far as I see this just affects extraction. Secret inttypes do not have the "=." operator so this should not affect those.

PR #593, opened 2022-08-23
Macro-based loop unrolling

msprotz:
Right now, krml has an option `-funroll-loops` that replaces for-loops whose bounds are statically known by a series of C blocks for each possible value of the index.

Following many discussions, it turns out that this isn't great:
- generated C code is ugly
- prevents different degrees of unrolling depending on target
- unrolling loops in F* isn't great either because this is not always the right thing to do.

This PR demonstrates the effect of https://github.com/FStarLang/karamel/pull/280, a krml change that relies on macros to perform loop unrolling.

Basically, if krml is passed the option `-funroll-loops N`, and encounters a for loop of the form `for (int i = Z; i < M; i += K)` where Z, M and K are constants, and where `(M - Z + K - 1)/K <= N` (i.e. `ceil((M-Z)/K) <= N`), then instead of generating a for-loop, krml generates `KRML_MAYBE_FORT` where `T` is the number of iterations computed using the formula above.

Then, at compile-time, the user can pass `-DKRML_UNROLL_MAX=N'` where `N' <= N` -- any loops shorter than `N'` will be macro-expanded, and other loops will macro-expand to an actual for-loop.

Note that the argument to `-funroll-loops` is perhaps something we can dispense with; I was just thinking that it made no sense to generate `KRML_MAYBE_FOR256` when we very well know that we're not going to macro-expand such a big loop. But we could also hardcode 16 as the limit.

Also note that owing to limitations of the C preprocessor, I had to hardcode an upper limit to the number of possible unrolling levels. This is essentially due to the fact that a macro's expansion cannot contain `#if`.

### DIFF
--- a/dist/gcc-compatible/Hacl_Curve25519_64_Slow.c
+++ b/dist/gcc-compatible/Hacl_Curve25519_64_Slow.c
@@ -66,34 +66,24 @@ static inline uint64_t add1_(uint64_t *out, uint64_t *f1, uint64_t f2)
 static inline void fadd_(uint64_t *out, uint64_t *f1, uint64_t *f2)
 {
   uint64_t c0 = (uint64_t)0U;
-  KRML_MAYBE_FOR1(i,
-    (uint32_t)0U,
-    (uint32_t)1U,
-    (uint32_t)1U,
-    uint64_t t1 = f1[(uint32_t)4U * i];
-    uint64_t t20 = f2[(uint32_t)4U * i];
-    uint64_t *res_i0 = out + (uint32_t)4U * i;
+  {
+    uint64_t t1 = f1[(uint32_t)4U * (uint32_t)0U];
+    uint64_t t20 = f2[(uint32_t)4U * (uint32_t)0U];
+    uint64_t *res_i0 = out + (uint32_t)4U * (uint32_t)0U;
     c0 = Lib_IntTypes_Intrinsics_add_carry_u64(c0, t1, t20, res_i0);
-    uint64_t t10 = f1[(uint32_t)4U * i + (uint32_t)1U];
-    uint64_t t21 = f2[(uint32_t)4U * i + (uint32_t)1U];
-    uint64_t *res_i1 = out + (uint32_t)4U * i + (uint32_t)1U;
+    uint64_t t10 = f1[(uint32_t)4U * (uint32_t)0U + (uint32_t)1U];
+    uint64_t t21 = f2[(uint32_t)4U * (uint32_t)0U + (uint32_t)1U];
+    uint64_t *res_i1 = out + (uint32_t)4U * (uint32_t)0U + (uint32_t)1U;
     c0 = Lib_IntTypes_Intrinsics_add_carry_u64(c0, t10, t21, res_i1);
-    uint64_t t11 = f1[(uint32_t)4U * i + (uint32_t)2U];
-    uint64_t t22 = f2[(uint32_t)4U * i + (uint32_t)2U];
-    uint64_t *res_i2 = out + (uint32_t)4U * i + (uint32_t)2U;
+    uint64_t t11 = f1[(uint32_t)4U * (uint32_t)0U + (uint32_t)2U];
+    uint64_t t22 = f2[(uint32_t)4U * (uint32_t)0U + (uint32_t)2U];
+    uint64_t *res_i2 = out + (uint32_t)4U * (uint32_t)0U + (uint32_t)2U;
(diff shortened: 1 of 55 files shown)


msprotz:
Another neat thing is that this also makes it possible to have different unrolling levels depending on the file, using Makefile syntax:

```
Hacl_Blake2s.o: CFLAGS += -DKRML_UNROLL_MAX=8
Hacl_Blake2s_128.o: CFLAGS += -DKRML_UNROLL_MAX=16
```

polubelova (approved):
Looks great, thanks a lot!

Is it possible to add two special cases? The first is to drop `KRML_MAYBE_FOR0` in the extracted code and the second is to replace `KRML_MAYBE_FOR1` with the loop body.

franziskuskiefer (changes requested):
Thanks! I think this makes things way nicer.
Some comments though.

Benchmarks with this give me

```
------------------------------------------------------------
Benchmark                  Time             CPU   Iterations
------------------------------------------------------------
BM_Blake2b_32            279 ns          279 ns      2505602
BM_Blake2s_32           1851 ns         1851 ns       378993
BM_Blake2s_vec128       1821 ns         1821 ns       384004
BM_Blake2b_vec256       1577 ns         1577 ns       443845
```

compared to 

```
------------------------------------------------------------
Benchmark                  Time             CPU   Iterations
------------------------------------------------------------
BM_Blake2b_32            310 ns          310 ns      2260485
BM_Blake2s_32           1890 ns         1890 ns       378160
BM_Blake2s_vec128       1825 ns         1825 ns       383793
BM_Blake2b_vec256       1580 ns         1580 ns       442440
```

before.
So looks like performance is the same. That's good!

First I was a little surprised that the performance is the same even without setting `KRML_UNROLL_MAX`. We should make clear that 16 is the default value for unrolling when not setting anything.

Setting `-DKRML_UNROLL_MAX=1` throws an error `Hacl_Hash_Blake2b_256.c:212:1: error: expected ‘while’ before ‘void’`. `target.h:190` is missing a closing `}`.

msprotz:
With the way the C preprocessor works, I would have to double the amount of macros to go to 32. (See the changes in dist/karamel). Do we want to unroll past 16 ?

franziskuskiefer:
> Do we want to unroll past 16 ?

I don't think so. We can always add it later if needed. We should just explain it somewhere.

msprotz:
I pushed the series of requested changes, including the fix for the missing } and eliminating FOR0 and FOR1. Look at this commit specifically to see the incremental effect of the latest set of changes: https://github.com/project-everest/hacl-star/pull/593/commits/8f63b07805c9b24265fcb56413af29e4ca2eff69

dist/gcc-compatible/Hacl_Curve25519_64_Slow.c
```diff
@@ -66,32 +66,24 @@ static inline uint64_t add1_(uint64_t *out, uint64_t *f1, uint64_t f2)
...
-    uint64_t t20 = f2[(uint32_t)4U * i];
-    uint64_t *res_i0 = out + (uint32_t)4U * i;
+    uint64_t t1 = f1[(uint32_t)4U * (uint32_t)0U];
+    uint64_t t20 = f2[(uint32_t)4U * (uint32_t)0U];
+    uint64_t *res_i0 = out + (uint32_t)4U * (uint32_t)0U;
     c0 = Lib_IntTypes_Intrinsics_add_carry_u64(c0, t1, t20, res_i0);
-    uint64_t t10 = f1[(uint32_t)4U * i + (uint32_t)1U];
-    uint64_t t21 = f2[(uint32_t)4U * i + (uint32_t)1U];
-    uint64_t *res_i1 = out + (uint32_t)4U * i + (uint32_t)1U;
+    uint64_t t10 = f1[(uint32_t)4U * (uint32_t)0U + (uint32_t)1U];
```
franziskuskiefer, line 74: Using `0`(above)/`1` etc. instead of the index style (`(uint32_t)4U * (uint32_t)0U`) would be nicer. But I'd be fine with leaving it as is if it's too much work.

franziskuskiefer (approved):
Thanks
There's one more thing but it's just cosmetics.

dist/gcc-compatible/Hacl_Curve25519_64_Slow.c
msprotz: krml doesn't have a CSE phase (normally this happens ahead of time anyhow in F*), I'm not against adding one, but this should be a separate issue... thanks! :-)

PR #603, opened 2022-09-15
 Tweaking some rlimits

aseemr:
There are two assertions that I have seen are hard for the solver to prove, would be good to prove them more explicitly to guard against future breakages.


beurdouche (approved):

msprotz (approved):
Looks good to me, thanks for adding comments that point to the assertion failures

PR #604, opened 2022-09-15
Update exponentiation

polubelova:
This PR introduces the following.
- adds support for an abelian group and proves its main properties 
  - it's useful for implementing new methods for Elliptic Curves Scalar Multiplication in the future as a point negation is a cheap operation
- efficiently computes a product of four exponentiations 
  - it's needed for implementing GLV in `secp256k1-ecdsa-verify`
- precomputes a table for fixed-window exponentiation methods more efficiently by calling `sqr` where it's possible
- exposes more functions from fixed-window exponentiation methods which take a precomputed table as an argument
  - a precomuted table can be shared between calls to an exponentiation
  - a table can be precomputed in advance for a known base


msprotz (commented):
I only have high-level questions (I only skimmed the proofs).
- Do you want me to push a dist refresh so that only the changes related to your PR appear in the diff?
- Could this be used to precompute tables for `secret_to_public` in curve25519?
- Should we provide some of these tables in the resulting C code? (Apologies if this is done already, I couldn't see it with the big diff.)

Thanks,

Jonathan

polubelova:
Thanks for looking into this!

> Do you want me to push a dist refresh so that only the changes related to your PR appear in the diff?

That would be great, thanks!

> Could this be used to precompute tables for secret_to_public in curve25519?

It could definitely improve the performance of `ed25519-ecdsa-sign` and `ed25519-ecdsa-verify` a bit (~5-10%).
`curve25519-secret-to-public` doesn't call ECSM from ed25519 but it could be done as future work.

> Should we provide some of these tables in the resulting C code? (Apologies if this is done already, I couldn't see it with the big diff.)

That was done as an experiment in [this commit](https://github.com/hacl-star/hacl-star/commit/ef886fc8079cb50fa3c5e1e9363c3cf59c99ddc9). Do you think it would be possible to generate such a table directly in F* from a [`lprecomp_table`](https://github.com/hacl-star/hacl-star/blob/polubelova_new_exp/code/bignum/Hacl.Impl.PrecompTable.fsti#L60) function with some magic? Otherwise, I don't see how we can add support for a larger window size. It will be useful for `secp256k1-ecdsa-verify`, where we could significantly improve the performance by setting the window size to 15 (around 8192 points): https://github.com/hacl-star/hacl-star/blob/polubelova_secp256k1_wnaf/code/k256/dist%20(GLV%20%2B%20wNAF%20%2B%20precomp15)/precomp_table_g.h

msprotz:
I would precompute it by reducing the spec-level functions at compile-time. What is the pure equivalent of this stateful function?

polubelova:
For now, the functional correctness of the `lprecomp_table` function (i.e., `forall (i:nat{i < table_len}). refl (get_ith_element table i) == Lib.Exponentiation.pow (refl a) i`) is [proved](https://github.com/hacl-star/hacl-star/blob/7957cd916fcefe17e6ef5f18b249ebe4193c93db/code/bignum/Hacl.Impl.PrecompTable.fsti#L55) without introducing any intermediate specifications. 

If it helps, here is a high-level structure of the proofs for ECSM in secp256k1.
```
felem = x:nat{x < prime} // where `prime` is 256 bit long

                                  a high-level specification
                             (an affine point (x, y) : felem & felem)
                   == to_aff_point                                   == refl
            (a low-level specification                                (a low-star implementation
(a projective point (X, Y, Z) : felem & felem & felem))      	(a projective point (X, Y, Z) : lbuffer uint64 15ul)
							         where `felem` is represented as an array of 5 elements, with a 2^52-radix)
```

The Low* implementation uses its own low-level specification only for a finite-field arithmetic (where `felem` is represented as a tuple of 5 elements so that we can directly call it in Low*) but not for ECSM or point addition/doubling.

Maybe we can compute this table using a low-level specification (`felem = nat{..}`) and then convert it to a low-level representation (`lseq uint64 5`)? At worst, I can spend some time on rewriting the proofs for `lprecomp_table`.

msprotz (approved):
Ok this looks all good to me. I'll let someone else provide a second review before merging.

Thanks for rebasing for a clean diff.

I looked into the perf issue with precomputing tables and `seq_of_list` seems to be the bottleneck (which wouldn't be surprising, since it's quadratic). I requires a little bit of thinking to optimize, so let's not hold up this PR on it.

karthikbhargavan (approved):
The changes look good to me. I would really love to see perf improvements from precomputation in all secret_to_public functions as well, so let's try to generalize this in a future PR.

msprotz:
@polubelova I'm happy for you to land this at any point -- let me know if you want me to hold off onto anything I have so that you can land this PR

PR #605, opened 2022-09-15
add GLV for secp256k1-ecdsa-verify

polubelova:
This PR improves the performance of `secp256k1-ecdsa-verify` from 298379 to 237964 CPU cycles, i.e., by 1.25x.

### DIFF
--- a/code/k256/Hacl.Impl.K256.GLV.Constants.fst
+++ b/code/k256/Hacl.Impl.K256.GLV.Constants.fst
@@ -1,0 +1,322 @@
+module Hacl.Impl.K256.GLV.Constants
+
+open FStar.HyperStack
+open FStar.HyperStack.ST
+open FStar.Mul
+
+open Lib.IntTypes
+open Lib.Buffer
+
+module ST = FStar.HyperStack.ST
+
+module S = Spec.K256
+module SG = Hacl.Spec.K256.GLV
+module SGL = Hacl.Spec.K256.GLV.Lemmas
+
+open Hacl.K256.Field
+open Hacl.K256.Scalar
+open Hacl.Impl.K256.Point
+
+#set-options "--z3rlimit 50 --fuel 0 --ifuel 0"
+
+inline_for_extraction noextract
+val make_minus_lambda: f:qelem -> Stack unit
+  (requires fun h -> live h f)
+  (ensures  fun h0 _ h1 -> modifies (loc f) h0 h1 /\
+    qas_nat h1 f == SG.minus_lambda)
+
+let make_minus_lambda f =
+  [@inline_let]
+  let x =
+   (u64 0xe0cfc810b51283cf,
+    u64 0xa880b9fc8ec739c2,
+    u64 0x5ad9e3fd77ed9ba4,
+    u64 0xac9c52b33fa3cf1f) in
+
+  assert_norm (SG.minus_lambda == qas_nat4 x);
+  make_u64_4 f x
+
+
+inline_for_extraction noextract
+val make_beta: f:felem -> Stack unit
+  (requires fun h -> live h f)
+  (ensures  fun h0 _ h1 -> modifies (loc f) h0 h1 /\
+    as_nat h1 f == SG.beta /\ inv_fully_reduced h1 f)
+
+let make_beta f =
+  [@inline_let]
+  let x =
+   (u64 0x96c28719501ee,
+    u64 0x7512f58995c13,
+    u64 0xc3434e99cf049,
+    u64 0x7106e64479ea,
+    u64 0x7ae96a2b657c) in
+
+  assert_norm (0x96c28719501ee <= max52);
+  assert_norm (0x7ae96a2b657c <= max48);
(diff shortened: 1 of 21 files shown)


msprotz (commented):
I'm having a hard time reviewing this PR.
- What is GLV? Can you put a comment somewhere like "this implements the GLV method as described in ..."
- Where is the central lemma? Is it `lemma_glv`? What does it say?
- I still see one `assume val`. Can you at least comment on why it has to be there?

My impression is that it's a different system of coordinates that allows for a more efficient multiplication, along with proofs that it's ok to convert back and forth, but it'd be good to have a high-level description :-). Thanks!

polubelova:
You should start with [`code/k256/Hacl.Spec.K256.GLV.fst`](https://github.com/hacl-star/hacl-star/blob/polubelova_secp256k1_glv/code/k256/Hacl.Spec.K256.GLV.fst) :)

It says that we implement the following two functions from [libsecp256k1](https://github.com/bitcoin-core/secp256k1): [`secp256k1_scalar_split_lambda`](https://github.com/bitcoin-core/secp256k1/blob/a43e982bca580f4fba19d7ffaf9b5ee3f51641cb/src/scalar_impl.h#L123) and  [`secp256k1_ecmult_endo_split`](https://github.com/bitcoin-core/secp256k1/blob/a43e982bca580f4fba19d7ffaf9b5ee3f51641cb/src/ecmult_impl.h#L618). 

As usual, the idea is to prove that our new optimized `secp256k1` implementation is functionally correct w.r.t the high-level specification given in [`Spec.K256.fst`](https://github.com/hacl-star/hacl-star/blob/main/specs/Spec.K256.fst). The correctness relies on the [`lemma_glv`](https://github.com/hacl-star/hacl-star/blob/be4a57ccaa038d838fd1db6c31d02300f8da711b/code/k256/Hacl.Spec.K256.GLV.Lemmas.fst#L18) lemma that states the following: 
```
[lambda](px, py) = (beta * px, py)
```
i.e., for any point on the curve `P = (px, py)`, we can replace the EC scalar multiplication by `lambda` with one modular multiplication by `beta`, where `lambda` and `beta` are primitive cube roots of unity and can be fixed for the curve. I don't plan to prove this lemma in F* :)

So, we just need to split a 256-bit scalar `k` into `k1` and `k2` s.t. `k = (k1 + lambda * k2) % q`, where `k1` and `k2` are 128-bit numbers:
```
[k]P = [(k1 + lambda * k2) % q]P = [k1]P + [k2]([lambda]P) = [k1](px, py) + [k2](beta * px, py)
```
(We also need to compute the "sign" for point `P` and new scalars `k1` and `k2` but we'll omit it for simplicity)

Using a double fixed-window method, we can save 128 `point_double`:
```
(window size w = 5)
before:
point_double: 256
point_add: 256 / 5 = 51
---------------
after:
point_double: 128
point_add: 128 / 5 + 128 / 5 + 1 = 25 + 25 + 1 = 51
```
As this method is used for `secp256k1-ecdsa-verify`, we need to compute a 4-scalar multiplication:
```
[a]P + [b]G = [a1](px, py) + [a2](beta * px, py) + [b1](gx, gy) + [b2](beta * gx, gy)
```
where we still save 128 `point_double` using a four fixed-window method (added in PR #604):
```
before:
point_double: 256
point_add: 256 / 5 + 256 / 5 + 1 = 103
---------------
after:
point_double: 128
point_add: 128 / 5 + 128 / 5 + 128 / 5 + 128 / 5 + 3 = 103
```
Note that two precomputed tables are enough for `secp256k1-ecdsa-verify`, as `[r_small]([lambda]P)` can be obtained via `[r_small]P`.

If it helps, I can copy this comment to the top of `code/k256/Hacl.Spec.K256.GLV.fst`.

msprotz:
Yup adding that comment would be great. Thanks!

msprotz:
> I don't plan to prove this lemma in F* :)

So this is the assume that you have in your code right now?

polubelova:
> So this is the assume that you have in your code right now?

In `code/k256`, there are only two assumed lemmas:
- the above [`lemma_glv`](https://github.com/hacl-star/hacl-star/blob/be4a57ccaa038d838fd1db6c31d02300f8da711b/code/k256/Hacl.Spec.K256.GLV.Lemmas.fst#L18) lemma
- the  [`lemma_order_of_curve_group`](https://github.com/hacl-star/hacl-star/blob/be4a57ccaa038d838fd1db6c31d02300f8da711b/code/k256/Hacl.Spec.K256.ECSM.Lemmas.fst#L23) lemma which a definition of `Spec.K256.q`: `[Spec.K256.q]P = point_at_infinity`

Ideally, we still need to prove that [a1, a2, b1, b2](https://github.com/hacl-star/hacl-star/blob/be4a57ccaa038d838fd1db6c31d02300f8da711b/code/k256/Hacl.Impl.K256.GLV.fst#L461) are 128-bit numbers to not call the fallback implementation.

karthikbhargavan (approved):
Good perf improvements. Let's merge it in.

msprotz (approved):

PR #607, opened 2022-09-17
Check maximum length is not exceeded for incremental hashing.

msprotz:
This is an ABI-breaking change that makes `update` return an error code, which is either Success or MaximumLengthExceeded (new).

This also relieves the OCaml bindings from having to track this through an extra reference.


msprotz:
Note that as a drive-by change I:
- fixed a notoriously flaky proof
- ditched those uu_ functions from EverCrypt_Error

victor-dumitrescu (approved):
The OCaml bindings part LGTM. Thanks!

bindings/ocaml/EverCrypt.mli
```diff
@@ -114,6 +115,8 @@ module Hash : sig
     When using the streaming interface, the total number of bytes passed through {!update} must not exceed
     - 2{^61} for SHA-224, SHA-256, and the legacy algorithms
     - 2{^125} for SHA-384 and SHA-512
+    - 2{^64} for BLAKE2s
+    - 2{^128} for BLAKE2b
```
victor-dumitrescu, line 119: I've added the mention of Blake2 to the streaming hash interface docstring

msprotz: thanks! although right now the internal length counter is 64 bits so you won't be able to hash *that* much

I will fix this in a followup PR so no need to change the doc for now

victor-dumitrescu: Yeah I remember this being the case, we were only checking up to 2^64 following the latest update to the bindings
```
assert (Z.lt !incr_len max_uint64);
```

sonmarcho (approved):

msprotz:
@hacl-star/maintainers need an approving review on this, thanks!

R1kM (approved):

PR #609, opened 2022-09-19
PowerPC integration with SHA implementation

mamonet:
Hacl integration includes the following features
- Basic vale support of PowerPC architecture.
- Memory layout and heapleat support.
- Stack functionalities (Alloc, Dealloc, Store, Load).
- QuickCode support.
- Additional basic and vector instructions that utilized in SHA implementation.
- Expand lemmas of which have use cases.
- Define bit select functions in `Vale.Def.Sel.fst`
- Add hint files for PowerPC modules.

SHA implementation now applies full proofs at every level point that follow the structure of x64 implementation. The verification time has down to 12 minutes from 50 minutes of previous version tested on Intel Core i5-10300H which is the best I can get for OpenSSL implementation. The fact the module consumes 31 vector registers all along with operand shifting makes it rather time-consuming process that seems borderline for latest optimizations.

The patch also introduces `ENABLEPOWERPC` build option to enable PowerPC-specific modules.

**Performance test**
The patch has source file `TestShaPPC64LE.cpp` to test the performance speed. I've benchmarked the output of implemented module on POWER9 processor, for 64 blocks it took `91 μs (12.2 cycles/byte)` while I can get `86 μs sec (11.5 cycles/byte)` for OpenSSL hardware-accelerated implementation.


karthikbhargavan:
Thanks! We will need a review from the vale team.

msprotz (commented):
I'll leave the review of the Vale parts to Vale experts. Some high-level comments:
- I don't think we should have conditional builds -- let's get rid of ENABLEPOWERPC
- am I correct that this Power/SHA2 primitive is not yet called anywhere?

Other than that, as far as I can tell, the changes look good, including the ones for the Makefile.

Thanks!

mamonet:
Thanks for your review!

> I'll leave the review of the Vale parts to Vale experts. Some high-level comments:
> 
> * I don't think we should have conditional builds -- let's get rid of ENABLEPOWERPC

I removed `ENABLEPOWERPC` build option.
> * am I correct that this Power/SHA2 primitive is not yet called anywhere?

You are correct. I pushed a little update that builds `ShaMainPPC64LE.ml` in Make, it's still a subject for further improvements tho. I'd suggest having this along with SCons support in another PR.

msprotz:
Ok, it's fine by me if you want to do a followup PR to add the production of sha2-256-ppc64-le.S to the build. Thanks for the comments!

vale/code/arch/ppc64le/Vale.PPC64LE.InsBasic.vaf
```diff
@@ -0,0 +1,329 @@
...
+| inout r8 | inout r9 | inout r10 | inout r11
+| inout r12 | inout r13 | inout r14 | inout r15
+| inout r16 | inout r17 | inout r18 | inout r19
+| inout r20 | inout r21 | inout r22 | inout r23
+| inout r24 | inout r25 | inout r26 | inout r27
+| inout r28 | inout r29 | inout r30 | inout r31
+;
+// 64-bit Memory operand
+operand_type Mem64(in h:heaplet, in base:reg64, inline offset:int, ghost b:buffer64, ghost index:int, inline t:taint):int;
+operand_type Stack(in base:reg_opr, inline offset:int, inline t:taint):int;
```
Chris-Hawblitzel, line 103: On a RISC architecture, do you need Mem64 and Stack operands?  If not, you could just omit these operand_types, along with the Mem64_in, Mem64_lemma, Stack_in, Stack_lemma.

Chris-Hawblitzel (approved):
Looks good to me.

vale/code/arch/ppc64le/Vale.PPC64LE.InsBasic.vaf
mamonet: Some load/store instructions have the form of `offset(base_reg)` for memory operands, that's why I kept these operand_types around. But since I flattened that form into two operands in all recurrences for simpleness, it makes sense to get rid of them.

R1kM (approved):

R1kM:
@mamonet There were some CI failures due to recent changes in F*, which were fixed in the main branch of HACL*.
Could you please merge main in your branch, which should also restart CI?

mamonet:
> @mamonet There were some CI failures due to recent changes in F*, which were fixed in the main branch of HACL*. Could you please merge main in your branch, which should also restart CI?

Done! It turns green now.

msprotz:
I'm wondering if there's a missing dependency in the Makefile because I'm getting this:

```
File "vale/code/crypto/sha/ShaMainPPC64LE.ml", line 17, characters 22-68:
17 |     ("sha256_update", Vale_SHA_PPC64LE.va_code_Sha_update_bytes_main, 4, false);
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound module Vale_SHA_PPC64LE
```

msprotz:
Ah, it was a failure to re-do the stage1 of the build on my machine... my bad!

R1kM:
@mamonet I was looking at the extraction part more deeply today because of another issue (#625)
Is the assembly syntax that you are printing OS-independent?
Also, as far as I understand, the PowerPC integration does not include yet a taint analysis to ensure the code is constant time, is that correct?

mamonet:
> Is the assembly syntax that you are printing OS-independent?

Yes, it's compatible with supported Linux distros and AIX which are the operating systems of concern.

> Also, as far as I understand, the PowerPC integration does not include yet a taint analysis to ensure the code is constant time, is that correct?

Correct. The roadmap prioritizes basic features and implementations of NIST standards over optional ones. Let me know if I miss something here.

R1kM:
Great thanks. Since Linux distros are the target, this means that we will also need to add `.section .note.GNU-stack,"",%progbits` in the generated .S (see https://github.com/cryspen/hacl-packages/issues/196), correct?

mamonet:
Thank you for handling it. I run the generated .S with `.section .note.GNU-stack,"",%progbits` appended and passed the test.

PR #611, opened 2022-09-19
Reduce C warnings in Vale Curve code

R1kM:
This PR does some minor changes to improve the quality of the generated inline assembly code from Vale for Curve25519, aiming at removing some warnings raised by C compilers.

In particular, the return value of `add1` is only used for spec purposes, and is not used later in the Curve code, which raises an unused variable warning. To avoid this, the return type is marked (in F*) as Ghost.erased, and the Vale wrappers are changed accordingly.

This PR also changes the trusted Vale inline assembly printer to use the `__asm__` keyword instead of `asm`, following https://stackoverflow.com/questions/35131350/error-asm-undeclared-first-use-in-this-function.

Note: this PR should be reviewed and merged after #610 , which should significantly reduce the diff on generated code in dist/


dist/c89-compatible/Hacl_Curve25519_64.c
```diff
@@ -340,11 +337,11 @@ static void store_felem(uint64_t *b, uint64_t *f)
   uint64_t o2;
   uint64_t o3;
   f[3U] = f30 & (uint64_t)0x7fffffffffffffffU;
-  carry0 = add_scalar0(f, f, (uint64_t)19U * top_bit0);
+  add_scalar0(f, f, (uint64_t)19U * top_bit0);
   f31 = f[3U];
   top_bit = f31 >> (uint32_t)63U;
   f[3U] = f31 & (uint64_t)0x7fffffffffffffffU;
-  carry = add_scalar0(f, f, (uint64_t)19U * top_bit);
```
R1kM, line 347: These instructions were the ones previously flagged as unused variables

msprotz (approved):
Looks good to me. I'm going to see if I can push a krml change to sort libevercrypt.def and ctypes.depend alphabetically to avoid un-necessary diffs in the future

parno (approved):
Looks reasonable to me.

PR #615, opened 2022-09-23
Several cleanups related to Blake2, hashes, etc.

msprotz:
- Blake2 does not follow the Merkle-Damgard padding scheme -- there is no point in defining padding for Blake2... it's unused!
- Lots of duplicated definitions are now gone.
- A lot of Blake2-specific helpers had no reason to be there. Remove them.
- Remove a lot of implementations that were unused.

(This is in preparation for the addition of SHA3 to the agile hash infrastructure.)

PR #617, opened 2022-09-26
Addition of SHA3 to agile hashing & streaming infrastructure

msprotz:
This is on top of #615 so make sure to read the diff in there first.

This touches a lot of areas in the code:
- makes maximum length optional for hashes
- fixes every single use of hashes accordingly (such as HPKE, all of the signature algorithms that need to hash their input data first, HMAC, HKDF, etc.)
- adds a few subtle restrictions: `update_last` now takes *at most* a block's worth of data: this makes the proof of "sha3 is incremental" much easier because there is now a 1:1 correspondence between update_multi / update_last and repeat_blocks process_block process_last_block... the previous behavior (`update_last` accepts any amount of data) was only used by HMAC, and there was no deep reason for that... so this is one of the things I fixed ahead of time in #615 
- another subtle restriction: the `finish` function is now allowed to modify the internal state... SHA3 does this, and this does not affect any other proof (the streaming functor takes great care of copying the internal state *before* finishing), so this behavior is now allowed for all hashes
- long, technically difficult proof of "sha3 is incremental" which should not have been hard except for a type equality that was strikingly difficult to establish (see https://github.com/hacl-star/hacl-star/blob/protz_sha3/specs/lemmas/Spec.Hash.Incremental.fst#L722) -- I wonder if this is because the type equality in words_state is written using `=` instead of `==` (a mistake for which I would take full responsibility!)
- a million other small fixes
- instantiation of the streaming functor for sha3
- generalization of some hash helpers to cover sha3
- addition of sha3 to agile hash infrastructure.

Looking forward:
- this is only for sha3-256, need to add all other three variants
- after the cleanups I mentioned, I believe the proof that blake2 in hash mode is incremental ought to be much easier! surely, there is a way to show that `repeati` is equivalent to `update_multi` and perform this proof in the abstract... TBD
- I think there are a couple fragile proofs so let's see if CI is happy or if I need to hone some of these more
- right now the incremental API has an arbitrary limit set to 2^64 - 1 ... I could either add a layer of parameterization to the streaming functor to disable length tracking (sha3 doesn't need it), or bump that internal length to a 128-bit integer... TBD


msprotz:
A couple proofs broke for no good reason, so I did some unrelated proof strengthening, notably in Spec.Agile.HPKE. Let's see if this passes CI.

msprotz:
@pnmadelaine is there any way to tell which task was running when this job timed out?

dist/gcc-compatible/EverCrypt_Hash.h
```diff
@@ -35,7 +35,9 @@ extern "C" {
 #include "krml/internal/target.h"
 
 
+#include "Lib_Memzero0.h"
```
R1kM, line 38: Why is this addition needed?

specs/Spec.Agile.HPKE.fst
```diff
@@ -180,8 +180,9 @@ val encap:
   Tot (option (key_kem_s cs & key_dh_public_s cs))
 
 #restart-solver
-#set-options "--z3rlimit 600 --fuel 0 --ifuel 2"
+#set-options "--z3rlimit 100 --fuel 0 --ifuel 0"
 let encap cs skE pkR =
+  let _ = allow_inversion Spec.Agile.DH.algorithm in
```
R1kM, line 185: Great to see typechecking of this spec simplified

specs/Spec.Hash.Definitions.fst
```diff
@@ -36,33 +36,42 @@ type hash_alg =
...
   | MD5 | SHA1 | SHA2_224 | SHA2_256 | SHA2_384 | SHA2_512 -> true
   | _ -> false
 
 let sha2_alg = a:hash_alg { is_sha2 a }
+let sha3_alg = a:hash_alg { is_sha3 a }
 let blake_alg = a:hash_alg { is_blake a }
 let md_alg = a:hash_alg { is_md a }
 
- inline_for_extraction
+inline_for_extraction
```
R1kM, line 69: Why no `noextract` here? Same for following definitions. This predates this PR, but it would be good to be consistent throughout the definitions in this module (this is clearly non-blocking given the current size of this PR)

R1kM (approved):
I left a few minor comments, but this looks good to me otherwise. Congrats on this huge piece of work!

code/hash/Hacl.Hash.SHA3.fsti
```diff
@@ -0,0 +1,21 @@
+module Hacl.Hash.SHA3
+
+// Just a few combinators needed to instantiate the streaming functor.
+
+open Spec.Hash.Definitions
+open Hacl.Hash.Definitions
+
+inline_for_extraction noextract
```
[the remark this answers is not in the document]
msprotz: just didn't want another file to appear in the code-gen so I marked everything as inline... another difference is that these functions just "redirect" to Hacl.SHA3, but with the agile hash signatures (so worth inlining since they just call absorb/squeeze/etc.); for Hacl.SHA2, the actual code is in there, so not a good idea to inline

code/rsapss/Hacl.Impl.RSAPSS.MGF.fst
```diff
@@ -23,11 +23,15 @@ val hash_len: a:Hash.algorithm{S.hash_is_supported a} -> Tot (res:size_t{v res =
 let hash_len a =
   Hacl.Hash.Definitions.hash_len a
 
+inline_for_extraction noextract
+let max_input_length = Hash.max_input_length
+
+let less_than_max_input_length = Spec.Hash.Definitions.less_than_max_input_length
```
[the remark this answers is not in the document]
msprotz: yeah I think it's only working out because of bundling... good point

specs/Spec.Agile.HPKE.fst
msprotz: 🙇

specs/Spec.Hash.Definitions.fst
msprotz: I generally don't bother with noextract in specs because all of those are eliminated anyhow via `-bundle` in the final krml invocation

dist/gcc-compatible/EverCrypt_Hash.h
msprotz: Good question, I don't know

msprotz:
A report after spending too many hours on this proof.
- The module was opening FStar.Integers, which was a terrible idea, because after 2.5h z3 was still incapable of dealing with a trivial assertion that fatefully referred to an overloaded operator
- I switched most of the proof to a nl-arith-free style until realizing what the problem was (after digging into the smt encoding...)

The proof now goes through smoothly. I wrote down the point where it no longer is nlarith-free. It might be interesting to be rigorous and write the entire function without resorting to nlarith reasoning, but one day on a single proof is enough for me.

I also updated this branch to merge in main and fixed up ppc64 sha to use the updated definitions.

msprotz:
Ok this is green, just need another review for this. Let's merge it soon as it might conflict with future work. Thanks!

dist/gcc-compatible/EverCrypt_Hash.h
msprotz: I mean it should go into the .c file, or at the very least the internal header... I can investigate later on if it's important

R1kM: I think it's fine

specs/lemmas/Spec.Hash.Incremental.fst
```diff
@@ -596,10 +596,291 @@ let md_is_hash_incremental
...
+    hash_incremental a input;
+  (S.equal) { } (
+    let s = Lib.Sequence.create 25 (u64 0), () in
+    let bs, l = split_blocks a input in
+    let s = update_multi a s bs in
+    let s = update_last a s (S.length bs) l in
+    finish a s
+  );
+  (S.equal) { } (
+    let s = Lib.Sequence.create 25 (u64 0), () in
```
polubelova, line 620: do you really need to duplicate so much code for each `calc` step?
```
 let s = Lib.Sequence.create 25 (u64 0), () in
 let rateInBytes = 1088 / 8 in
 let delimitedSuffix = byte 0x06 in
```
can these proofs be used for other instances of sha3? if so, should we make them generic?

specs/lemmas/Spec.Hash.Incremental.fst
```diff
@@ -596,10 +596,291 @@ let md_is_hash_incremental
      }
 #pop-options
 
-let hash_is_hash_incremental (a: hash_alg) (input: bytes { S.length input <= max_input_length a }):
+let sha3_is_incremental1
```
polubelova, line 599: should we split this file? e.g., each new file will correspond to one family of hash functions, i.e., blake2, sha2, sha3, etc.

polubelova (approved):
Looks great, thanks!

polubelova:
Btw, could you please add a C test for sha3-256 that uses a new streaming API?

specs/lemmas/Spec.Hash.Incremental.fst
msprotz: > do you really need to duplicate so much code for each calc step?
it took me a lot of steps to get to the point where I understood what the problem was (the lack of type equality between `words_state SHA3_256` and `Spec.SHA3.state` so I think I broke it down into *way* too many steps!

> if so, should we make them generic?
I structured things in such a way that I can easily make them generic... but in a followup PR :-)

dist/gcc-compatible/libevercrypt.def
```diff
@@ -254,7 +256,9 @@ EXPORTS
   Hacl_Streaming_SHA2_finish_256
   Hacl_Streaming_SHA2_free_256
   Hacl_Streaming_SHA2_create_in_384
+  LowStar_BufferOps_op_Bang_Star__Hacl_Streaming_Functor_state_s__uint64_t____
```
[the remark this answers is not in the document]
msprotz: I think this one does not appear in the generated C code but it's the generation of the .def file that is buggy -- can you file a bug for krml?

specs/lemmas/Spec.Hash.Incremental.fst
msprotz: sure, I can do that, it might require me to friend a module or two, but ok!

msprotz:
I'll add a test on the next PR (#626) because it needs the refresh of dist/

msprotz:
For the record, after days of debugging, it turns out that `Spec.Hash.Lemmas.update_multi_associative` would occasionally send z3 off the rails (why? not sure). This would only happen randomly, sometimes in batch mode, sometimes in interactive mode, sometimes not upon the first try, but upon the second, etc. "Catching" the bad query took a lot of effort, because merely adding --log_queries would suddenly turn the failing query into a succeeding one.

Eventually, I "caught" the bad query. Ran z3 on it overnight -- it took several hours (with rlimit 100) for z3 to terminate and even tell me (via qi profile) what was the faulty lemma. After that, it was merely the usual discrepancy between platforms, F* versions, CI machine vs dev machine, etc. and a little bit of bumping rlimits until this worked out.

Earlier successes on this branch were probably that someone had managed to verify the file *then* write a hint file, and then CI would get a hint hash match and skip the query.

This particular issue wth `update_multi_associative` is on top of several other issues debugged earlier in the branch, such as FStar.Integers standing in the way of proving very simple equalities, non-linear reasoning with modulo facts that were hindering the proof, and so on.

franziskuskiefer:
@msprotz you added limits here with type `u128`.
This is problematic when using the APIs in FFI because `u128` might not have a well-defined ABI. This makes it for example impossible to use stateful SHA384/512 in Rust now.

msprotz:
Which function in the public API is now using uint128?

franziskuskiefer:
> Which function in the public API is now using uint128?

https://github.com/hacl-star/hacl-star/blob/51a72a953a4ee6f91e63b2816ae5c4e62edf35d6/dist/gcc-compatible/Hacl_Hash_SHA2.h#L64-L78

karthikbhargavan:
We could replace the counter with two uint64s if that is more portable?

msprotz:
Ugh. Thanks. Two thoughts:
- this argument is unused so conceivably we should be able to ditch that `prev_len`... can you file a separate issue for that?
- you should be using the streaming API directly, rather than using the low-level update/update_last API, shouldn't you?

karthikbhargavan:
Indeed, @franziskuskiefer does this issue matter if the functions are never used or documented externally?

franziskuskiefer:
> * you should be using the streaming API directly, rather than using the low-level update/update_last API, shouldn't you?

If the API shouldn't be used, it should be removed.

It can of course be ignored (as it is right now). But this needs to be done explicitly.

I filed #673 to fix this. But maybe it should be changed to "remove the API"?

msprotz:
Thanks for filing. I think it's mostly a historical thing, as the streaming API didn't come until later.

franziskuskiefer:
@msprotz by only adding SHA3 256 here you made it use `8` as `Spec_Hash_Definitions_hash_alg`, which is a little unexpected. Do you want to keep this order or shall it be changed (at the cost of breaking the API)?

msprotz:
What did you expect for the constant associated to SHA3_256? My assumption was always that because we have macro names, the actual constant values don't matter...?

franziskuskiefer:
Generally yes, but in bindings like Rust we had all these already to make it usable. So we have to break the API there for anyone using the raw values. Also, it's generally not safe to expect no one is using the literal values instead of macros I'm afraid.

msprotz:
Sorry, I don't understand what change you're proposing, and I still don't understand why `8` specifically is unexpected. Do you have a better numbering scheme in mind?

Do you have a link to the Rust bindings and the comparable piece of code?

PR #618, opened 2022-09-26
Fix extraction for bignum loading and storing functions

polubelova:
This PR avoids VLA and extra copy for loading and storing functions when a size of bignum is known.


msprotz (approved):
Very cool, but you can do even better by synthesizing is_known_length automatically. Here's an example using regular ints, you'll have to make sure it works in your case, but it probably will:

```
module Test

open FStar.Tactics

let is_constant (x: term): Tac unit =
  match inspect_ln x with
  | Tv_Const _ -> exact (`true)
  | _ -> exact (`false)

let test (x: int) (#[is_constant (quote x)] meta_flag: bool) =
  meta_flag

let unknown (x: int) = test x
let known = test 0
```

extracting this code:

```
let (unknown : Prims.int -> Prims.bool) = fun x -> test x false
let (known : Prims.bool) = test Prims.int_zero true
```

so the bool magically turns into `true` when the previous argument is a constant, and `false` otherwise. In the case of bignum, you might need `quote (norm [ primops ] x)` to make sure the multiplication is reduced and the term *does* look like a constant by the time it is inspected

not necessary to implemented this! just a suggestion in case you'd like something more automatic

polubelova:
Thanks @msprotz for the suggestion! I'll play with it later and create another PR as it didn't work out the first time :/

msprotz:
Yeah if you think that it's worth having I can help. Let me know.

R1kM (approved):

PR #619, opened 2022-09-27
Upstream blake2 changes

franziskuskiefer:
This PR upstreams Cryspen blake2 changes that
* optimise loop unrolling for blake2 to speed it up
* use optimised blake2 in evecrypt hash


msprotz:
This is a bug in evercrypt-external-headers which I believe is removed in another PR... let me see if I can find it

msprotz:
Unfortunately I merged the krml alignment changes too early, meaning that this PR will only be green when #622 is merged, too.

code/blake2/Makefile
```diff
@@ -4,33 +4,43 @@ HACL_HOME=$(realpath ../..)
...
-	  -funroll-loops 8 \
+	  -funroll-loops 12 \
 	  -fbuiltin-uint128 \
 	  -tmpdir dist \
+	  -skip-compilation \
 	  -ccopts -std=gnu11,-g,-O3 \
-	  -I$(KRML_HOME)/include \
-	  -I$(HACL_HOME)/lib/c" \
-	  -add-include '"krml/internal/target.h"' \
+	  -I "../../../karamel/include" \
```
R1kM, line 32: This seems to assume that karamel is installed in the same folder as hacl-star. That seems like a strong assumption (which doesn't hold on my machine for instance), what is the rationale for dropping the use of KRML_HOME?

code/blake2/Makefile
```diff
@@ -4,33 +4,43 @@ HACL_HOME=$(realpath ../..)
...
+	  -I "../../lib/c" \
+	  -add-include '"kremlin/internal/target.h"' \
```
R1kM, line 34: Should it be karamel/ ?

R1kM (approved):

msprotz:
These remarks are all for @karthikbhargavan -- I didn't understand the changes to the local Makefile either

msprotz:
@karthikbhargavan will:
- merge upstream into this branch
- perform the requested followup changes: move lemmas to lib and fixup local makefile

code/blake2/Makefile
msprotz: for the record the previous form with $(KRML_HOME) is the correct one, I believe

providers/evercrypt/fst/EverCrypt.Hash.fst
```diff
@@ -170,7 +268,13 @@ let init #a s =
   | SHA2_384_s p -> Hacl.Hash.SHA2.init_384 p
   | SHA2_512_s p -> Hacl.Hash.SHA2.init_512 p
   | Blake2S_s p -> let _ = Hacl.Hash.Blake2.init_blake2s_32 p in ()
+  | Blake2S_128_s _ p ->
+      if EverCrypt.TargetConfig.hacl_can_compile_vec128 then
```
sonmarcho, line 272: Why such checks (same question for other functions like `update2`)? There is no way we created a `Blake2S_128_s` state if we don't have the proper flag. And I guess if we still managed to do so with some hacking, isn't it better to fail by trying to use illegal instructions than having the functions silently behave as no-ops?

msprotz: this is to prevent a link-time error; without this check, EverCrypt_Hash.o contains a reference to the symbol Blake2s_128_init; if you are on a platform for which vec128 is not available, then Blake2s_128_init is not compiled and you get a link-time error

with this check, there is an #ifdef surrounding the reference to Blake2s_128_init and all is well

sonmarcho: Indeed, I had forgotten that those checks were extracted to #ifdef

sonmarcho (approved):

msprotz:
@karthikbhargavan can you merge main in this branch?

lib/Lib.ByteSequence.fst
```diff
@@ -834,3 +834,40 @@ let rec nat_from_intseq_le_public_to_secret #t len b =
     nat_from_intseq_le_lemma0 (slice b_secret 0 1);
     eq_intro (slice (map secret b) 1 len) (map secret (slice b 1 len));
     () end
+
+
+let lemma_uints_to_bytes_le_sub #t #l #len s i =
```
polubelova, line 839: You could simplify a `lemma_uints_to_bytes_le_i_j`lemma by having only one last argument:
```
let lemma_uints_to_bytes_le_sub #t #l #len s i =
  let lemma_uints_to_bytes_le_i_j (j : size_nat {j < numbytes t}):
      Lemma(index (uints_to_bytes_le #t #l s) (i * numbytes t + j) == index (uint_to_bytes_le (index s i)) j) =
    index_uints_to_bytes_le #t #l #len s (i*numbytes t + j);
    assert (index (uints_to_bytes_le #t #l s) (i*numbytes t + j) ==
           index (uint_to_bytes_le (index s i)) j) in

  Classical.forall_intro lemma_uints_to_bytes_le_i_j;
  eq_intro (sub (uints_to_bytes_le #t #l s) (i * numbytes t) (numbytes t)) (uint_to_bytes_le (index s i))
```

lib/Lib.ByteSequence.fst
```diff
@@ -834,3 +834,40 @@ let rec nat_from_intseq_le_public_to_secret #t len b =
...
+    index_uints_to_bytes_le #t #l #len s (i*numbytes t + j);
+    assert (index (uints_to_bytes_le #t #l s) (i*numbytes t + j) ==
+           index (uint_to_bytes_le (index s i)) j) in
+  
+  Classical.forall_intro (lemma_uints_to_bytes_le_i_j #t #l #len s i);
+  eq_intro (sub (uints_to_bytes_le #t #l s) (i * numbytes t) (numbytes t)) (uint_to_bytes_le (index s i))
+
+
+let lemma_uints_to_from_bytes_le_preserves_value #t #l #len s =
+  let lemma_uints_to_from_bytes_le_preserves_value_i
```
polubelova, line 856: Same here

lib/Lib.ByteSequence.fst
karthikbhargavan: Done

lib/Lib.ByteSequence.fst
karthikbhargavan: Done

msprotz:
@karthikbhargavan your remote was out of date, can you do this:

```
git checkout cryspen_blake2
git fetch git@github.com:hacl-star/hacl-star
git merge FETCH_HEAD
```

alternatively, @franziskuskiefer if you want to update this branch, go ahead, that'd allow to move forward and merge this... thanks

franziskuskiefer:
>  @franziskuskiefer if you want to update this branch, go ahead, that'd allow to move forward and merge this... thanks

I just clicked the "update branch" button 🙂

msprotz:
Ok, well, looks like this definitely breaks HMAC. Need to investigate.

PR #620, opened 2022-09-27
Upstream fix of inlining issues

franziskuskiefer:
This PR upstreams Cryspen changes that fix build errors because functions were inlined across compilation units.


msprotz (approved):
Looks good, thanks.

polubelova (approved):
Thanks!

PR #621, opened 2022-09-27
Upstream MLCRYPTO removal

franziskuskiefer:
This PR upstreams Cryspen changes that remove `MLCRYPTO` in favour of linking against system OpenSSL when necessary.


Makefile
```diff
@@ -1029,6 +1022,7 @@ dist/test/c/%.c: $(ALL_KRML_FILES)
 	  -library Hacl.P256,Hacl.K256.*,Hacl.Impl.*,EverCrypt.* \
 	  -fparentheses -fcurly-braces -fno-shadow \
 	  -minimal -add-include '"krmllib.h"' \
+	  -add-include '"libintvector.h"' \
```
R1kM, line 1025: I imagine that, since this is only for dist/test/, that's not much of a problem, but why is it necessary to always include the libintvector header?

Makefile
```diff
@@ -1054,18 +1048,15 @@ compile-%: dist/Makefile.tmpl dist/configure dist/%/Makefile.basic | copy-krmlli
...
-LDFLAGS 	+= -L$(OPENSSL_HOME)
-
 CFLAGS += -Wall -Wextra -g \
   -Wno-int-conversion -Wno-unused-parameter \
-  -I$(OPENSSL_HOME)/include \
-  -O3 -march=native -mtune=native -I$(KRML_HOME)/krmllib/dist/minimal \
+  -O3 -I$(KRML_HOME)/krmllib/dist/minimal \
   -I$(KRML_HOME)/include -Idist/gcc-compatible
 
+ifneq ($(shell uname -m),arm64)
```
R1kM, line 1056: Why is it needed to prevent the use of these flags on arm?

Makefile.openssl
```diff
@@ -0,0 +1,9 @@
+# Exporting proper linking flags -- used by Makefile and tests/Makefile
+
+ifeq ($(shell uname),Darwin)
```
R1kM, line 3: Why do you only use OpenSSL on Mac?

R1kM (commented):
I left a few comments directly in the files.
I'm not sure I fully understand when linking against OpenSSL is needed here, as far as I understand, the current Makefile changes always adds OpenSSL when compiling C code on Mac exclusively. Could you clarify when OpenSSL is necessary?

Makefile
msprotz: this is because the test contains a copy of the struct definition for the evercrypt hash internal state, which contains a blake2b_256 state, which refers to the type Lib_IntVector_Vec256, so this means that this type needs to be in scope for the test to compile

Makefile
msprotz: they are not recognized on arm, the compiler will error out

Makefile.openssl
msprotz: on Linux, `-lcrypto` works unconditionally; on OSX, OpenSSL is oftentimes installed via `brew` and the path to openssl includes and libraries is *NOT* on the default LD_LIBRARY_PATH and CFLAGS path, so on OSX (and on OSX only) I perform some "auto-detection" that attempts to find a suitable path to the OpenSSL install directory

msprotz:
OpenSSL is only needed for the C *tests* -- for performance comparisons, specifically.

On OSX, I try to perform some smart auto-detection; on Linux, I assume `-lcrypto` and `#include <openssl/foobar.h>` work without customizing LD_LIBRARY_PATH and CFLAGS, respectively.

This is where we link tests against openssl: https://github.com/hacl-star/hacl-star/blob/main/tests/Makefile#L94

R1kM (approved):

polubelova (approved):
Thanks!

Makefile
R1kM: We discussed this offline, this addition is actually related to another PR, which adds Blake2 to Evercrypt's hashes and requires libintvector.
The problem here is that this additional header is now included in all C tests generated from F*, including for instance CSHAKE or ECDSA.
However, since this is only for test purposes, we think it's fine to keep the makefile simple and avoid special-casing for hash tests.

PR #628, opened 2022-10-01
Split Vale PowerPC SHA implementation in several files

R1kM:
@polubelova reported that the Vale implementation of SHA for PowerPC added in #609 was taking a long time to verify, particularly compared to all other files in the codebase.
While there might be room for optimizing and stabilizing the proofs, this PR provides a more immediate solution, by splitting the implementation into different files, enabling more parallel builds.
Beyond splitting files, this PR does not introduce any changes to the implementation.

Prior to this change, the 5 slower files to verify on the CI machine were as follows:
[VERIFY] Vale.SHA.PPC64LE.fst, 23:00.45
[VERIFY] Hacl.Impl.HPKE.fst, 8:23.17
[VERIFY] Vale.AES.X64.AESGCM.fst, 7:04.94
[VERIFY] Spec.Agile.HPKE.fst, 6:05.79
[VERIFY] Hacl.Meta.HPKE.fst, 6:15.20

We now have:
12:23.48 Vale.SHA.PPC64LE.Loop.fst
11:19.31 Vale.SHA.PPC64LE.Rounds.Core.fst
7:35.90 Vale.AES.X64.AESGCM.fst
7:30.46 Hacl.Impl.HPKE.fst
6:06.35 Spec.Agile.HPKE.fst


polubelova (approved):
Thanks!

Chris-Hawblitzel (approved):
Is NBE faster? ```unfold let normal (x:Type0) : Type0 = norm [nbe; iota; zeta; simplify; primops; delta_attr [`%va_qattr]; delta_only normal_steps] x```

R1kM:
I tried running CI with nbe (https://everest-ci.paris.inria.fr/build/1846/download/1/stats.txt) and without nbe (https://everest-ci.paris.inria.fr/build/1835/download/1/stats.txt). Activating nbe might be marginally faster, but it is unclear how much of it is just CI noise.
The CI build also doesn't display stats about memory consumption, only about verification time. Is there any documentation about what the impact of NBE on RAM usage might be?

Chris-Hawblitzel:
I don't know about RAM usage.  It looks like nbe had little effect on the X64 Vale files (based on modifying `vale/code/arch/x64/Vale.X64.QuickCodes.fsti` ).  However, @msprotz reported a large speedup for the PPC Vale files (based on modifying `vale/code/arch/ppc64le/Vale.PPC64LE.QuickCodes.fsti` ).

msprotz:
The speedup would have to be reproduced in controlled conditions, e.g. without using hints. That's what I observed on my Linux machine, but there were multiple processes running.

The CI machine cannot be trusted because so many jobs run in parallel.

R1kM:
Indeed, while the environment is not fully controlled, NBE for Vale PowerPC seems faster on CI (https://everest-ci.paris.inria.fr/build/1855/download/1/stats.txt, see that verification of non-Vale.PPC files remains in line with previous builds, while verification time decreases for Vale.PPC).
I'll enable nbe for both, happy to revert if anybody is observing performance issues locally.

mamonet:
Thanks for your observations. The change significantly decreases the verification time of PPC SHA modules.

> there might be room for optimizing and stabilizing the proofs
I've already started with implementing AES-GCM for PowerPC, I could get back to SHA implementation for more optimizations and stabilizations once I'm done with AES-GCM and ECC.

PR #632, opened 2022-10-03
fix github workflows

pnmadelaine:
This is a work around the github limitation that a workflow can't be triggered by another one.
The dist-ci workflow was not triggered on PRs generated by the hints-and-dist workflow.
See https://github.com/peter-evans/create-pull-request/issues/48.


polubelova (approved):
Thanks!

franziskuskiefer:
FYI: you can trigger workflows from other workflows. see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_call

PR #633, opened 2022-10-03
Try bumping minimum version of iOS for CI

R1kM:
Necessary to restore the macOS build action since the addition of align_alloc in Krml. Done with @protz


polubelova (approved):
Thanks!

pnmadelaine (approved):

Issue #625, opened 2022-09-29
Vale: add .note.GNU-stack


franziskuskiefer:
*See https://github.com/cryspen/hacl-packages/issues/196 for the original report*

It looks like `ld` will require `.note.GNU-stack` in future version and warns about it since 2.39. All vale code for Linux should therefore get a `.note.GNU-stack` section (see original issue for a possible fix).

@R1kM could you have a look at this?

PR #635, opened 2022-10-05
Cleanup of some warnings

R1kM:
This PR does some minor changes to the codebase to remove some warnings raised by F*.

In particular, it gets rid of:
* The "definition is recursive but not used in body" warning
* The "name is not used and will be ignored" warning
* The "argument is effectful, consider let-binding it" warning
* The "special treatment of syntax @ is deprecated, add an open FStar.List.Tot" warning

This PR should not have any impact on the generated code or on verification performance.

Note, it currently builds on top of #625 and should be merged after.


polubelova (approved):
Thanks!

PR #636, opened 2022-10-06
Avoid a needless proof step and solve it with a tactic

nikswamy:
The additional step in the calc is unnecessary ... and the goal can be solved directly with a simple tactic, rather than relying on SMT to complete it.


msprotz (approved):

polubelova (approved):
Thanks!

PR #638, opened 2022-10-06
Exposing streaming hash functions api

Frosne:
I would like to expose the streaming API for hash functions. 

Thanks.


polubelova:
btw, you can update C code for `dist/mozilla` by downloading the files from https://everest-ci.paris.inria.fr/build/2024 (`dist.tar`)

msprotz (approved):

R1kM:
@Frosne It seems this addition leads to the extraction of more files than expected to dist/mozilla (see https://github.com/hacl-star/hacl-star/pull/640/files#diff-206457368a9d1a940e09f56e2fdc90c542f10f53355649d1df02a95890391775)
One way to solve this is by adding more extracted files, in particular some of the hashes to the mozilla dist. Would that be okay for you?

PR #644, opened 2022-10-10
Extract some bignum functions as static inline

polubelova:
This PR is supposed to fix the performance regression for P256 (PR #563) that was introduced by PR #620.


franziskuskiefer (commented):
Thanks for the changes.
This gets us close to the previous performance and is certainly worth doing.
I'd be fine to keep it like this until the larger rework happens.

### Benchmarks
This change:

```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
P256_SHA256_ECDSA_Sign       150596 ns       150584 ns         4645
P256_SHA256_ECDSA_Verify     146344 ns       146345 ns         4783
P256_ECDH                    146189 ns       146191 ns         4780
```

Before https://github.com/hacl-star/hacl-star/pull/563
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
P256_SHA256_ECDSA_Sign       149053 ns       149054 ns         4695
P256_SHA256_ECDSA_Verify     143373 ns       143374 ns         4797
P256_ECDH                    144687 ns       144689 ns         4837
```

msprotz (approved):

R1kM (approved):

msprotz:
This is done already for the result of malloc when in c89 mode. If you are positive that this only happens for MSVC, when in C++ mode, I can expose an additional option that we use when generating dist/gcc-compatible. Let me know!

franziskuskiefer:
> This is done already for the result of malloc when in c89 mode. If you are positive that this only happens for MSVC, when in C++ mode, I can expose an additional option that we use when generating dist/gcc-compatible. Let me know!

g++ and clang++ both require this as well for `malloc` and `alloca`.

```
error: cannot initialize a variable of type 'uint8_t' (aka 'unsigned char') with an rvalue of type 'void *
```

So we'll always need this for C++. We don't see the issue here because we use VLA on non-MSVC.

franziskuskiefer:
@msprotz can you expose the options to do this? Or just always do it in code that's going into header files?

msprotz:
What are your thoughts @franziskuskiefer? Cast all of the allocations or just in header files? I think it's safer to cast all allocations because it means that a large chunk of our code can be compiled out of the box with a C++ compiler. (We do rely a bunch on C11 struct literals, which are not valid C++, so not *all* HACL code can be compiled by a C++ compiler. But we could have krml decompose those at the expense of readability, but that's another discussion.)

franziskuskiefer:
> What are your thoughts @franziskuskiefer? Cast all of the allocations or just in header files? I think it's safer to cast all allocations because it means that a large chunk of our code can be compiled out of the box with a C++ compiler.

Yeah that's probably the easiest. But other code (in .c files) shouldn't be seen by a C++ compiler so it's not really an issue there. There are religious discussions around casting the result of `malloc`. But I hink being explicit about what type of pointer it is makes it more descriptive and makes the code easier to read.

> (We do rely a bunch on C11 struct literals, which are not valid C++, so not _all_ HACL code can be compiled by a C++ compiler. But we could have krml decompose those at the expense of readability, but that's another discussion.)

This sort of code should just never end up in a header 😄

PR #647, opened 2022-10-17
OCaml API: finer-grained passing of buffers in NaCl functions

victor-dumitrescu:
This PR makes the following changes to the `Detached` functions of the OCaml NaCl API:
- Since the use case supported by these functions is to use the same buffer for both plaintext and ciphertext, thus allowing in-place encryption and decryption, the `pt` and `ct` arguments are replaced by a single `buf` argument.
- Following a feature request, users can now optionally only pass portions of `buf` through optional arguments `offset`, which gives a starting position in `buf`, and `len`, which specifies how many bytes to take. This saves a copy on the OCaml side if the buffer needs to contain additional data.

The NaCl tests have also been expanded to test this additional functionality and have been rewritten using the Alcotest library.

I plan to create a follow-up PR to rewrite all test to use Alcotest.

### DIFF
--- a/bindings/ocaml/Hacl.ml
+++ b/bindings/ocaml/Hacl.ml
@@ -255,7 +255,7 @@ module NaCl = struct
     assert (C.size tag = 16);
     assert (C.size n = 24);
     assert (C.disjoint tag buf);
-    assert (C.disjoint n buf);
+    assert (C.disjoint n buf)
   module Noalloc = struct
     let box_beforenm ~pk ~sk ~ck =
       check_key_sizes pk sk;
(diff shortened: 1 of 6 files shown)


msprotz (approved):
Looks good to me, but this is an API-breaking change. Do we care about maintaining backwards-compatibility? I guess we could add `box_inplace` (the new in-place API), then implement `box` as `copy + box_inplace`, but is it worth it?

bindings/ocaml/Hacl.ml
```diff
@@ -293,30 +290,47 @@ module NaCl = struct
         get_result @@ hacl_NaCl_crypto_secretbox_open_easy (C.ctypes_buf pt) (C.ctypes_buf ct) (C.size_uint32 ct) (C.ctypes_buf n) (C.ctypes_buf key)
     end
     module Detached = struct
-      let box ~pt ~n ~pk ~sk ~ct ~tag =
+      let get_sub_buffer buf offset len =
```
polubelova, line 293: Usually, you need to check for `0 <= offset && 0 <= len && len <= C.size buf - offset`, as, for example, `offset + len` or `C.size buf - offset` may overflow in some cases.

franziskuskiefer (approved):
Generally looks good to me with @polubelova's comment addressed.
I think using new `_inplace` functions as Jonathan mentions would be good to stay backwards compatible.

victor-dumitrescu:
Regarding the fact that this is a breaking change, my thinking was that the use case enabled by the existing `Noalloc.Detached` interface and eliminated here would be in-place encryption in 2 different buffers for plaintext and ciphertext _and_ not wanting the tag to be part of the ciphertext buffer (otherwise, `Noalloc.Easy` is still available). I'm not sure it's worth adding another set of functions to an already large API to maintain this specific use case, but do let me know if you think this is  actually useful. In Tezos, for example, this API is used by passing the same buffer as `pt` and `ct`.

franziskuskiefer:
> Regarding the fact that this is a breaking change

I'm fine with breaking APIs as long as you don't expect any backlash from other users of the package.
I think generally adopting semver here in future should be the goal. This would mean breaking changes require major version bump. But the mantra is usually anything goes pre-1.0. So I'm fine with this.

bindings/ocaml/Hacl.ml
victor-dumitrescu: Thanks, updated

victor-dumitrescu:
Yeah I guess there's another discussion to be had about versioning and whether we can start labelling packages as 1.x.x

msprotz:
Yeah don't think it's a big deal here, but once there's an official release, let's start to be more careful with APIs, and let's do a review to make sure the APIs have the shape we want. Ok to merge for me.

msprotz:
Cross-referencing with #648 since ocaml bindings still aren't under CI, so let's hope that this PR was right!

(My bad for not remembering we don't do OCaml testing on CI.)

Issue #540, opened 2022-03-22
Addition of SHA3 to Mozilla breaks build on OSX


msprotz:
Hi Natalia,

I just noticed that your recent addition of SHA3 to dist/mozilla breaks the build on OSX with this:

```
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: libevercrypt.a(Vale.o) has no symbols
Undefined symbols for architecture x86_64:
  "_Lib_Memzero0_memzero", referenced from:
      _Hacl_Impl_SHA3_state_permute in Hacl_SHA3.o
      _Hacl_Impl_SHA3_loadState in Hacl_SHA3.o
      _Hacl_Impl_SHA3_storeState in Hacl_SHA3.o
      _Hacl_Impl_SHA3_absorb in Hacl_SHA3.o
      _Hacl_Impl_SHA3_squeeze in Hacl_SHA3.o
      _Hacl_SHA3_sha3_224 in Hacl_SHA3.o
      _Hacl_SHA3_sha3_256 in Hacl_SHA3.o
      ...
```

it appears that dist/mozilla is missing Lib_Memzero0.{c,h}. Do you know why this wasn't caught earlier?

Thanks,

Jonathan

Frosne:
In this case the dist/mozilla will have both, Lib.Memzero.c and the header? Is it possible to have just the header? We implement secure memory erasure slightly differently than HACL* :)

msprotz:
Adding the header is a good idea, but will still leave unresolved symbols. Can we add the .c file just to give a successful build, and then you replace it with your own implementation on the Mozilla side?

msprotz:
Great! Thanks so much. It'll unblock the build for all OSX users.

msprotz:
Hi @Frosne, since your initial patch broke the build, can you submit a PR that implements the fix I suggest in https://github.com/project-everest/hacl-star/issues/540#issuecomment-1075440160? Part 2 is necessary, and part 1 of my suggested fix is good to have, but not mandatory. Thanks!

Frosne:
Hi, but I already did it! It is here: 

https://github.com/project-everest/hacl-star/pull/532/commits/e95e9380f3c8dfa91d8be0b97e8ae22093463d55

msprotz:
Ah great. For next time, can you add a cross-reference in the commit message? Just write "this fixes #540" in the commit message and a little link will appear here pointing to the commit, saying that this issue is fixed in your commit on the PR.

[...]

PR #649, opened 2022-10-18
Simplify the way dist/mozilla is assembled

msprotz:
Right now, constructing dist/mozilla (i.e., producing the C code that lives there) requires a high degree of expertise.
- Need to understand bundling, reachability, how krml works, etc.
- This requires maintaining a different set of bundling flags because some algorithms are not enabled when producing dist/mozilla

There were a lot of historical reasons to have dist/mozilla:
- problems with file C including the headers for files A and B even though it didn't need them (now fixed)
- problems with inclusion of libintvector, libintrinsics everywhere (also fixed)
- problems with inclusion of evercrypt_targetconfig.h everywhere (fixed by this PR)

This PR, instead of invoking krml to produce dist/mozilla, now "cherry-picks" from dist/gcc-compatible the algorithms that mozilla is interested in, thus producing the minimum subset of dist/gcc-compatible that contains the algorithms that mozilla wants to package.

This greatly simplifies maintenance going forward: adding SHA3 streaming merely requires adding one line in dist/package-mozilla.sh

Now about the drawbacks:
- some header details: mozilla gets away with just including stdbool.h because no code contains calls to KRML_CHECK_SIZE, but the code now includes krml/internal/target.h -- not a big deal, because this was going to happen sooner or later
- ~in dist/gcc-compatible, EverCrypt_CTR calls chacha20_update, meaning it gets exposed in internal/Hacl_Chacha20.h, and gets a full name Hacl_Impl_Chacha20_chacha20_update... even though mozilla does *not* include EverCrypt_CTR, the code of Hacl_Chacha20.c is still affected~ (solved by https://github.com/hacl-star/hacl-star/pull/655/files#diff-6e948535a1d66f51f595870bdb5c0b21b2f5db521a6f1e47eaaf974f9cc9b761L149)
- Hacl_Curve25519_64.c now needs to include ~evercrypt_targetconfig.h~ `config.h` ~(like it does in dist/gcc-compatible), which in dist/mozilla is a single include `#include "config.h"`~. Curve64 needs the macro "HACL_CAN_COMPILE_INLINE_ASM", which is defined in EverCrypt_TargetConfig.fsti. The fact that dist/mozilla/Hacl_Curve25519_64.c did *not* include this file leads me to think that NSS was always missing out on the fast inline assembly version of curve64, unless the flag was set through the build system!

@Frosne can you look at this and tell me what you think of this new process? I think this would greatly simplify your life, and mine too :-)


msprotz:
@Frosne if you wanted to add streaming SHA3, all you'd have to do is add Hacl_SHA3 and Hacl_Streaming_SHA3 to this list: https://github.com/hacl-star/hacl-star/pull/649/files#diff-c8ac3e6ff09d0f581a297a78ff08d2faf0bd19456c14276e43fe8bd636667a6cR25

msprotz:
```
/nix/store/p7bpdnxqd3i5hwm92mrscf7mvxk66404-bash-5.1-p16/bin/bash: ./package-mozilla.sh: /bin/bash: bad interpreter: No such file or directory
```

@pnmadelaine ? will it get fixed if I used `/bin/env bash` instead?

Frosne:
> ```
> /nix/store/p7bpdnxqd3i5hwm92mrscf7mvxk66404-bash-5.1-p16/bin/bash: ./package-mozilla.sh: /bin/bash: bad interpreter: No such file or directory
> ```
> 
> @pnmadelaine ? will it get fixed if I used `/bin/env bash` instead?

Worked for me with  ``/usr/bin/env ``

msprotz:
Thanks @Frosne I will try that. In the meanwhile, can you let me know if this would be an acceptable workflow for Mozilla going forward? Thanks!

Frosne:
We have two main concerns.
The first one is that now all the modifications influence everyone. It means that if there is any bundle change, then every HACL* user will need to change their build system.
Second, I think we would still prefer to leave the dist/mozilla folder and to build it (using the script) every time, such that we are sure that there will be no modification that will ruin it.

msprotz:
- Regarding your second concern: I'm happy to leave dist/mozilla under version control (as this PR does) so that we can track changes and stop PRs that might have undesirable changes for dist/mozilla
- Regarding your first concern, then: if we have dist/mozilla under version control, we can prevent a bundle change in the Makefile from affecting dist/mozilla by keeping an eye on the changes.

There is a larger discussion to be had about ABI stability, inappropriate renamings, etc. -- which would be good to have for all downstream consumers in general, and Mozilla in particular.

Please let me know if keeping dist/mozilla under version control works for you, or if you have specific changes you would like to see on this PR. Thanks!

Frosne:
>>There is a larger discussion to be had about ABI stability, inappropriate renamings, etc. -- which would be good to have for all downstream consumers in general, and Mozilla in particular.

We can have a call all together to (help) discuss if we could have any ideas/concerns/experience sharing. 

>> Please let me know if keeping dist/mozilla under version control works for you, or if you have specific changes you would like to see on this PR. Thanks!

It looks pretty good. Thanks for your work.

msprotz:
@Frosne do you want streaming sha3 as part of this PR, while we're at it?

Frosne:
> @Frosne do you want streaming sha3 as part of this PR, while we're at it?

yes, it would be lovely.

msprotz:
> yes, it would be lovely.
We only have sha3-256, is that ok?

msprotz:
> Then should I close the RSAPSS PR?

It'll close automatically once this PR is merged.

Frosne:
> > yes, it would be lovely.
> 
> We only have sha3-256, is that ok?

It's an extremely good start, so yes, it's ok.

Frosne (approved):

msprotz:
@pnmadelaine hydra seems to be stuck here, did I do something wrong?

msprotz:
Ok this PR is ready for merging.

Three important things, @Frosne:
- streaming APIs now return an error code to indicate whether the maximum length has been exceeded when calling `update`-- please make sure you check it on the NSS side!
- I have added SHA3 per your request
- dist/mozilla contains two files that Mozilla does not care about, but that are needed to successfully keep dist/mozilla under CI on our side:
  - `configure`: I assume you either write config.h by hand, or pass `-D` flags to your compiler, and
  - Lib_Memzero0.c: I believe you have your own implementations for zeroing out memory.

This latter point about Lib_Memzero0 means that this PR fixes #540 

Then, I'd like to revive #532 and remove those VLAs.

karthikbhargavan (approved):
Looks good to me. We can do another pass if something doesn't work on mozilla's side, but this seems to be an immediate improvement to HACL*.

Frosne:
> Ok this PR is ready for merging.
> 
> Three important things, @Frosne:
> 
>     * streaming APIs now return an error code to indicate whether the maximum length has been exceeded when calling `update`-- please make sure you check it on the NSS side!
> 
>     * I have added SHA3 per your request
> 
>     * dist/mozilla contains two files that Mozilla does not care about, but that are needed to successfully keep dist/mozilla under CI on our side:
>       
>       * `configure`: I assume you either write config.h by hand, or pass `-D` flags to your compiler, and
>       * Lib_Memzero0.c: I believe you have your own implementations for zeroing out memory.
> 
> 
> This latter point about Lib_Memzero0 means that this PR fixes #540
> 
> Then, I'd like to revive #532 and remove those VLAs.

Hello, and thanks for the PR!
1) Ok, I will keep that in mind. Just a small question: do you/we/anyone want to have the maximum length in the comments? https://github.com/hacl-star/hacl-star/blob/main/dist/gcc-compatible/Hacl_Streaming_SHA1.h For instance, to specify here that the max length is ... bytes here. 
2) Thanks!
3) Ack. 

Finally, about the SHA3 PR (https://github.com/hacl-star/hacl-star/pull/532). Feel free to open it if needed. If any work is needed from my side, I am happy to help.

msprotz:
@duesee has been doing a round of improvements for comments in HACL*, so maybe specifying the maximum lengths would be good here?

duesee:
I think it is helpful to include key sizes, output lengths, max lengths, and other constraints in *some* form in the C docs. The more difficult question is how we make sure to not introduce mistakes. For now, I just put the (hopefully) correct numbers everywhere. But this is maybe something we want to revisit later. We could make sure that there are identifiers (and aliases) for everything.

PR #652, opened 2022-10-19
Use new option from fstarlang/karamel#299 and help with #644

msprotz:
See #644. Basically, the addition of functions that perform allocations in static headers mean that these function bodies may now be parsed by a C++ compiler, which mandates that allocations be cast to their destination type (C89 also forbids this, but C99 adds implicit casts.)

We now cast every allocation. Simple. I'll push a commit that performs a refresh for dist/gcc-compatible just so that we can see the impact.


msprotz:
@franziskuskiefer can you confirm this fixes your build issue? thanks!

polubelova (approved):
Thanks!

R1kM (approved):
Looks good to me, thanks!

franziskuskiefer:
> @franziskuskiefer can you confirm this fixes your build issue? thanks!

Yes, this fixes the issue. (CI: https://github.com/cryspen/hacl-packages/pull/232/checks)

PR #654, opened 2022-10-20
Fix issues with bindings/js/test.html

lprichar:
It previously referenced a file not in source control and needed to include the `Hacl_Bignum_Base` module


msprotz:
Super great, thanks Lee!

PR #656, opened 2022-10-21
remove extra EC scalar multiplication when validating a public key for P256

polubelova:
This PR removes the extra EC scalar multiplication when validating a public key for P256.
This expensive operation is not needed for P256 as the cofactor is equal to one.


Frosne (approved):

code/ecdsap256/Hacl.Impl.P256.Signature.Common.fst
```diff
@@ -262,33 +262,30 @@ let isOrderCorrect p tempBuffer =
...
-    let pubKeyY = sub pubKey (size 32) (size 32) in 
-    let tempBuffer = create (size 120) (u64 0) in 
-      let tempBufferV = sub tempBuffer (size 0) (size 100) in 
-      let publicKeyJ = sub tempBuffer (size 100) (size 12) in 
-      let publicKeyB = sub tempBuffer (size 112) (size 8) in  
-	let publicKeyX = sub publicKeyB (size 0) (size 4) in 
-	let publicKeyY = sub publicKeyB (size 4) (size 4) in 
+    let pubKeyX = sub pubKey (size 0) (size 32) in
+    let pubKeyY = sub pubKey (size 32) (size 32) in
+    let publicKeyJ = create (size 12) (u64 0) in
```
R1kM, line 277: Is it cheaper here to have two successive allocations rather than one larger for publicKeyJ and publicKeyB, followed by a sub ?
It might be almost the same given the sizes and that the two arrays are stack-allocated.

R1kM (approved):
Thanks!

code/ecdsap256/Hacl.Impl.P256.Signature.Common.fst
polubelova: Actually, this function doesn't need to convert a point from affine `publicKeyB` to jacobian coordinates `publicKeyJ`, so we could save one allocation. It's usually better to allocate a larger array.

PR #658, opened 2022-10-24
Remove C89 distribution. This fixes #522

### LINKED ISSUE #522: c89-compatible is not c89 compatible
There are a typedef redefinitions in the c89 code. (They are not supported in old compilers and not allowed in c89.)

https://github.com/project-everest/hacl-star/blob/ef1d9bfe7393f8a43c6bbbe279c8b25871039f1e/dist/c89-compatible/EverCrypt_DRBG.c#L78

https://github.com/project-everest/hacl-star/blob/ef1d9bfe7393f8a43c6bbbe279c8b25871039f1e/dist/c89-compatible/EverCrypt_DRBG.h#L67

```
1 warning generated.
[72/77] Building C object CMakeFiles/hacl_static.dir/Debug/src/c89/EverCrypt_DRBG.c.o
src/c89/EverCrypt_DRBG.c:78:1: warning: redefinition of typedef 'EverCrypt_DRBG_state_s' is a C11 feature [-Wtypedef-redefinition]
EverCrypt_DRBG_state_s;
^
include/c89/EverCrypt_DRBG.h:67:41: note: previous definition is here
typedef struct EverCrypt_DRBG_state_s_s EverCrypt_DRBG_state_s;
```

Same goes for states in ctr and aead.

msprotz:
Rationale.
- Serious issues that are not fixed: https://github.com/FStarLang/karamel/issues/283
- Not a single high-profile consumer of the pure c89 mode
- More maintenance burden for us here.

The one mode that matters is the MSVC mode, but for that we have -fmsvc.

Maybe the ElectionGuard folks need ancient compiler compatibility, but I'm not even sure about that. I can check.

In any case I'd feel better removing this now that we have code in the tree that triggers https://github.com/FStarLang/karamel/issues/283


polubelova (approved):

R1kM (approved):
Thanks!
Shouldn't this PR also clean up occurences of c89 across the codebase?
After a quick grep, I see some in doc/Obtaining.rst and Makefile.

PR #659, opened 2022-10-25
OCaml API: finer-grained passing of buffers in NaCl functions (part 2)

victor-dumitrescu:
Expanding on #647, which introduced optional `offset` and `len` arguments to NaCl `Noalloc.Detached` functions, this PR applies the same treatment to `Noalloc.Easy`. Since these functions use separate `pt` and `ct` buffers, both can now optionally have offsets, as well as `pt_len` or `ct_len`. Tests and documentation have been updated accordingly.


polubelova (approved):
Thanks!

bindings/ocaml/Hacl.mli
```diff
@@ -405,50 +405,68 @@ module NaCl : sig
...
-        authentication tag into a single buffer.
+    (** The {i easy} interface concatenates the ciphertext and the 16-byte long
+        message authentication tag into a single buffer.
+
+        By default, these functions use the whole of [pt] and [ct], but users
+        can choose to only pass portions of these buffers, by passing some of
+        these optional arguments:
+        - [pt_offset], [ct_offset]: start at the specified position in [pt] or
+          [ct] (0 by default)
+        - [pt_len] or [ct_len]: specify the number of bytes to take in [pt] or
```
msprotz, line 453: is it implicit that if pt_offset is specified, then pt_len is taken to be "all the way to the end of the buffer"?

bindings/ocaml/Hacl.mli
```diff
@@ -405,50 +405,68 @@ module NaCl : sig
...
+          [ct] (by default, the entire buffer)
+
+        {i Note 1:} Since it must always be the case that [ct] be 16 bytes
+        longer than [pt], functions accept only one of these arguments
+        ([pt_len] for encryption functions, [ct_len] for decryption functions)
```
msprotz, line 458: and error out otherwise?

msprotz (approved):
Looks good to me, modulo the little comments here and there to clarify the documentation.

bindings/ocaml/Hacl.mli
victor-dumitrescu: No, it's that these functions only define one or the other as an argument. For encryption functions I've added `pt_offset`, `pt_len`, `ct_offset` and for decryption functions it's `ct_offset`, `ct_len`, `pt_offset` (so the len argument always refers to the buffer being read. Should I rephrase this note to make that clearer?

bindings/ocaml/Hacl.mli
victor-dumitrescu: It's always the case (whether `pt_offset` is specified or not) that not specifying `pt_len` means "all the way to the end".

franziskuskiefer (approved):
Thanks. lgtm

PR #660, opened 2022-10-26
Import Nix derivation from hacl-star/hacl-nix

pnmadelaine:
Moving the derivation for hacl from https://github.com/hacl-star/hacl-nix.
Then hacl-star/hacl-nix will only hold derivations for dependencies (z3, fstar, karamel, vale).
There's no good reason to keep the derivation separate, and doing so creates issues when a change requires an update to the Nix code.


msprotz (approved):
Looks good to me. Does this have to be at the repository root?

pnmadelaine:
`flake.nix` has to be, `default.nix` can be moved

R1kM (approved):

PR #663, opened 2022-10-27
Add an simplified, proved equivalent version of the Blake2 spec

R1kM:
This PR adds a simplified version of the `blake2_update` function used in the Blake2 spec.
This simplified version and part of the equivalence proof were written by @protz.


lib/Lib.Sequence.Lemmas.fsti
```diff
@@ -98,6 +98,18 @@ val repeati_right_extensionality:
     Loops.repeat_right 0 n (Loops.fixed_a a) f acc0 ==
     Loops.repeat_right lo_g (lo_g + n) (Loops.fixed_a a) g acc0)
 
+/// A specialized version of the lemma above, for only shifting one computation,
+/// but specified using repeati instead
```
msprotz, line 102: I know it's not your code, but why is the lemma above stated in terms of repeat_right when its name implies it's a lemma related to repeati?

specs/Spec.Blake2.fst
```diff
@@ -491,6 +491,17 @@ let blake2_update a kk k d s =
      else blake2_update_blocks a (size_block a) d s
   else blake2_update_blocks a 0 d s
 
+val blake2_key_block:
```
msprotz, line 494: can this definition be reused in blake2_update_key above?

msprotz (approved):
Looks good to me. My comments are more like questions... feel free to merge once you get a green.

lib/Lib.Sequence.Lemmas.fsti
R1kM: I wondered the same thing. My best guess is that the name comes from the left part of the equality, which is the same as repeati n f acc0 (which can be derived using LoopCombinators.repeati_def). We could maybe replace the left part directly by a repeati

specs/Spec.Blake2.fst
R1kM: Seems like it. Should I make the change in this PR?

msprotz: if you don't mind, that's slightly cleaner, and justifies why this function is in the main spec file (right now it's defined but unused in the main spec)

polubelova:
do you plan to update [Low* code](https://github.com/hacl-star/hacl-star/blob/7d31b8c2742127633afb253a676733dc7f26bce6/code/blake2/Hacl.Impl.Blake2.Generic.fst#L757) with `blake2_update'`?

R1kM:
We're not completely sure for now. The simplified spec should be useful to integrate Blake2 into the streaming functor, but this can likely be done without changing the code, by writing a transparent wrapper around update and calling the spec equivalence lemma.
If needed, we'll do it in a future PR, assuming this does not introduce performance regressions.

polubelova (approved):
Thanks!

PR #665, opened 2022-10-28
stabilize proofs for lib.exponentiation

polubelova:
This PR fixes CI for #649.

We can also slit `Lib.Exponentiation` into `Lib.Exponentiation` and `Lib.MultiExponentiation` but it can be done later.


R1kM (approved):
Thanks!

msprotz (approved):
Looks good, thanks! Nice removing those opens. I wish we had a "unused open" warning in F*...

PR #669, opened 2022-11-03
docs: Add documentation for EverCrypt AEAD, Curve25519, and HKDF.

duesee:
Hope this is not too tiring :-) I have a feeling that we might want to discuss here a bit. That's why this PR is a bit smaller.

### DIFF
--- a/providers/evercrypt/EverCrypt.AEAD.fsti
+++ b/providers/evercrypt/EverCrypt.AEAD.fsti
@@ -40,6 +40,9 @@ open EverCrypt.Error
 /// first.
 
 [@CAbstractStruct]
+
+[@@ Comment "Both encryption and decryption require a state that holds the key.
+The state may be reused as many times as desired."]
 val state_s: alg -> Type0
 
 inline_for_extraction noextract
@@ -127,6 +130,11 @@ val frame_invariant: #a:alg -> l:B.loc -> s:state a -> h0:HS.mem -> h1:HS.mem ->
 noextract
 let bytes = Seq.seq uint8
 
+[@@ Comment "Return the algorithm used in the AEAD state.
+
+@param s State of the AEAD algorithm.
+
+@return Algorithm used in the AEAD state."]
 val alg_of_state (a: G.erased alg) (s: state (G.reveal a)): Stack alg
   (requires (fun h0 -> invariant #(G.reveal a) h0 s))
   (ensures (fun h0 a' h1 ->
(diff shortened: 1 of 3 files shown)


providers/evercrypt/EverCrypt.AEAD.fsti
```diff
@@ -40,6 +40,9 @@ open EverCrypt.Error
 /// first.
 
 [@CAbstractStruct]
```
polubelova, line 42: I think the syntax for defining a set of attributes is as follows:
```
[@@ CAbstractStruct; Comment "Both encryption and decryption require a state that holds the key.
The state may be reused as many times as desired."]
```
@protz can confirm that

providers/evercrypt/EverCrypt.AEAD.fsti
```diff
@@ -199,6 +207,14 @@ let alloca_st (a: alg) =
 /// required for encrypt.
 (** @type: true
 *)
+[@@ Comment "Create the context element for the AEAD algorithm.
```
polubelova, line 210: Should we use `state` instead of `context`?
```suggestion
[@@ Comment "Create the AEAD state.
```

providers/evercrypt/EverCrypt.AEAD.fsti
```diff
@@ -199,6 +207,14 @@ let alloca_st (a: alg) =
...
+
+@param a The argument `a` must be either of:
+  * `Spec_Agile_AEAD_AES128_GCM`,
+  * `Spec_Agile_AEAD_AES256_GCM`, or
+  * `Spec_Agile_AEAD_CHACHA20_POLY1305`.
+@param dst A caller allocated reference where the context will be written to.
+@param k Pointer to memory where the key is read from. The size depends on the used algorithm."]
```
polubelova, line 217: Should we specify the size of a key for each algorithm?

providers/evercrypt/EverCrypt.AEAD.fsti
```diff
@@ -319,6 +335,19 @@ let encrypt_st (a: supported_alg) =
...
+
+@param s Pointer to the The AEAD state created by `EverCrypt_AEAD_create_in`. It already contains the encryption key.
+@param iv Pointer to `iv_len` bytes of memory where the nonce is read from.
+@param iv_len Length of the nonce. Note: ChaCha20Poly1305 requires a 12 byte nonce.
+@param ad Pointer to `ad_len` bytes of memory where the associated data is read from.
+@param ad_len Length of the associated data.
+@param plain Pointer to `plain_len` bytes of memory where the to-be-encrypted plaintext is read from.
+@param plain_len Length of the to-be-encrypted plaintext.
+@param cipher Pointer to `plain_len` bytes of memory where the ciphertext is written to.
+@param tag Pointer to the tag. The length of the `tag` must be of a suitable length for the chosen algorithm. There is no length parameter for `tag`.
```
polubelova, line 348: Should we add the length of `tag` for each supported algorithm?

providers/evercrypt/EverCrypt.AEAD.fsti
polubelova: Add a comment on calling `EverCrypt_AEAD_free`?

polubelova (approved):
Thanks!

duesee:
Thanks for the review!

There might be a better (and more aesthetic) way to document the tag and key lengths. But it is certainly helpful to already have something in the docs. So this is a bit of an best effort for now.

R1kM (approved):
Looks good to me, the best effort approach makes sense, better documentation is already an improvement.
Thanks!

PR #681, opened 2022-11-09
JS: published hacl-wasm 1.2.0

victor-dumitrescu:
Small changes made to publish the [hacl-wasm npm package](https://www.npmjs.com/package/hacl-wasm)

### DIFF
--- a/bindings/js/Makefile
+++ b/bindings/js/Makefile
@@ -14,5 +14,7 @@ endif
 all:
 	cp $(KRML_HOME)/krmllib/js/loader.js .
 	cp ../../dist/wasm/*.{wasm,js,html} .
+	cp ../../dist/wasm/layouts.json .
+
 	node api_test.js
 	node test2.js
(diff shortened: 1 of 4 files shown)


bindings/js/Makefile
```diff
@@ -14,5 +14,7 @@ endif
 all:
 	cp $(KRML_HOME)/krmllib/js/loader.js .
 	cp ../../dist/wasm/*.{wasm,js,html} .
+	cp ../../dist/wasm/layouts.json .
```
R1kM, line 17: Why is this now needed?

R1kM (approved):
Thanks!

bindings/js/Makefile
victor-dumitrescu: It was introduced in `dist` after the previous release

msprotz (approved):
Looks good to me, thanks Victor for taking care of this!

PR #693, opened 2022-11-15
Split Blake2 into Blake2b_32 and Blake2s_32

pnmadelaine:
As a first step towards https://github.com/hacl-star/hacl-star/issues/675 I split the `Hacl_Hash_Blake2` bundle into `Hacl_Hash_Blake2b_32` and `Hacl_Hash_Blake2s_32`.
At the moment `Hacl_Blake2b_32_blake2b` and `Hacl_Blake2b_32_blake2s` are both located in `Hacl_Hash_Blake2.h` while `Hacl_Blake2b_256_blake2b` and `Hacl_Blake2s_128_blake2s` are located in `Hacl_Hash_Blake2b_256.h` and `Hacl_Hash_BLake2s_128.h`.
This PR makes it so that `Hacl_Blake2[x]_[y]_blake2[x]` is located in `Hacl_Hash_Blake2[x]_[y].h`.

## to do
- [ ] rename `Hacl_Blake2[x]_[y]_blake2[x]` to `Hacl_Hash_Blake2[x]_[y]_hash` according to the new scheme


Makefile.common
```diff
@@ -108,7 +108,8 @@ CURVE_BUNDLE=-bundle Hacl.Curve25519_64=Hacl.Impl.Curve25519.Field64.Vale \
 # First, match the Blake2 stuff
 BLAKE2_BUNDLE_BASE= \
   -bundle Hacl.Impl.Blake2.Constants -static-header Hacl.Impl.Blake2.Constants \
-  -bundle 'Hacl.Blake2b_32+Hacl.Blake2s_32=Hacl.Hash.Blake2,Hacl.Impl.Blake2.\*,Hacl.Hash.Core.Blake2[rename=Hacl_Hash_Blake2]'
+  -bundle 'Hacl.Blake2b_32=Hacl.Hash.Blake2,Hacl.Impl.Blake2.\*,Hacl.Hash.Core.Blake2[rename=Hacl_Hash_Blake2b_32]' \
+  -bundle 'Hacl.Blake2s_32=Hacl.Hash.Blake2,Hacl.Impl.Blake2.\*,Hacl.Hash.Core.Blake2[rename=Hacl_Hash_Blake2s_32]'
```
msprotz, line 112: krml will NEVER duplicate code, so it does not make sense to repeat Hacl.Impl.Blake2 in two consecutive bundles

dist/gcc-compatible/internal/Hacl_Hash_Blake2b_32.h
```diff
@@ -36,7 +36,8 @@ extern "C" {
 
 
 #include "internal/Hacl_Krmllib.h"
-#include "../Hacl_Hash_Blake2.h"
+#include "internal/Hacl_Hash_Blake2s_32.h"
```
msprotz, line 39: bundling mistake here: blake2b includes a header for blake2s... need to understand why

msprotz (commented):
I would leave the diff for election-guard and others out of the PR, and keep only the diff for gcc-compatible. Much easier to review and discuss.

Have you read https://fstarlang.github.io/lowstar/html/AdvancedTopics.html#bundling ?

Looks like there are several issues with your bundles. Bearing in mind that krml will never duplicate code, it looks like blake2b needs a definition that ends up being bundled underneath blake2s. Two solutions:
- hoist the common parts into an "as-needed" bundle, e.g. `-bundle Hacl.Impl.Blake2.*[rename=Blake2_Common]`
- figure out which definitions are causing this dependency arrow, and mark them as inline_for_extraction noextract (if applicable).

Happy to chat more (e.g. on Slack) to help you out if that doesn't make any sense. Thanks!

msprotz:
Another detail that I neglected to mention: the reason Blake2s/128 and Blake2b/256 are in separate files is because those files need different compilation options (such as -mavx and -mavx2 respectively on x64 toolchains). Blake2b/s were not split because there was no such need.

msprotz:
Oh and also, I don't see the *deletion* of the old files in the diff. Is that intended?

msprotz:
OOM?

msprotz (approved):
FYI this is good for me

bindings/ocaml/Hacl.ml
```diff
@@ -18,9 +18,8 @@ module Hacl_HMAC = Hacl_HMAC_bindings.Bindings(Hacl_HMAC_stubs)
 module Hacl_Poly1305_32 = Hacl_Poly1305_32_bindings.Bindings(Hacl_Poly1305_32_stubs)
 module Hacl_HKDF = Hacl_HKDF_bindings.Bindings(Hacl_HKDF_stubs)
 module Hacl_NaCl = Hacl_NaCl_bindings.Bindings(Hacl_NaCl_stubs)
-module Hacl_Hash_Blake2 = Hacl_Hash_Blake2_bindings.Bindings(Hacl_Hash_Blake2_stubs)
-module Hacl_Blake2b_32 = Hacl_Hash_Blake2
-module Hacl_Blake2s_32 = Hacl_Hash_Blake2
+module Hacl_Blake2b_32 = Hacl_Hash_Blake2b_32_bindings.Bindings(Hacl_Hash_Blake2b_32_stubs)
```
franziskuskiefer, line 21: We should remove the the ocaml bindings here now. They diverged from what's published through hacl-packages and it will get painful the more we diverge.

@victor-dumitrescu @hacl-star/maintainers are we good to drop the ocaml bindings here or do we need anything else to happen before?

dist/gcc-compatible/internal/Hacl_Hash_Base.h
```diff
@@ -22,8 +22,8 @@
  */
 
 
-#ifndef __internal_Hacl_Hash_Blake2_H
-#define __internal_Hacl_Hash_Blake2_H
+#ifndef __internal_Hacl_Hash_Base_H
```
franziskuskiefer, line 25: Is this really generic for all hashing? Looks like it's only used in blake2.
And if it's used across multiple algorithms, I wonder if it's not breaking the ability to use only a subset of algorithms.

bindings/ocaml/Hacl.ml
victor-dumitrescu: Not sure how to proceed for the time being. I'm currently working on adding the OCaml bindings for https://github.com/hacl-star/hacl-star/blob/main/dist/gcc-compatible/Hacl_K256_ECDSA.h, which I don't think is in hacl-packages yet.

Other than for the opam files, have they diverged in the code itself?

franziskuskiefer: It's mostly the opam files I think. But it requires manually merging code, which is annoying and time consuming.

The plan is to finish up the CI integration, i.e. finish the remaining issues from #695 (comments with CI runs), and then do the move. This ensures that we get feedback on the ocaml bindings here in the hacl-star repo.

dist/gcc-compatible/internal/Hacl_Hash_Base.h
pnmadelaine: @msprotz what should we do?
The line that defines this bundle is  `-bundle Hacl.Hash.Definitions=Hacl.Hash.*[rename=Hacl_Hash_Base]`, how is it that only blake2 uses it?
It also seems like this PR is adding a lot of code in `Hacl_Hash_Base.c`.

msprotz:
@pnmadelaine the refactoring has now landed, maybe time update this branch and see where we stand? thanks!

msprotz (commented):
Nice to see that it builds on Hydra. Two thoughts:
- why is Hacl_Hash_Blake2.{c,h} not affect in the current diff on this PR? I would've expected it to be reduced to very little, perhaps even the empty file.
- Blake2 acts both as a hash and as a keyed hash -- do we want the files to start with  Hacl_Hash_Blake2... or Hacl_Blake2...?

pnmadelaine:
replaced by #789

PR #694, opened 2022-11-15
bignum: avoid VLAs in loops

### LINKED ISSUE #653: VLAs in loops compile to stack-growing `alloca`s for MSVC
In C11, the storage duration of VLAs is specified so that they are freed when they go out of scope.

> One exception is the [VLAs](https://en.cppreference.com/w/c/language/array#Variable-length_arrays); their storage is allocated when the declaration is executed, not on block entry, and deallocated when the declaration goes out of scope, not when the block is exited (since C99).

See https://en.cppreference.com/w/c/language/array, https://en.cppreference.com/w/c/language/storage_duration, etc.

Concretely, the array `x` gets freed after each loop body.

```
for (...) {
  t x[foo];
}
```

However, when compiling for MSVC, VLAs are not allowed and such allocations get compiled to `alloca`, which has different semantics. See https://man7.org/linux/man-pages/man3/alloca.3.html -- `alloca` grows the *stack frame*.

This means that in MSVC mode, the for-loop is compiled as follows, which means that it grows the stack monotonically for each loop iteration.

```
for (...) {
  t *x = alloca(sizeof t * foo);
}
```

MSVC has explicitly stated they do not want to support VLA (security risk). Two options:
- krml changes this to a pair of malloc/free, but then this means that code in the `Stack` effect is no longer guaranteed to only Stack-allocate
- fatal warning in krml, and you have to rewrite the code to either use malloc/free, or to hoist the stack allocation outside of the for-loop

Unfortunately, this happens in bignum code:
- https://github.com/hacl-star/hacl-star/blob/cd82cb8391f48e819fac229d32c6cacbefd11cc5/dist/msvc-compatible/Hacl_Bignum.c#L1836
- https://github.com/hacl-star/hacl-star/blob/cd82cb8391f48e819fac229d32c6cacbefd11cc5/dist/msvc-compatible/Hacl_Bignum.c#L2127
- https://github.com/hacl-star/hacl-star/blob/cd82cb8391f48e819fac229d32c6cacbefd11cc5/dist/msvc-compatible/Hacl_Bignum.c#L2324

There are other occurrences in GenericField64, but these might be stemming from the same root cause.

polubelova:
This PR fixes #653.

### DIFF
--- a/code/bignum/Hacl.Impl.Exponentiation.fst
+++ b/code/bignum/Hacl.Impl.Exponentiation.fst
@@ -268,17 +268,20 @@ let lmul_acc_pow_a_bits_l_st
   -> b:lbuffer (uint_t a_t SEC) bLen
   -> table:clbuffer (uint_t a_t SEC) (table_len *! len)
   -> i:size_t{v i < v bBits / v l}
-  -> acc:lbuffer (uint_t a_t SEC) len ->
+  -> acc:lbuffer (uint_t a_t SEC) len
+  -> tmp:lbuffer (uint_t a_t SEC) len ->
   Stack unit
   (requires fun h ->
     live h a /\ live h b /\ live h acc /\ live h ctx /\ live h table /\
+    live h tmp /\ disjoint tmp a /\ disjoint tmp b /\
+    disjoint tmp acc /\ disjoint tmp ctx /\ disjoint tmp table /\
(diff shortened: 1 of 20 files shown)


msprotz (approved):
Looks great, thanks!

R1kM (approved):

PR #695, opened 2022-11-16
add workflow to update hacl-packages

pnmadelaine:
When a PR is opened (or updated), this workflow will create (or update) a branch on cryspen/hacl-packages with the corresponding code update.
~~The workflow will then post a link to the branch on the hacl-star PR.~~
If the PR is from branch `[ref]`, then the branch on hacl-packages will be called `hacl-star-[ref]`.
A second workflow deletes the branch when the PR is closed.

### DIFF
--- a/.github/workflows/hacl-packages-create-branch.yml
+++ b/.github/workflows/hacl-packages-create-branch.yml
@@ -1,11 +1,11 @@
-name: Update hacl-packages
+name: Create branch on hacl-packages
 
 on:
   pull_request:
     branches: [ main ]
 
 jobs:
-  update-hacl-packages:
+  hacl-packages-create-branch:
     runs-on: ubuntu-latest
     steps:
       - name: checkout hacl-star
--- a/.github/workflows/update-hacl-packages.yml
+++ b/.github/workflows/update-hacl-packages.yml
@@ -1,0 +1,43 @@
+name: Update hacl-packages
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  update-hacl-packages:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout hacl-star
+        uses: actions/checkout@v3
+        with:
+          path: hacl-star
+      - name: checkout hacl-packages
+        uses: actions/checkout@v3
+        with:
+          repository: cryspen/hacl-packages
+          path: hacl-packages
+          token: ${{ secrets.HACL_BOT }}
+      - name: update hacl-packages
+        working-directory: hacl-packages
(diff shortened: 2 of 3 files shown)


.github/workflows/update-hacl-packages.yml
```diff
@@ -0,0 +1,43 @@
...
+          git config --local user.name "Hacl Bot"
+          git config --local user.email "hacl-star@mailo.com"
+          git checkout -b $branch
+          git add src include
+          git commit -m "[CI] update code"
+          git push --force --set-upstream  origin $branch
+      - name: comment pr
+        uses: thollander/actions-comment-pull-request@v1
+        with:
+          message: "[CI] corresponding branch on hacl-packages: https://github.com/cryspen/hacl-packages/tree/hacl-star-${{ github.head_ref }}"
```
franziskuskiefer, line 42: Let's link to the action triggered on this branch. That's what people should look at.
I'd be fine doing this in a followup. Nothing is triggered yet anyway.

franziskuskiefer (changes requested):
Thanks for getting this started.

Let's add a job that's triggered when a PR is closed or merged to delete the branch. Shouldn't be too difficult and I'd like to avoid all the dangling branches in hacl-packages.

Otherwise I'm fine with this. The exact link to the GH action can be added later.

.github/workflows/update-hacl-packages.yml
pnmadelaine: I removed it for now

msprotz:
Looks good!

R1kM (approved):
Looks good to me, thanks.
Do you already have the tokens needed on the hacl-packages side?

.github/workflows/hacl-packages-create-branch.yml
```diff
@@ -1,11 +1,11 @@
-name: Update hacl-packages
+name: Create branch on hacl-packages
 
 on:
   pull_request:
     branches: [ main ]
```
franziskuskiefer, line 5: Should this get a `reopened` as well to create the branch again?

franziskuskiefer (commented):
Thanks for adding the delete flow.

.github/workflows/hacl-packages-create-branch.yml
pnmadelaine: I just tested on a fork and, indeed, reopening a PR does not trigger the worflow.

franziskuskiefer (approved):
thanks, let's get it in and do the hacl-packages workflow and link it here in the comment.

PR #696, opened 2022-11-17
Remove gcc64 and election-guard distributions

msprotz:
I talked to the ElectionGuard folks and they're onboard with consuming code directly from hacl-packages. One less set of bundling flags to maintain.


polubelova (approved):

R1kM (approved):
Thanks!

PR #698, opened 2022-11-17
Fix bundling issue in WASM

msprotz:
Now that krml is capable of gracefully handling many non-WASM idioms (such as vector types), we don't need to "disable" as many files with bundling.

This simply removes wasm-specific bundling flags, thus fixing #692. This comes at the expense of a few possibly-not-really-helpful wasm files in dist/wasm... but it's not a big deal because clients can choose to only load a subset of the files. (See examples in bindings/js).


R1kM (commented):
As far as I can see, the changes to wasm bundling focus on fixing the bundles for the hashes, but still filter out many other bundles, right? There still seem to be a large number of wasm-specific bundles in the Makefile. Are there plans to clean them up as well?

Also, could you refresh dist/wasm with the changes in your PR so that we can look at the diff?

Thanks!

msprotz:
> There still seem to be a large number of wasm-specific bundles in the Makefile. Are there plans to clean them up as well?

yes, very soon

R1kM (approved):

polubelova (approved):

PR #699, opened 2022-11-17
Blake2 simplifications, 1/n

msprotz:
@R1kM @sonmarcho and I have started simplifying the Blake2 proofs, Agile Hash infrastructure, and so on. The goals are to:
- remove an enormous amount of proofs that we believe we can do without, simplify the rest
- reinstate blake2 streaming with key
- remove the proof that spec.blake2 == spec.mk_update_multi (spec.blake2.update): all that we care about is that blake2 obey the hash laws (associativity, empty-sequence is a null element, incrementality), the detour through the update-multi style spec is totally unnecessary
- same deal with sha3.

This is a first PR in a series of many. This one specifically removes a large burden of proof from Hacl.Streaming.Blake2.fst, a prelude to restoring the keyed mode.


R1kM (approved):
Glad to see all these proofs and the reliance on Spec.Agile.Hash and Spec.Hash.Incremental gone!

msprotz:
I am still seeing an out of memory on Test.Vectors...

I thought the corresponding F* regression was fixed?! Do we have another performance/memory regression?

polubelova (approved):
Thanks! 
Could you please also refresh `dist/gcc-compatible`?

msprotz:
I pushed a refresh of the C code on this branch. It's quite an improvement @polubelova

msprotz:
Does anyone know what happened here? @R1kM I got a green on the second-to-last commit but my most recent commit is empty, and still turned red. F* regression?

R1kM:
Not sure what happened here... I could reproduce locally, but only from the command-line, verification from emacs worked. Bumping the rlimit solved it fully on my machine, let's see what CI says.

sonmarcho (approved):

msprotz:
Huge thanks to both @R1kM and @sonmarcho for helping me finish this PR. It's both a lot of proofs that are now gone, and a lot of improvements on the C code side. Really nice change.

PR #704, opened 2022-11-19
Remove occurences of Warning 240

R1kM:
We had several occurences of Warning 240, indicating a val without a corresponding def, where the corresponding definitions were in other modules.

This PR performs a small clean-up by putting the vals and their corresponding let in the same module, rewiring existing calls to the vals when needed, and marking 240 as fatal.


lib/Lib.IntTypes.Compatibility.fst
```diff
@@ -2,20 +2,6 @@ module Lib.IntTypes.Compatibility
 
 open Lib.IntTypes
 
-val pow2_values: n:nat ->  Lemma (
```
R1kM, line 5: This lemma is subsumed by the version in FStar.Int

msprotz (approved):

msprotz:
You may have forgotten to push your changes to the Makefile

polubelova (approved):

PR #709, opened 2022-11-21
use hacl-packages update process

pnmadelaine:
instead of manually copying files, call `./mach update` inside hacl-packages


franziskuskiefer (approved):
Thanks

R1kM (approved):
Looks good to me, thanks

PR #711, opened 2022-11-22
Unify hash state types between different specs

R1kM:
There currently is a mismatch between the state type of Spec.Hash.Definitions, which uses Seq.seq with a refinement for the length, and the Blake2 and SHA3 specs, which directly use Lib's lseq.
While the two are conceptually equivalent, this leads to many issues with Z3, and the need to call lemmas stating the type equality.

This PR unifies both types, by using `lseq` in Spec.Hash.Definitions, and removing the now unneeded type_equalities lemmas.


msprotz (approved):
This is a really great change, Aymeric... thanks!

sonmarcho (approved):

PR #716, opened 2022-11-29
Move merkle tree to fresh repository

msprotz:
Long-standing issue... please review this, and also the structure of the new repository at https://github.com/hacl-star/merkle-tree

I have an everest branch running to see if we can add it to the everest regressions just like before, except as a separate project.

This would be a good "sample project" if anyone wanted to build anything atop the HACL* repository. Makefile is less than 100 lines.


msprotz:
```
cp: cannot stat 'gcc-compatible/internal/Hacl_Krmllib.h': No such file or directory
```

@pnmadelaine is there a hardcoded list of files somewhere?

polubelova (approved):
Thanks!

Should we also remove obsolete files from `hints/`?

msprotz:
Several issues here...
- I fixed packaging for mozilla. @pnmadelaine looks like we're all good, the error was next to something related to hacl-star-raw so I thought it had to do with the special build of hacl-star-raw, but it was just a coincidence
- I got sucked into a rabbit hole of DRBG changes... I am recapping here for posterity

1. Merkle Tree treats EverCrypt as completely opaque via the `-library` flag of krml -- no C symbols should be generated for EverCrypt.
2. The `-library` flag of krml still keeps type *definitions*, on the basis that they need to be in scope for the prototypes to type-check. Example:

```fstar
module Library // fsti
type t = A | B
val f: t -> Int32.t

module Library // fst
let f = function A -> 0l | B -> 1l
```

```fstar
module Client
let main () = Library.f A
```

```bash
# Invocation
krml Library.fst -o liblibrary.a # separate compilation, compile the library
krml Client.fst Library.fst -bundle Library[rename=Client_Library] -library Library liblibrary.a # assume Library is already compiled and linked
```

Notably, the second invocation will generate `Client_Library.h` with *Client's view of Library*:

```
// Client_Library.h
typedef enum { A, B } t;
extern int32_t f(t x);
```

3. This is all good. EXCEPT that when the type `t` in `Library` is decorated as `[@CAbstractStruct]`, then the client is asserting that the definition of `t` is not needed to type-check (in C) the header of `Library`.
4. krml was not honoring CAbstractStruct when used in conjunction with -library
5. This resulted in a needless type definition for EverCrypt.Hash.state being generated in the code of MerkleTree
6. Because EverCrypt.Hash now can call blake2s/128, in order to type-check, this C definition for EverCrypt_Hash_state needed libintvector.h in scope
7. I fixed krml, so now Merkle Tree does not generate code for EverCrypt.Hash
8. Obviously, some place in HACL was relying on the previous behavior, notably the EverCrypt tests, which behave as a client of EverCrypt and therefore use `-library EverCrypt`
9. Specifically, providers/Test.NoHeap.fst was relying on `EverCrypt.Hash.alloca` which is inline_for_extraction noextract and therefore violates the CAbstractStruct guarantee, because it generates in the client code (Test.fst) an occurrence of EverCrypt_Hash_state which is *not* guarded behind a pointer type
10. Lacking a type definition for EverCrypt_Hash_state, Test.c could not compile
11. I fixed that by making the test become a heap-allocating state
12. Also noticed that HMAC.DRBG did not have a free function (fixed it)
13. Had to rework the test helpers

pnmadelaine (approved):
lgtm

Issue #724, opened 2022-12-02
Improvements in HACL* for the Python PR


msprotz:
For the Python PR (https://github.com/python/cpython/pull/99109) I manually tweaked many small details in order to generate the most compact code. Here is the list of these fixes, and how we should backport them upstream on the Low* code rather than hand-patching the generated C. This will also allow addressing one key review comment, namely the request for a script that can automatically refresh the "vendored" copy of HACL* in Python directly from this repo.

- [x] ~separate sha2_streaming into sha2_256_streaming and sha2_512_streaming (the former also contains 224, the latter also contains 384)~ NOT DOING FOR NOW
- [x] do not include types.h from these files -- they don't need any of those definitions --> fixed with a script, not easy to do actually
- [x] ditch sha2_update_224 -- it's a copy of update_256 and we have a lemma somewhere that these two are identical... if we want to be backwards-compatible, consider using `[@@ CPrologue "#define Hacl_Streaming_SHA2_256_update_224 Hacl_Streaming_SHA2_256_update_256"]` (optional) --> https://github.com/hacl-star/hacl-star/pull/725
- [x] tweak the streaming code so that it does not generate instances of Hacl_Impl_SHA2_Types_uint8_2p --> fixed in https://github.com/FStarLang/karamel/pull/314
- [x] tweak krml to avoid KRML_CHECK_SIZE in streaming --> fixed by https://github.com/FStarLang/karamel/pull/315
- [x] streaming should not need sha2_types --> fixed in https://github.com/FStarLang/karamel/pull/314
- [x] builtin, callconv.h are not needed -- move in krml the `#define inline __inline__` to a more widely-included header, such as target.h https://github.com/FStarLang/karamel/pull/320
- [x] find a way to trim down target.h from a script (I'd rather not break it up upstream)
- [x] expose a state copy operation in streaming functor (#750)
- [x] tweak bundling to fold the constants into the streaming file since from here on this will be our only header to expose SHA2 (#752)
- [x] ditch sha224_update and sha224_update_nblocks from Hacl.SHA2.Scalar32 --> #742, and same thing with 384 --> later

msprotz:
I'm wondering if it's really worth splitting everything into two separate file for the 32-bit and 64-bit versions.

Pros:
- clients don't have to include 4 algorithms if they only want one
- minimizes Python PR

Cons:
[...]

PR #725, opened 2022-12-02
Remove redundant functions from Hacl.Streaming.SHA2; comment the file

msprotz:
## Proposed changes

This removes redundant functions from Hacl.Streaming.SHA2.
- The `update` function is identical for SHA2/224 and SHA2/256 -- I removed the former.
- Same deal for `update` with SHA2/384 and SHA2/512.

I also added a bunch of comments to this module which remained woefully undercommented. This fixes one bullet point from #724.

## Types of changes

[x] Bugfix (non-breaking change which fixes an issue)

## Further comments

We have no verified clients of SHA2-224 or SHA-384, so as far as I know, no one is impacted by the removal of these functions. For C clients, I have added backwards-compat macros.

In the future, should a *verified* client wish to use SHA2-224 with the incremental API, they should be able to rely (famous last words?) on Spec.Lemmas.SHA2.fsti which contains the suitable lemmas, along with a suitable fact exported from Hacl.Streaming.Functor, e.g. given two instantiations of the streaming functor whose types are identical and whose update_multi functions are identical, then the two update functions are interchangeable).

Same deal with SHA-384 and SHA2-512.


code/streaming/Hacl.Streaming.SHA2.fst
```diff
@@ -62,42 +62,56 @@ let hash_t (a: sha2_alg) =
...
+
+[@@ Comment
+"Feed an arbitrary amount of data into the hash. This function returns 0 for
+success, or 1 if the combined length of all of the data passed to `update_256`
+(since the last call to `init_256`) exceeds 2^63-1 bytes.
+
+This function is identical to the update function for SHA2_224.";
+
+CPrologue
+"#define Hacl_Streaming_SHA2_update_224 Hacl_Streaming_SHA2_update_256" ]
```
R1kM, line 95: Relying on an untrusted define for this is a bit annoying. This seems fine for SHA2, as the specs do not change, and we can reasonably assume that update_224 will always have exactly the same signature as update_256, but it might be better to check for it in krml instead, and automatically generate the #define based on the F* code.
Beyond this PR, what would you think for instance of a CDefine/CAbbrvDefine (name to be determined) feature in krml doing exactly that?

R1kM (approved):

msprotz (commented):
@polubelova I've addressed your review comments, thanks for the good catches

there's a little bit of a diff on top of my changes because I'm using the latest krml master which has a few code-quality changes from last Friday

msprotz:
FYI, this PR is not enough: Hacl.Impl.SHA2 currently generates two copies of the update function for sha2-256... we should replace Hacl.SHA2.Scalar32.sha224_update with sha256_update in order to generate better-quality code (the right lemma is in Spec.SHA2.Lemmas).

I'll try to submit a followup PR unless someone beats me to it.

PR #730, opened 2022-12-06
Test hacl-packages CI

pnmadelaine:
This is a test  for https://github.com/hacl-star/hacl-star/pull/719


msprotz:
@pnmadelaine should we close this or do you need it open still for testing?

PR #732, opened 2022-12-06
don't inline scalar32_sha512_init, it's used in ed25519

franziskuskiefer:
## Proposed changes

The `scalar32_sha512_init` function is inlined but used by ed25519. This breaks builds on Windows

## Types of changes

## Checklist

- [x] I have read the [CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md) doc
- [x] I have read and agree to submit my changes under the [LICENSE](LICENSE.md)
- [X] I have edited [CHANGES.md](CHANGES.md) (if appropriate)

## Further comments

See the build issues for example here https://github.com/cryspen/hacl-packages/actions/runs/3631761531/jobs/6126803616

### DIFF
--- a/code/sha2-mb/Hacl.SHA2.Scalar32.fst
+++ b/code/sha2-mb/Hacl.SHA2.Scalar32.fst
@@ -36,7 +36,7 @@ open Hacl.Impl.SHA2.Generic
 [@CInline] let sha384_update_last = update_last #SHA2_384 #M32 sha384_update
 [@CInline] let sha384_finish = finish #SHA2_384 #M32
 
-[@CInline] let sha512_init = init #SHA2_512 #M32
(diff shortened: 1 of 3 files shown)


R1kM:
Thanks! Could you also push a refresh of dist/gcc-compatible so that we can look at the generated C code?

franziskuskiefer:
@pnmadelaine looks like the hacl-packages trigger doesn't handle branch names likes this. Maybe it needs `**`?



> Thanks! Could you also push a refresh of dist/gcc-compatible so that we can look at the generated C code?

Done.

polubelova (approved):
Thanks!

R1kM (approved):
Looks good, thanks!

PR #735, opened 2022-12-08
Blake2 simplifications, 2/n

msprotz:
Followup to #699.

This PR is the result of several weeks' worth of work by Aymeric (and I, to a much lesser extent). Huge thanks to Aymeric!

- Abandon "alternate" specs of sha3/blake2 using higher-order update_multi combinator: all that matters is that these algorithms obey the hash laws. Rewire agile specs to use Spec.Blake2 and Spec.SHA3 respectively.
- Ditch re-implementations of sha3/blake2 in code/hash that were effectively duplicating the code for the sake of having low-level combinators with the expected type signature. Use "thin wrappers" instead that redirect to Impl.SHA3 and Impl.Blake2
- Simplify incremental spec to deal with optional extra state for blake2 -- this had permeated every single part of the implementations and specifications and was, for the most part, completely un-necessary. See, for instance, the amount of code removed in EverCrypt.Hash

There are still a few tidbits left that I'll point out in an upcoming review.


Makefile
```diff
@@ -1033,3 +1033,6 @@ obj/libhaclml.cmxa: $(filter-out $(HACL_HOME)/obj/Meta_Interface.cmx,$(ALL_CMX_F
 %/Makefile:
 	echo "HACL_HOME=$(shell realpath . --relative-to $(dir $@))" > $@
 	echo "include \$$(HACL_HOME)/Makefile.common" >> $@
+
```
msprotz, line 1036: need to remove this (my bad)

code/blake2/Hacl.Blake2b_256.fst
```diff
@@ -51,3 +51,6 @@ let load_state256b_from_state32: Core.load_state_st Spec.Blake2B Core.M256 =
   Core.load_state_from_state32 #Spec.Blake2B #Core.M256
 let store_state256b_to_state32: Core.store_state_st Spec.Blake2B Core.M256 =
   Core.store_state_to_state32 #Spec.Blake2B #Core.M256
+
```
msprotz, line 54: this needs to be marked private, so that it appears in the internal header instead of the public one

code/blake2/Hacl.Blake2b_32.fst
```diff
@@ -45,3 +45,6 @@ let blake2b_finish : Impl.blake2_finish_st Spec.Blake2B Core.M32 =
 @param k Pointer to `kk` bytes of memory where the key is read from."]
 let blake2b : Impl.blake2_st Spec.Blake2B Core.M32 =
   Impl.blake2 #Spec.Blake2B #Core.M32 blake2b_init blake2b_update blake2b_finish
+
```
msprotz, line 48: same for this one, and others -- NOTE: this one is not used anywhere, but perhaps good to have it for consistency

dist/gcc-compatible/EverCrypt_HMAC.c
```diff
@@ -571,51 +566,17 @@ EverCrypt_HMAC_compute_blake2s(
...
-  r1[3U] = iv7;
-  uint64_t es = (uint64_t)0U;
-  K____uint32_t__uint64_t scrut0 = { .fst = s, .snd = es };
-  uint32_t *s0 = scrut0.fst;
+  Hacl_Blake2s_32_blake2s_init(s, (uint32_t)0U, (uint32_t)32U);
+  uint32_t *s0 = s;
   uint8_t *dst1 = ipad;
-  uint64_t ev0 = Hacl_Hash_Core_Blake2_init_blake2s_32(s0);
-  uint64_t ev10;
+  Hacl_Blake2s_32_blake2s_init(s0, (uint32_t)0U, (uint32_t)32U);
```
msprotz, line 572: there are two inits in a row here -- is this a bug in Hacl.HMAC? would be good to figure out why we re-initialize (or how to skip it, e.g. by revealing that alloca provides an already-initialized buffer or something)

specs/Spec.Agile.Hash.fsti
```diff
@@ -6,11 +6,80 @@ include Spec.Hash.Definitions
...
-val update_multi (a:hash_alg) (hash:words_state a) (blocks:bytes_blocks a):
-  Tot (words_state a) (decreases (S.length blocks))
+let init_extra_state (a:hash_alg) : extra_state a = match a with
+  | Blake2B | Blake2S -> 0
+  | _ -> ()
+
+(* The individual update function. This is an implementation detail, and clients
+should reason in terms of update_multi to be fully agile. None of the hash laws
+refer to update. *)
+val update (a:md_alg): update_t a
```
msprotz, line 60: do we still need this? I ditched EverCrypt.Hash.update...

specs/Spec.Hash.PadFinish.fst
```diff
@@ -7,8 +7,19 @@ open Lib.ByteSequence
 open Spec.Hash.Lemmas0
 open Spec.Hash.Definitions
 
-(** This module contains specifications shared across all the Merkle-Damgård
-    constructions. *)
+(** This module contains two things.
```
msprotz, line 10: should we move finish to Spec.Hash.Incremental and pad to Spec.Hash.MD? (this module is weirdly designed)

code/blake2/Hacl.Blake2b_256.fst
msprotz: let's do it as part of a followup PR, when we fix bundling for Blake2 in #693 -- this should be feasible easily with something like `-bundle Hacl.Streaming.Blake2b=Hacl.Blake2b` -- right now making it private obviously prevents EverCrypt.Hash from referring to this

polubelova (approved):

R1kM (approved):

dist/gcc-compatible/EverCrypt_HMAC.c
R1kM: This is solved in #739 . Marking as resolved.

specs/Spec.Hash.PadFinish.fst
R1kM: Yes, this has been bothering me for a while as well. The proposed plan looks reasonable.
Some other related cleanups:
* Spec.Hash.Lemmas should go in specs/lemmas
* Since Spec.Hash.Lemmas now only contains lemmas about update_multi, it does not need to open Spec.Hash.PadFinish, and the extra Spec.Hash.Lemmas0 does not seem needed anymore.

msprotz: 👍

PR #750, opened 2022-12-19
Extend streaming functor with a copy function

R1kM:
## Proposed changes

This PR extends the streaming functor with a copy facility. The implementation is very similar to functions like create_in, except it copies key parts from the initial state instead of initializing them from scratch.
It also adds a SHA2 instantiation of the copy function.

Note, this PR builds on top of #748, which in addition to reestablishing keyed Blake2 was also simplifying and stabilizing many proofs related to the streaming functor.

## Types of changes


msprotz (approved):
Super great, thanks Aymeric!

franziskuskiefer (approved):
Thanks, lgtm.

But I wonder why this is needed or how it should be used. Adding an example of how/when to use this into the comment of `Hacl_Streaming_SHA2_copy_256` would greatly help everyone using the library.

R1kM:
I'm not completely sure when this function is needed, however, this was one of the requirements for the Python integration (see #724). This mostly gives you a facility to do a deep copy of the current hash state.

franziskuskiefer:
@msprotz can you tell us why we need this function?
HACL exposes a lot of functions that aren't really useful and we're working our way through them to figure out what makes sense to expose and what not. So I just want to make sure we make clear in the comments why this function exists.

msprotz:
Added a comment in f2fd29eba8

PR #731, opened 2022-12-06
OCaml API for K-256

victor-dumitrescu:
## Proposed changes

This PR adds an OCaml interface for K-256. It contains:
- the complete OCaml interface in `Hacl.mli`
- the complete documentation, in the style already used for the interface of P-256 and relying on the documentation already present in `Hacl_K256_ECDSA.h`
- test vectors and a QCheck-based test for compatibility with [libsecp256k1](https://github.com/bitcoin-core/secp256k1)

What still needs to be added:
- [x] API and docs for functions added in #734 
- [x] test vectors and testing for all functions

JS bindings for the same functions added in #751 

## Types of changes

What types of changes does your code introduce to HACL*?

- [ ] Proof maintenance (non-breaking change which fixes a proof regression)
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New algorithm or feature (non-breaking change which adds functionality)
- [ ] Improved performance (fix or feature that would improve performance of some algorithm on some platform)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

- [x] I have read the [CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md) doc
- [x] I have read and agree to submit my changes under the [LICENSE](LICENSE.md)
- [x] I have added necessary documentation (if appropriate)
- [x] I have edited [CHANGES.md](CHANGES.md) (if appropriate)

PR #751, opened 2022-12-19
JS bindings for K-256

victor-dumitrescu:
## Proposed changes

This MR adds JS bindings and tests for functions in `Hacl_K256_ECDSA.wasm`. The main change is in `bindings/js/api.json`

## Types of changes

## Checklist

- [x] I have read the [CODE_OF_CONDUCT.md](CODE_OF_CONDUCT.md) doc
- [x] I have read and agree to submit my changes under the [LICENSE](LICENSE.md)
- [x] I have edited [CHANGES.md](CHANGES.md) (if appropriate)


bindings/js/api.json
```diff
@@ -1518,6 +1517,407 @@
...
+        },
+        {
+          "name": "signature",
+          "kind": "input",
+          "type": "buffer",
+          "size": 64,
+          "interface_index": 2,
+          "tests": [
+              "241097efbf8b63bf145c8961dbdf10c310efbb3b2676bbc0f8b08505c9e2f795021006b7838609339e8b415a7f9acb1b661828131aef1ecbc7955dfb01f3ca0e","241097efbf8b63bf145c8961dbdf10c310efbb3b2676bbc0f8b08505c9e2f795021006b7838609339e8b415a7f9acb1b661828131aef1ecbc7955dfb01f3ca0e",
+              "240097efbf8b63bf145c8961dbdf10c310efbb3b2676bbc0f8b08505c9e2f795021006b7838609339e8b415a7f9acb1b661828131aef1ecbc7955dfb01f3ca0e","241097efbf8b63bf145c8961dbdf10c310efbb3b2676bbc0f8b08505c9e2f795021006b7838609339e8b415a7f9acb1b661828131aef1ecbc7955dfb01f3ca0e"
```
victor-dumitrescu, line 1604: Blocking: this test should actually have the result be `false`, it it just the passing first test with the second byte in the signature set to 0. This is correctly rejected by the C function, but here it's returning `true`. I don't know why this is.

bindings/js/api.json
```diff
@@ -1518,6 +1517,407 @@
...
```
victor-dumitrescu, line 1699: Blocking: same as above

victor-dumitrescu:
This shouldn't be merged, the negative test for signature verification is not passing. @msprotz  do you have any idea why this might be the case? Thanks!

msprotz:
I can take a look. Thanks for flagging this

msprotz:
You had *four* inputs for the first two parameters: correct, correct, bad, correct
You had *two* expected results for the returned boolean: correct, bad

I suspect some copy/pasting issue. See my most recent commit.

Rest looks good. Thanks!

victor-dumitrescu:
Ah, not sure how I managed to do that, thanks for spotting it!

msprotz:
No worries. Maybe you could add something in the testing infrastructure that warns against inconsistent lengths of test fields? For instance, at line 56 of api_test.js, you could add something like (sketch):

```javascript
if (func_sig.args.length > 0) {
  let n_tests = func_sig.args[0].tests.length;
  for (let arg of func_sig.args)
    if (arg.tests.length != n_tests)
      ... error ...
  if (func_sig.return.tests.length != n_tests)
    ... error ...
}
```

Sounds like something that's easily avoidable

victor-dumitrescu:
I've added a check for the consistency of the number of test vectors. I think this PR, along with #731, are now good to go.

msprotz (approved):
Looks good to me, thanks for doing this!

R1kM (approved):
Looks good to me, thanks!

PR #752, opened 2022-12-19
Enable internal, static headers and clean up public interface of HACL

msprotz:
## Proposed changes

Heads-up on the effect of branch protz_static_internal of krml

## Types of changes

## Further comments

See https://github.com/FStarLang/karamel/pull/316 for a rationale on the krml side

### Short summary of the changes

**background**: the `-static-header` flag moves definitions from the C file to a header file, and marks such definitions as `static inline`; this means that clients of those definitions each get their own copy in their translation unit, which in turns allows the C compiler to perform a lot more intra-compilation-unit optimizations

typical example is bignum helpers: they're all marked as static inline, so that when Curve25519 includes those, they are inlined and optimized in Curve25519.c; other clients of bignums (e.g. K256, RSA-PSS) also get a copy of those helper bignum functions in K256.c, RSAPSS.c, etc. and the C compiler can aggressively optimize each translation unit

**problem**: previously, `-static-header` automatically made those headers public; this is not great because it pollutes the scope of C clients of HACL with internal definitions that were never intended to be exposed to the public

**the change:** with this krml rewrite (I rewrote a core interprocedural analysis as a fixpoint with a proper lattice of visibility), you can have "static inline" headers that are *internal*, meaning that they appear in `internal/` but not in the client-facing headers

**benefit:** no more scope pollution... and "HACL-internal" helpers like bignum move to `internal/`

Heads-up to @polubelova, possibly @victor-dumitrescu, @franziskuskiefer for downstream hacl-packages... thanks!

~PS: need some help from @pnmadelaine to understand why the mozilla build started failing on the second commit~
UPDATE: realized I need to merge the krml PR first for this to turn green 🤦 

As mentioned on the krml PR, this yields an Everest green, so I'll merge this PR right after the krml PR.


franziskuskiefer:
Thanks, looks like a great cleanup! 👍🏻

R1kM (approved):
Looks great, thanks!

franziskuskiefer (approved):
Thanks, lgtm.
To see the impact of the inline changes (https://github.com/FStarLang/karamel/pull/316#issuecomment-1363424686) the msvc-compatible code would be good to have in here (that's where it would break). But since nothing else changed this should be fine.

msprotz:
I am now unconditionally removing problematic `inline` qualifiers, even for gcc-compatible. The fact that there is no diff indicates that we have no incorrect `inline` qualifier left in the tree.

To confirm it works, I re-instated the CInline qualifier you had removed from sha512_init in #732 -- after regenerating code for gcc-compatible, there was no diff, because the new behavior is to issue a warning (warning 8, which we could conceivably make fatal in HACL*), and drop the inline qualifier.

msprotz:
(Bypassed branch protection because this only will become green after the krml merge.)

PR #756, opened 2022-12-30
Remove old SHA2 implementation

msprotz:
## Proposed changes

This PR is a WIP and aims to remove the old SHA2 implementation in code/hash, in favor of the multibuffer implementation in code/sha2-mb, with n=1 (scalar case).

The goal is to re-implement Hacl.Hash.SHA2 (old implementation) with calls to Hacl.SHA2.Scalar32, performing proofs along the way that Hacl.Hash.SHA2 *does* implement the low-level function signatures from Hacl.Hash.Definitions.

That latter point is important: by having a single unified signature for *all* hash functions (in Hacl.Hash.Definitions), we can implement an agile HMAC, HKDF, HPKE, and so on, that are written against a *generic* hash function, then instantiated for *any* hash function that obeys the generic signature in Hacl.Hash.Definitions. In the new world, this becomes the sole purpose of Hacl.Hash.SHA2: adapting Hacl.SHA2.Scalar32 to the type signatures in Hacl.Hash.Definitions.

Note that this is what we did for Blake2 and SHA3 already: Hacl.Hash.Blake2 and Hacl.Hash.SHA3 are completely marked inline_for_extraction noextract, because they contain no runtime code: they only wrap Hacl.Blake2 and Hacl.SHA3 within the Hacl.Hash.Definition types framework.

## Further comments

I spent a little bit of time on this, and I think we need to discuss before going any further. At a high-level, the thing that makes this change complicated is that the new sha2 uses a different Low* type for its state representation than the old sha2. The new sha2 uses `lbuffer (vec_t U32 1) 8`, i.e. an array of size 8 containing size-1 vectors of uint32s. We do NOT have `vec_t U32 1 == uint32` (as far as I can tell), which means that this is strictly different from `lbuffer uint32 8`, the state used by the old sha2 implementation.

First, the good news:
- with the many rounds of cleanups from the past few months, we have all of the lemmas we need
- by "surgically" modifying Hacl.Hash.Definitions to use the `state_spec_v` function, the proofs go through relatively smoothly
- because of all the work that was done earlier in Hacl.Hash.MD, we could transfer all of the sha2-related reasoning into Hacl.Hash.SHA2, and revert Hacl.Hash.MD to its earlier state where all of the algorithms go through the definitions in Hacl.Hash.Definitions (I don't think it's a drawback: we had to proceed modularly and it was good to tackle the streaming part first, without getting mired in the complexities of all the other clients).

The neutral news:
- need to fixup a few proofs in Hacl.Spec.SHA2.EquivScalar, notably, move the reasoning for update_multi in there to make it completely seamless
- need to expose the same kind of proofs for update_last and finish, they're currently not in the fsti

The bad news:
- some clients were relying on the old state type, notably EverCrypt's mixed SHA2 that calls Vale to leverage SHAEXT.

Now allow me to say more about this latter point...

The old SHA2 implementation was a completely generic Merkle-Damgård hash implementation. This means that EverCrypt.Hash redefines `update_multi` (https://github.com/hacl-star/hacl-star/blob/main/providers/evercrypt/fst/EverCrypt.Hash.fst#L304) then uses it to derive `update_last` (https://github.com/hacl-star/hacl-star/blob/main/providers/evercrypt/fst/EverCrypt.Hash.fst#L379), and so on all the way up to the hash function.

This is a problem, because unless we do something, EverCrypt.Hash will keep the old implementation alive.

I think we can come up with a relatively painless fix, relying on the fact that:
- unlike the Curve25519 case, the SHAEXT and Scalar32 implementations of SHA2 do share the same representation under the hood (minus the fact that `vec_t U32 1 == uint32` is not exposed)
- this means that for each call to `update_nblocks`, we can choose to use either the scalar32 or the shaext implementation of update_multi
- there is no inline assembly version of shaext exposed by Vale (so no file compilation issues)
- sha2-mb uses the same higher-order pattern

So, in essence, at the EverCrypt level, do the same as before but with Hacl.Impl.SHA2.Generic

```
val update_nblocks_shaext: Hacl.Spec.SHA2.Vec.update_nblocks_t' SHA2_256 M32
let update_nblocks_shaext ... =
  if hash_shaext
  ... call Vale's version ... reveal the state type representation (or change Vale implementation) ...
  else
    call SHA2.Scalar32.update_nblocks

val update_last_shaext: Hacl.Spec.SHA2.Vec.update_last_t

val hash_shaext: Hacl.Spec.SHA2.Vec.hash_t
```

then, cast these type signatures to the Hacl.Hash.Definitions types and use them as before in the rest of the module.

Would love to hear thoughts from @R1kM and @polubelova. Thanks!


msprotz:
@polubelova quick question... do you remember off the top of your head if the representation of the hash state between sha2-old and sha2-mb-scalar32 is identical?

in other words, if I do:
```
let st1 = Hacl.SHA2.Scalar32.create () in
Hacl.SHA2.Scalar32.init st1;
Hacl.SHA2.Scalar32.update st1 ...data...;
let st2 = Hacl.Hash.SHA2.create () in
Hacl.Hash.SHA2.init st2;
Hacl.Hash.SHA2.update st2 ...same data...;
```

is it the case that `st2` and `st1` contain exactly the same contents?

code/hash/Hacl.Hash.Definitions.fst
```diff
@@ -87,8 +88,11 @@ type state (i:impl) =
 
 inline_for_extraction noextract
 let as_seq (#i:impl) (h:HS.mem) (s:state i) : GTot (words_state (get_alg i)) =
-  match get_alg i with
-  | MD5 | SHA1 | SHA2_224 | SHA2_256 | SHA2_384 | SHA2_512 | SHA3_256 -> B.as_seq h s
+  let a = get_alg i in
+  match a with
+  | MD5 | SHA1 | SHA3_256 -> B.as_seq h s
+  | SHA2_224 | SHA2_256 | SHA2_384 | SHA2_512 ->
```
msprotz, line 94: TODO: add comment here and explain Seq.index 0

polubelova:
hi Jonathan, yes, I think so.
- for `init`: `Hacl.Spec.SHA2.Vec.init a M32` == `Hacl.Spec.SHA2.init a` == `Spec.SHA2.init a` relies on https://github.com/hacl-star/hacl-star/blob/main/code/sha2-mb/Hacl.Spec.SHA2.Equiv.fst#L295
- for `update`: https://github.com/hacl-star/hacl-star/blob/main/code/sha2-mb/Hacl.Spec.SHA2.EquivScalar.fsti#L13 and https://github.com/hacl-star/hacl-star/blob/main/code/sha2-mb/Hacl.Spec.SHA2.Equiv.fst#L434

 [`words_state`](https://github.com/hacl-star/hacl-star/blob/main/specs/Spec.Hash.Definitions.fst#L206) is the same for both `Hacl.Spec.SHA2` and `Spec.SHA2`, but [`k_w`](https://github.com/hacl-star/hacl-star/blob/main/specs/Spec.SHA2.fst#L33) is not.

msprotz:
Ah great. Then I think we can get away without too many drastic changes, provided I expose `vec_t U32 1 == uint32` through a `reveal_...`-style lemma. Not sure when I'll get back to this, there are a few spec-equivalence lemmas missing from the fsti for update_last, pad and finish...

karthikbhargavan:
Shall we try to make progress on this PR? It seems like a good idea. We can expose `vec_t t 1 == t` (for t <> U128) in `Lib.IntVector.fsti` but I don't know what else you will need for this PR.

karthikbhargavan:
I can help out a bit, can you summarize what needs to be done?

msprotz:
Yes. The most important things are:
- Revert my changes in Hacl.Hash.Definitions, per Marina's comment
- Go to Hacl.Hash.SHA2 and finish the work in there, notably, write all variants of update_last and finish to use the *existing* multibuffer implementation rather than generate new code via Hacl.Hash.*
  - This might require adding a few missing bits in `Hacl.Spec.SHA2.EquivScalar` and `Hacl.Spec.SHA2.Equiv`

Then, I can take care of the rest (streaming, evercrypt).

msprotz (commented):
Aymeric, this is super great. Thanks so much for helping out with this, I know how difficult these things are.

I'm mostly offline this week, so here are some quick comments (beyond: "this is super great").
- Why is `finish` relying on the old `Hacl.Hash` infrastructure?
- Can the proofs of `update_last` be shared?

Let's merge this in ASAP, then focus on:
- simplying Hacl.Streaming.MD
- having @pnmadelaine rename and re-bundle accordingly

R1kM:
Re: finish, I missed this one while finishing the proofs, I'll take care of it now.

Regarding update_last, I agree with you, but this is a bit more tricky. I tried to make a generic `mk_update_last`, but I did not manage to have verification going through. I tracked down the issue to the typechecking of the term `Lib.Sequence.repeat_blocks_multi (block_length SHA2_224) blocks1 (Hacl.Spec.SHA2.update SHA2_224) st0'`, currently appearing in calc.
Replacing the concrete `SHA2_224` by an abstract `a` leads to non-descriptive `assertion failed` messages.
Things I unsuccessfully tried:
* Moving this definition outside of the calc into a let-binding -> Typechecking still fails, even with all of the calc commented out and assumed
* Adding assertions about the lengths of the arguments
* Annotating all implicit parameters and types

providers/evercrypt/fst/EverCrypt.Hash.fst
```diff
@@ -413,7 +413,7 @@ let update_multi #a s prevlen blocks len =
 // Re-using the higher-order stateful combinator to get an instance of
 // update_last that is capable of calling Vale under the hood
```
msprotz, line 414: this comment is outdated and the function itself can probably go since it no longer serves its original purpose

providers/evercrypt/fst/EverCrypt.Hash.fst
```diff
@@ -410,8 +410,6 @@ let update_multi #a s prevlen blocks len =
 
 #pop-options
 
-// Re-using the higher-order stateful combinator to get an instance of
-// update_last that is capable of calling Vale under the hood
 let update_last_256 s prev_len input input_len =
```
msprotz, line 413: can we get rid of the definition entirely? or is there a reason (e.g. different `val`)?

R1kM: We might be able to, however, it is used in other places in EverCrypt (HMAC, Incremental). So far, I kept it to avoid a larger diff, but we could remove it altogether, and call directly Hacl.Hash.SHA2.update_last in these modules.
We could also remove the wrapper for update_last_224 since I don't think it brings much value now that we are not using the generic mk_update_last combinator anymore.
Wdyt?

msprotz: as long as Hacl.Hash.SHA2.update_last_224 is an alias for Hacl.Hash.SHA2.update_last_256 (or vice-versa), we're good -- the goal is to avoid code duplication, both in C and in the Low* source

R1kM: This is going through several layers of definitions, but it seems all of them are inlined to in the end yield `Hacl.SHA2.Scalar32.update_last_224`, which is itself defined as `Hacl.SHA2.Scalar32.update_last_256` with a proof of equivalence. I'll make the change.

dist/gcc-compatible/EverCrypt_HMAC.c
```diff
@@ -237,12 +238,15 @@ EverCrypt_HMAC_compute_sha2_256(
     uint8_t yi = key_block[i];
     opad[i] = xi ^ yi;
   }
-  uint32_t
-  s[8U] =
-    {
-      (uint32_t)0x6a09e667U, (uint32_t)0xbb67ae85U, (uint32_t)0x3c6ef372U, (uint32_t)0xa54ff53aU,
-      (uint32_t)0x510e527fU, (uint32_t)0x9b05688cU, (uint32_t)0x1f83d9abU, (uint32_t)0x5be0cd19U
-    };
+  uint32_t st[8U] = { 0U };
```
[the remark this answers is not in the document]
msprotz: wondering if it's an artifact of the new sha2-mb implementation? @polubelova?

in any case the C compiler should see both as being exactly the same, shouldn't it?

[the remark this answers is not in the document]
polubelova: yes, this change comes from the new sha2-mb implementation, where the [`init`](https://github.com/hacl-star/hacl-star/blob/9b477d2e8e4069ba387356e56ade96fa6ea89ecf/code/sha2-mb/Hacl.Impl.SHA2.Generic.fst#L179) function is written generically for all cases (`scalar32`, `vec128` and `vec256`)

code/hash/Hacl.Hash.SHA2.fst
```diff
@@ -1,45 +1,380 @@
...
+  reveal_vec_v_1 st.[2];
+  reveal_vec_v_1 st.[3];
+  reveal_vec_v_1 st.[4];
+  reveal_vec_v_1 st.[5];
+  reveal_vec_v_1 st.[6];
+  reveal_vec_v_1 st.[7];
+  reveal_vec_1 (word_t a);
+  eq_intro #(word a) #8 (Vec.state_spec_v st).[0] st
+
+let init_224 st =
```
polubelova, line 58: should we share the proofs between the four variants of `init`? same for `alloca`. 
should we move lemmas (including ones for `mk_update_multi`, `mk_update_last` and `mk_finish`) from this module to a standalone file?

dist/gcc-compatible/Hacl_Hash_SHA2.h
```diff
@@ -35,47 +35,7 @@ extern "C" {
...
-void
-Hacl_Hash_SHA2_update_last_512(
-  uint64_t *s,
-  FStar_UInt128_uint128 prev_len,
-  uint8_t *input,
-  uint32_t input_len
-);
+#include "Hacl_Streaming_SHA2.h"
 
 void Hacl_Hash_SHA2_hash_224(uint8_t *input, uint32_t input_len, uint8_t *dst);
```
polubelova, line 40: should we remove this file? all the one-shot hash functions are already defined in `Hacl_Streaming_SHA2.h`

dist/gcc-compatible/Hacl_Streaming_SHA2.c
```diff
@@ -28,7 +28,7 @@
 #include "internal/Hacl_SHA2_Generic.h"
 #include "internal/Hacl_Krmllib.h"
 
-static inline void sha256_init(uint32_t *hash)
+void Hacl_SHA2_Scalar32_sha256_init(uint32_t *hash)
```
polubelova, line 31: I guess `static inline` can be restored by removing a `Hacl_Hash_SHA2.h` file and fixing calls in `EverCrypt_`?

polubelova (approved):
Thanks, looks great to me!

Could you please also post the performance comparison numbers (`main` vs your branch)? Looking at the [performance regression tests](https://github.com/cryspen/hacl-packages/actions/runs/4961751330/jobs/8878940891#step:7:645), the performance improvement is about 20%. 

By the way, is [this job](https://github.com/cryspen/hacl-packages/actions/runs/4961751330/jobs/8878940891#step:7:1333) expected to fail?

dist/gcc-compatible/Hacl_Hash_SHA2.h
msprotz: I was thinking of doing this in a followup PR, along with bundling and everything + removing extra proofs in Hacl.Streaming.MD for SHA2 (no longer needed)... is that ok? it's just that this one has been open for a while. if you feel strongly that we might as well do everything all at once, why not, I'm ok with it, let me know

dist/gcc-compatible/Hacl_Streaming_SHA2.c
msprotz: that would still remove the `static inline` because there would be a cross-compilation unit call from EverCrypt_HMAC.c to Hacl_Hash_SHA2.c, meaning the init function would not be static inline

if there are performance concerns, we can always do

```
let sha256_init_public st = sha256_init st
```

and then EverCrypt_HMAC.c calls `sha256_init_public` (marked `extern`), while `sha256_init` remains `static inline`

what do you think?

dist/gcc-compatible/Hacl_Hash_SHA2.h
polubelova: yep, this can be done as a followup PR :)

code/hash/Hacl.Hash.SHA2.fst
R1kM: Good point thanks. I made generic `mk_init` and `mk_alloca`. Regarding the lemmas, it seems there only is `state_spec_v_lemma`. We could move it to another file, but since it's very specific to this correspondance, I'd be okay with keeping it here as well until we need it somewhere else. Wdyt?

dist/gcc-compatible/Hacl_Streaming_SHA2.c
R1kM: Should we fix `mk_compute` in this PR, or can it be done in a further PR?

code/hash/Hacl.Hash.SHA2.fst
polubelova: I meant moving the `calc` statements from code to lemmas. It is not a big deal but I prefer to separate proofs from stateful computations

dist/gcc-compatible/Hacl_Streaming_SHA2.c
polubelova: if this fix is quick/simple, we can do it in this PR

msprotz:
Benchmarks:

```
BEFORE

----------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations
----------------------------------------------------------------------------
HACL_Sha2_oneshot/sha2_224              4721 ns         4700 ns       135533
HACL_Sha2_oneshot/sha2_256              4648 ns         4641 ns       146570
HACL_Sha2_oneshot/sha2_384              3149 ns         3141 ns       228754
HACL_Sha2_oneshot/sha2_512              3043 ns         3040 ns       197184
HACL_Sha2_streaming/sha2_224            6018 ns         6003 ns       110954
HACL_Sha2_streaming/sha2_256            5944 ns         5936 ns       110770
HACL_Sha2_streaming/sha2_384            4070 ns         4064 ns       169970
HACL_Sha2_streaming/sha2_512            4132 ns         4126 ns       174251

AFTER

----------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations
----------------------------------------------------------------------------
HACL_Sha2_oneshot/sha2_224              4329 ns         4321 ns       165449
HACL_Sha2_oneshot/sha2_256              4265 ns         4255 ns       162655
HACL_Sha2_oneshot/sha2_384              2684 ns         2675 ns       277104
HACL_Sha2_oneshot/sha2_512              2519 ns         2514 ns       267969
HACL_Sha2_streaming/sha2_224            6032 ns         6025 ns       109642
HACL_Sha2_streaming/sha2_256            6166 ns         6147 ns        98388
HACL_Sha2_streaming/sha2_384            4519 ns         4507 ns       144641
HACL_Sha2_streaming/sha2_512            4333 ns         4328 ns       159947
```

Not sure why the performance regressed. Let me try to see if I can fix that static inline business first.

code/hash/Hacl.Hash.Definitions.fst
```diff
@@ -156,6 +156,43 @@ let hash_len (a: fixed_len_alg): n:size_t { v n = hash_length a } =
...
+///
+/// Note that we keep the total length at run-time, on 64 bits, but require that
+/// it abides by the size requirements for the smaller hashes -- we're not
+/// interested at this stage in having an agile type for lengths that would be
+/// up to 2^125 for SHA384/512.
+
+module U64 = FStar.UInt64
+
+inline_for_extraction noextract
+let max_input_len64 a: U64.(x:t { 0 < v x /\ v x `less_than_max_input_length` a }) =
```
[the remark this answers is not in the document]
msprotz: sadly we can't because the algorithm `a` is erased -- this is intentional, because the `hacl_keccak` works for *all* instances of keccak without doing any branching at run-time

(see line 120 of Hacl.Streaming.Keccak:

```
let hacl_keccak (a: G.erased alg): block alg =
  Block
    Erased
    stateful_keccak (* state *)
    (stateful_unused alg) (* key *)
    Lib.IntTypes.(x:size_t { v x > 0 }) (* output_length_t *)

    (fun _ -> 0xffffffffffffffffUL) (* max_input_len *)
```
)

I've therefore implemented one of your earlier suggestions:

```
    (fun _ -> Lib.IntTypes.(ones U64 PUB)) (* max_input_len *)
```

which is of course much more readable and better.

This also works:

```

    (fun _ ->
      [@inline_let]
      let max = Lib.IntTypes.(ones U64 PUB) in
      assert (forall (a: alg). Hacl.Hash.Definitions.max_input_len64 a == max);
      max) (* max_input_len *)
```

dist/gcc-compatible/Hacl_Hash_SHA2.h
msprotz: done in protz_sha2_thin_wrapper2

msprotz:
On branch protz_sha2_thin_wrapper2, which does bundling + restructuring to have everything in Hacl_Hash_SHA2.c (instead of having stuff spread across two files), we get the expected perf

```
----------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations
----------------------------------------------------------------------------
HACL_Sha2_oneshot/sha2_224              4120 ns         4117 ns       130853
HACL_Sha2_oneshot/sha2_256              4348 ns         4338 ns       164125
HACL_Sha2_oneshot/sha2_384              2596 ns         2588 ns       286928
HACL_Sha2_oneshot/sha2_512              2477 ns         2472 ns       285253
HACL_Sha2_streaming/sha2_224            5643 ns         5639 ns       113675
HACL_Sha2_streaming/sha2_256            5684 ns         5681 ns       110219
HACL_Sha2_streaming/sha2_384            4112 ns         4097 ns       174003
HACL_Sha2_streaming/sha2_512            3990 ns         3987 ns       168624
```

This is actually even better than before. So let me get a green on that.

dist/gcc-compatible/Hacl_Streaming_SHA2.c
msprotz: The fix is not simple. We do not have a way to say "I need any *instance* of the streaming functor that happens to exactly produce the hash functions". We can say "I want the MD streaming functor (non-agile, non-multiplexing)", or "I want the Evercrypt streaming functor (agile & multiplexing)", but we don't have a generic quantification that covers both.

I'm sure it's feasible, but seeing that I got good perf with just bundling stuff together, I'd rather not write super-complicated type signatures that will be hard to maintain in any case. Hope that's ok with you? Let me know!

code/hash/Hacl.Hash.SHA2.fst
R1kM: I'd usually be in favor, in this specific case though, I'm afraid about the amount of additional context that would need to be encoded in the Lemma (e.g., relations between `st`, `st'`, `blocks0` and `blocks0_multi`, ...). I'm happy to revisit later if it turns out to be unstable in the current state.

R1kM (approved):

R1kM:
#819 addresses the last concerns about performance. Let's merge this PR, and move the discussion to #819.

PR #771, opened 2023-02-08
Nix update

pnmadelaine:
This PR introduces some changes to the Nix code.

- It adds a dependency to the F* flake, which was merged upstream.
- It moves the expressions for Vale and Karamel to this repository.
- Thus it removes the dependency to hacl-star/hacl-nix and the ugly mutual reference between here and there.
- It filters the source before building, which should avoid some unnecessary rebuilds (for instance an update in `dist/` or in `.github/` will not trigger a full rebuild in Hydra).

Notably, this will help clarify the role of hacl-star/hacl-nix, which sole purpose will become to generate daily lockfiles for Hacl*.
As a future update, it could be nice to move the Nix code for Karamel in the Karamel repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,5 +31,4 @@ dist/wasm/*.wast
 doc/_build
 config.h
 *.smt2
-flake.lock
 *.runlim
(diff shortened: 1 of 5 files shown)


pnmadelaine:
once again Hydra failed to report the green status, but the branch passed CI: https://everest-ci.paris.inria.fr/jobset/hacl-star/branch-pnmadelaine-nix#tabs-jobs

msprotz:
>  As a future update, it could be nice to move the Nix code for Karamel in the Karamel repository.

I'd be happy to approve a PR for that. As you know, I've always been a big fan of Nix since day 1 ;-)

PR #773, opened 2023-02-08
Cleanups for SHA2 in preparation of the next Python PR

msprotz:
This is basically replicating some of the work that I did for sha2-224/256 onto sha2-384/512:
- add a copy function
- write spec equivalence proofs so as to enable code sharing between sha2-384 and sha2-512

This is all in preparation for https://github.com/python/cpython/pull/101707


msprotz:
Sadly, there's a lot of whitespace changes because of the latest krml changes

R1kM (approved):
Thanks!

polubelova (approved):
Thanks!

PR #775, opened 2023-02-14
F*+dune: New layout for F* libs, improved handling of FSTAR_HOME

tahina-pro:
## Proposed changes

This PR is a companion to FStarLang/FStar#2815
With F* built with dune, the F* libraries are organized differently: fstar-lib, fstar-tactics-lib and fstar-compiler-lib are all merged into a single library, fstar.lib, which is now installed in `$FSTAR_HOME/lib/fstar/lib/fstar_lib.cmx*` instead of `$FSTAR_HOME/bin/*.cmx*` . Thus, this PR fixes invocations of OCaml compilers/linkers appropriately.

In this process, this PR also improves the handling of the `FSTAR_HOME` environment variable to allow users having installed F* as an opam package (or with `make install`), with `fstar.exe` in the `PATH`, to verify, extract and compile HACL* without setting `FSTAR_HOME`. (In particular, in such installations, the F* library sources are in `lib/fstar` instead of `ulib`.)

Once this PR is approved, I will merge the corresponding F* and Karamel PRs (FStarLang/karamel#333) first, before merging this one.

## Types of changes

- [x] Build update (to adapt to ecosystem changes, e.g. F*, Karamel, etc.)

## Checklist

## Further comments

I have an everest branch reporting green CI with these F*, Karamel and HACL* PRs, cf. project-everest/everest@925043549b2573160e93a23fab3abd10f969e70a

This PR will likely interfere with #771 , heads up @W95Psp, @pnmadelaine and other Nix specialists.

Thank you all in advance!


msprotz (changes requested):
I have reservations about the part regarding file paths. In short:
- in the current state of things, F* generates paths relative to the bin/ subdirectory in .fstar-depend-full, even though the initial invocation of fstar --depend does not *receive* anything related to the bin/ directory, since the arguments go through $(notdir)
- this is bad for make since it generates a "double-vision" problem where two identical files are referred to different paths

I don't see how your change to the sed invocation helps... can you explain a little?

My hunch is that you need to establish the invariant that your FSTAR_HOME, regardless of which way it's computed, is always fed through $(realpath) (to avoid those issues) AND keep the sed as-is (to workaround what looks like an F* bug in the generation of .fstar-depend-full)

Thanks

Makefile
```diff
@@ -280,7 +280,7 @@ ifndef MAKE_RESTARTS
 	  $(FSTAR_NO_FLAGS) --dep $* $(notdir $(FSTAR_ROOTS)) --warn_error '-285' $(FSTAR_DEPEND_FLAGS) \
 	    --extract 'krml:*' \
 	    --extract 'OCaml:-* +Vale.Arch +Vale.X64 -Vale.X64.MemoryAdapters +Vale.Def +Vale.Lib +Vale.Bignum.X64 -Vale.Lib.Tactics +Vale.Math +Vale.Transformers +Vale.AES +Vale.Interop +Vale.Arch.Types +Vale.Arch.BufferFriend +Vale.Lib.X64 +Vale.SHA.X64 +Vale.SHA.SHA_helpers +Vale.SHA2.Wrapper +Vale.SHA.PPC64LE.SHA_helpers +Vale.PPC64LE +Vale.SHA.PPC64LE +Vale.Curve25519.X64 +Vale.Poly1305.X64 +Vale.Inline +Vale.AsLowStar +Vale.Test +Spec +Lib -Lib.IntVector -Lib.Memzero0 -Lib.Buffer -Lib.MultiBuffer +C -C.String -C.Failure' > $@.tmp && \
-	  $(SED) 's!$(HACL_HOME)/obj/\(.*.checked\)!obj/\1!;s!/bin/../ulib/!/ulib/!g' $@.tmp && mv $@.tmp $@ \
```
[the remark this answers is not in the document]
tahina-pro: Thanks for your description! Here are the 2 reasons of this particular change here:
1. In sed, `.` will match any character, as opposed to `\.`, which will exactly match `.`
2. If the user installs F* via opam, or via `make install` (or with the binary package, in the case of the F*+dune PR), then `ulib` is replaced with `lib/fstar`, so I need to take into account `/bin/../lib/` to be turned into `/lib/`. Thus, I decide to turn wholesale `/bin/../` into `/`, regardless of what follows.

Makefile.common
```diff
@@ -15,14 +15,33 @@ HACL_HOME 	?= $(shell pwd)
...
+    FSTAR_HOME := $(dir $(FSTAR_EXE))/..
+  endif
+  ifeq ($(OS),Windows_NT)
+    FSTAR_HOME := $(shell cygpath -m $(realpath $(FSTAR_HOME)))
+  else
+    FSTAR_HOME := $(realpath $(FSTAR_HOME))
+  endif
+endif
+
+FSTAR_ULIB ?= $(shell if test -d "$(FSTAR_HOME)/ulib" ; then echo "$(FSTAR_HOME)/ulib" ; else echo "$(FSTAR_HOME)/bin/fstar" ; fi)
```
tahina-pro, line 35: Oops, this should be `$(FSTAR_HOME)/lib/fstar`
```suggestion
FSTAR_ULIB ?= $(shell if test -d "$(FSTAR_HOME)/ulib" ; then echo "$(FSTAR_HOME)/ulib" ; else echo "$(FSTAR_HOME)/lib/fstar" ; fi)
```

tahina-pro:
After further discussions, we decided to explicitly sanitize all of `FSTAR_HOME`, `KRML_HOME`, `VALE_HOME` and `HACL_HOME` with their absolute paths, using their Windows names on Cygwin. I just pushed the corresponding changes.

Makefile.common
```diff
@@ -8,21 +8,62 @@ define newline
...
+endif
+
+ifndef FSTAR_ULIB
+FSTAR_ULIB := $(shell if test -d "$(FSTAR_HOME)/ulib" ; then echo "$(FSTAR_HOME)/ulib" ; else echo "$(FSTAR_HOME)/lib/fstar" ; fi)
+endif
+
+ifndef VALE_HOME
+# assuming an Everest source tree
+VALE_HOME	:= $(HACL_HOME)/../vale
+endif
```
msprotz, line 42: is the ifndef syntax better than `?=`?

Makefile.common
```diff
@@ -8,21 +8,62 @@ define newline
...
+FSTAR_HOME := $(realpath $(FSTAR_HOME))
+FSTAR_ULIB := $(realpath $(FSTAR_ULIB))
+VALE_HOME  := $(realpath $(VALE_HOME))
+ifeq ($(OS),Windows_NT)
+  HACL_HOME  := $(shell cygpath -m $(HACL_HOME))
+  KRML_HOME  := $(shell cygpath -m $(KRML_HOME))
+  FSTAR_HOME := $(shell cygpath -m $(FSTAR_HOME))
+  FSTAR_ULIB := $(shell cygpath -m $(FSTAR_ULIB))
+  VALE_HOME  := $(shell cygpath -m $(VALE_HOME))
+endif
```
msprotz, line 59: I'm generally not a big fan of GNU make functions, but this might be one of the few cases where they might actually make the code more readable, see https://www.gnu.org/software/make/manual/html_node/Call-Function.html -- it's a small nit

msprotz (approved):
This is as discussed, and looks good to me. Thanks!

I left a couple minor comments; up to you to figure out whether they would help readability or not.

Makefile.common
tahina-pro: From https://www.gnu.org/software/make/manual/html_node/Conditional-Assignment.html: `FOO ?= bar` is equivalent to:
```
ifeq ($(origin FOO), undefined)
  FOO = bar
endif
```
which uses `=` instead of `:=`, thus preventing overwriting the variable when I sanitize it on lines 56 sqq.

Makefile.common
tahina-pro: > I'm generally not a big fan of GNU make functions, but
That's exactly what I was expecting! Let me fix that now.

Makefile.common
msprotz: Great. Thanks for clarifying!

Makefile.common
tahina-pro: Done in 03543fdc3e5a80a269eb079c7afb455690abc77f

msprotz:
Ok looks great to me. There's still one unresolved conversation, above, so I'll let you mark it as resolved (I think you fixed this particular issue in a subsequent commit.) Thanks for all of the work catching up on these changes! Much appreciated

tahina-pro:
Thank you again Jonathan for your feedback!

As of now, this PR is the last piece blocking the merging of the F*+dune PR. While I have an everest green, the Hydra CI is broken here because I don't know how to fix the Nix CI to make use of the corresponding F*+dune branches on F* and Karamel. @pnmadelaine @W95Psp and other Nix experts, how should we proceed? Thank you all in advance!

msprotz:
I'm happy to override CI on this one if you have an Everest green

pnmadelaine:
to pass the Hydra CI we can temporarily override the F* revision used in the Nix expressions, wait for the PR to be merged on the F* side, then remove the override

pnmadelaine:
done ! I took the liberty to merge #771 in this PR since it required to update karamel's derivation anyway.

flake.nix
```diff
@@ -2,35 +2,33 @@
   description = "Hacl*";
 
   inputs = {
-    fstar-src = {
-      url = "github:fstarlang/fstar";
-      flake = false;
-    };
+    fstar.url = "git+https://github.com/fstarlang/fstar?ref=_taramana_dune";
```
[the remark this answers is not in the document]
msprotz: let's coordinate:
- Tahina merges F* and krml PRs
- we push a commit here that removes the specific reference to _taramana_dune
- get a green here, merge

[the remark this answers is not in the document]
tahina-pro: Yes, I plan to merge today, in this order: the F* PR first, then the Karamel PR, then this PR. Thus, when merging the latter two, I can in turn remove the corresponding branch references.

Thank you PNM for your fixes to the HACL* CI and thanks again Jonathan and Aymeric for approving!

PR #776, opened 2023-02-14
format C code from `tests/`

polubelova:
Format C code from `tests/` with `clang-format --style=Mozilla *.c *.h -i`.


msprotz (changes requested):
Can we add a target to automate this? e.g. `make format` or something, otherwise, this is going to slowly degrade over time... other than that looks good... thanks!

msprotz (approved):

PR #777, opened 2023-02-16
[PowerPC] Implement optimized AES-GCM on Vale

mamonet:
This patch adds optimized implementation of AES-GCM mode on Vale. The general scheme follows x86_64 modules as usual, however it uses big-endian definitions of almost everything so the patch re-does AES helper modules in big-endian mode.
Unlike Arm and x86, ppc64le only applies little-endian byte order on memory so once the data is loaded into registers all operations perform in big-endian.

The core implementation of GCTR uses 6-way AES block encryption which get the optimal performance out of POWER9 architecture. While GHASH takes advantage of intel papers [1] [2] to enable high-performance implementation, same specifications are implied in OpenSSL implementation by the way of both x64 and ppc64le.

This patch adds `vale/code/crypto/aes/ppc64le/aes-gcm.h` and `vale/code/crypto/aes/ppc64le/TestAesGcm.cpp` for testing example.
The implementation achieves 2.2 cycles/bytes for AES-GCM128 and 1.14 cycles/bytes for GHASH.


mamonet:
@Chris-Hawblitzel @R1kM this PR needs approval of at least 2 reviewers, would you like to take a look at?

R1kM:
Thanks for the heads-up @mamonet , I'll have a look towards the end of the week.

Chris-Hawblitzel:
I'm looking at it too.  One question.  We put code ported from OpenSSL in https://github.com/hacl-star/hacl-star/tree/main/vale/code/thirdPartyPorts/OpenSSL to make sure that it was credited to OpenSSL and marked with the appropriate license.  Does any of the code in this pull request also belong under the `OpenSSL` directory for this reason?  This would include code that was derived from code in https://github.com/hacl-star/hacl-star/tree/main/vale/code/thirdPartyPorts/OpenSSL, and thus was indirectly derived from OpenSSL code (unless it was pure proof code, which we wrote ourselves).

mamonet:
Thank you both.

> Does any of the code in this pull request also belong under the `OpenSSL` directory for this reason?

No, the code in this PR isn't derived from OpenSSL by any means. In fact, I previously implemented AES-GCM mode for PowerPC on GNU nettle and NSS libraries with my own copyright and both have been implemented with a different GHASH algorithm.

OpenSSL implementation for ppc64le uses densely hardware pipelining tricks for GHASH part which I don't see it's much appropriate to be utilized here or maybe at all as it's low-key obfuscating the code for the reader and yields a minor performance improvement. One could use more enhanced algorithm for GHASH to offset the performance speed up but this is a different subject besides this PR uses intel white papers as explicit reference.

vale/code/arch/ppc64le/Vale.PPC64LE.InsBasic.vaf
```diff
@@ -124,7 +124,17 @@ procedure LoadImm64(out dst:reg_opr, inline src:simm16)
     {:quick exportOnly}
     {:instruction Ins(S.LoadImm64(dst, src))}
     ensures
-        dst == old(int_to_nat64(src));
+        dst == old(src % pow2_64);
```
Chris-Hawblitzel, line 127: This is a minor comment, but if you can use addition and subtraction in place of mod-large-number for the `.vaf` instruction definitions, it will sometimes speed up verification.  This is why `Add64Wrap` uses `add_wrap`, defined as `if x + y < n then x + y else x + y - n`, rather than `(x + y) % pow2_64`.

Chris-Hawblitzel (approved):

vale/specs/crypto/Vale.AES.AES_BE_s.fst
```diff
@@ -0,0 +1,157 @@
+module Vale.AES.AES_BE_s
+
+// IMPORTANT: This specification is written assuming a little-endian mapping from bytes to quad32s
+//            This is explicit in key_schedule_to_round_keys when we construct the round_key rk,
```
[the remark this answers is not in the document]
mamonet: You're right, these comments need to be edited on big-endian modules.

vale/specs/defs/Vale.Def.Types_helpers_BE.fsti
```diff
@@ -0,0 +1,126 @@
...
+open FStar.Seq
+open FStar.Math.Lemmas
+open Vale.Lib.Seqs
+open Vale.Lib.Seqs_s
+open Vale.Def.Words.Seq
+
+unfold let pow2_24 = 0x1000000
+let nat24 = natN pow2_24
+
+val seq_four_to_seq_to_seq_four_BE (#a:Type) (x:seq a{length x % 4 == 0}) :
```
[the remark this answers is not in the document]
mamonet: Done! Moved to `vale/code/crpto/aes` directory, just a little concerned if we need to utilize these lemmas by future implementations but we can figure that out until then.

vale/code/arch/ppc64le/Vale.PPC64LE.InsBasic.vaf
mamonet: I was missing the usage of `sub_wrap` for subtraction instructions so I replaced the old procedure in the new commit. Regarding mod-large-number of inline integers, I can replace them with `nat` but we will omit the boundaries that have the same behavior of checking instruction parameters by assembler at compile-time https://github.com/hacl-star/hacl-star/blob/36ab944703a52b52218bce295614d15fc09b22f3/vale/code/arch/ppc64le/Vale.PPC64LE.InsBasic.vaf#L109-L110

vale/code/crypto/aes/Vale.AES.Types_helpers_BE.fsti
```diff
@@ -0,0 +1,126 @@
...
+open FStar.Seq
+open FStar.Math.Lemmas
+open Vale.Lib.Seqs
+open Vale.Lib.Seqs_s
+open Vale.Def.Words.Seq
+
+unfold let pow2_24 = 0x1000000
+let nat24 = natN pow2_24
+
+val seq_four_to_seq_to_seq_four_BE (#a:Type) (x:seq a{length x % 4 == 0}) :
```
R1kM, line 19: It seems that there are similar lemmas already in, e.g., Vale.Arch.Types and Vale.Def.Words.Seq. Could we consolidate all lemmas on sequences and vale types there to avoid code duplication in the future?

vale/code/crypto/aes/ppc64le/Main.ml
```diff
@@ -0,0 +1,26 @@
+let parse_cmdline :
```
R1kM, line 1: Could you put this with the cmdline parsing for x64 in code/lib/util/CmdLineParser.ml, and rename both functions in, say, parse_cmdline_x64 and parse_cmdline_ppc64le?

vale/specs/crypto/Vale.AES.AES_BE_s.fst
```diff
@@ -0,0 +1,157 @@
...
+assume val inv_sub_bytes (q:quad32) : quad32
+assume val shift_rows (q:quad32) : quad32
+assume val inv_shift_rows (q:quad32) : quad32
+assume val rot_word (w:nat32) : nat32
+assume val sub_word (w:nat32) : nat32
+
+assume val commute_rot_word_sub_word (x:nat32) (rcon:nat32) : Lemma
+  (rot_word (sub_word x *^ ishl32 rcon 16) == sub_word (rot_word x) *^ rcon)
+
+type algorithm:eqtype = | AES_128 | AES_192 | AES_256
```
R1kM, line 32: Many definitions in this module seem copy-pasted from the Vale.AES.AES_s. Could we merge the two files in one, using _LE/_BE suffixes when there is a difference? If not possible, could we have common definitions (e.g., algorithm type) in a single file?

R1kM (changes requested):
Sorry for the delay.
This looks overall good to me, I left a few comments in addition to those that Chris mentioned.

Thanks!

vale/code/crypto/aes/Vale.AES.AES256_helpers_BE.fst
```diff
@@ -0,0 +1,88 @@
...
+let lemma_reveal_expand_key_256 (key:aes_key_word AES_256) (round:nat) : Lemma
+  (expand_key_256 key round == (
+    if round = 0 then Mkfour key.[3] key.[2] key.[1] key.[0]
+    else if round = 1 then Mkfour key.[7] key.[6] key.[5] key.[4]
+    else round_key_256 (expand_key_256 key (round - 2)) (expand_key_256 key (round - 1)) round
+  ))
+  =
+  expand_key_256_reveal ()
+
+#reset-options "--initial_fuel 8 --max_fuel 8 --max_ifuel 0"
```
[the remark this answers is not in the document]
mamonet: The change makes sense. Edited for this PR!

vale/code/crypto/aes/Vale.AES.OptPublic_BE.fsti
```diff
@@ -0,0 +1,6 @@
+module Vale.AES.OptPublic_BE
+
+open FStar.Mul
+open Vale.Def.Types_s
+
+val hkeys_reqs_pub (hkeys:FStar.Seq.seq quad32) (h_BE:quad32) : Vale.Def.Prop_s.prop0
```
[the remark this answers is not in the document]
mamonet: The other two propertied used by specific EverCrypt modules which not implemented for ppc64le arch.

vale/code/crypto/aes/Vale.AES.Types_helpers_BE.fsti
mamonet: I agree. Lemmas consolidated now to Vale.Arch.Types and Vale.Def.Words.Seq modules.

vale/code/crypto/aes/ppc64le/Main.ml
mamonet: Done!

vale/specs/crypto/Vale.AES.AES_BE_s.fst
mamonet: Done!

mamonet:
@R1kM Do you have any idea why CI fails?

R1kM:
@pnmadelaine it seems that the tests on hacl-packages are not supported from forks of the repo? Also, the Hydra CI reported evaluation errors on the Nix script: https://everest-ci.paris.inria.fr/eval/21515#tabs-errors

R1kM:
@mamonet Thanks for your patience. We'll look at what is happening on the CI front. The PR seems in good shape, if I can manage to ensure that at least verification works on the CI machine, I'll happily override the checks to merge this.
Could I let you take care of the conflicts by removing changes to hints? Thanks!

mamonet:
@R1kM Sure. I couldn't manage to discard the changes of hints so I pushed the patch again without hint files.

mamonet:
@R1kM Thanks for your review. I added the generated `aesgcm-ppc64le.S` in dist/gcc-compatible directory.

PR #778, opened 2023-02-16
speeding up sha3

karthikbhargavan:
## Proposed changes

Step 1: avoiding a state copy in chi. 
Step 2: removing calls to salloc1 and replacing them with create.
Step 3: marking local constant arrays as "private"

Of these, we should revisit Step 2 when we have a systematic approach towards zeroing out stack-allocated values.
In the current code, almost all the state is in registers so the value of memzero is unclear.

The result of these changes is a 200% perf improvement when using clang on Linux and Mac OS. 
With GCC, there is a more modest performance improvement of 25-30%.
We will try to reconcile GCC and clang performance over time.

## Types of changes

- [X] Improved performance (fix or feature that would improve performance of some algorithm on some platform)

## Checklist

- [X] I have read the [CODE_OF_CONDUCT.md](https://github.com/hacl-star/hacl-star/blob/main/CODE_OF_CONDUCT.md) doc
- [X] I have read and agree to submit my changes under the [LICENSE](https://github.com/hacl-star/hacl-star/blob/main/LICENSE)

## Further comments

This a series of optimizations that close the performance gap between our SHA3 implementation and other C implementations. There remains a gap between this code and implementations that use AVX2 SIMD instruction.


polubelova (approved):
looks great, thanks!

dist/gcc-compatible/Hacl_SHA3.h
```diff
@@ -35,16 +35,6 @@ extern "C" {
...
-
-extern const uint32_t Hacl_Impl_SHA3_keccak_rotc[24U];
-
-extern const uint32_t Hacl_Impl_SHA3_keccak_piln[24U];
-
-extern const uint64_t Hacl_Impl_SHA3_keccak_rndc[24U];
-
-uint64_t Hacl_Impl_SHA3_rotl(uint64_t a, uint32_t b);
-
 void Hacl_Impl_SHA3_state_permute(uint64_t *s);
```
msprotz, line 38: that file is not bundled behind an interface (see my comment in the other PR as to why), so you probably would benefit from adding `private` qualifiers in state_permute, loadState and storeState

dist/gcc-compatible/internal/Hacl_SHA3.h
```diff
@@ -0,0 +1,52 @@
+/* MIT License
```
msprotz, line 1: you might want to wait for my other PR to land first, as it will make this file disappear, I think

specs/lemmas/Spec.SHA3.Equivalence.fst
```diff
@@ -0,0 +1,102 @@
+module Spec.SHA3.Equivalence
```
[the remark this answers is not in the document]
msprotz: maybe, there's also a precedent for putting a lot of spec equivalence lemmas in here

msprotz (changes requested):
Looks good minus my comment about marking a couple more things private. Let's wait for #780 to land first, as I think it'll avoid creating a new file

specs/lemmas/Spec.SHA3.Equivalence.fst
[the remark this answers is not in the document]
karthikbhargavan: I don't mind either way. I could turn Spec.SHA3.Equivalence into code/sha3/Hacl.Spec.Sha3 or I could leave it here.

dist/gcc-compatible/Hacl_SHA3.h
karthikbhargavan: I think permute is called from the streaming APIs and it complains when I turn it private. I guess it would be better to make these functions private and instead expose "absorb_inner" which functions as "update-1-block" for SHA3?

dist/gcc-compatible/internal/Hacl_SHA3.h
karthikbhargavan: Oh year, I was wondering why this file was being generated. Happy to have it go away.

dist/gcc-compatible/Hacl_SHA3.h
msprotz: curious why streaming calls these functions... wondering if something is inlined too aggressively

msprotz: yeah so the reason is because streaming calls absorb_inner to do the loop over a multiple of a number of blocks, and that one is inline_for_extraction noextract, which means the implementation of SHA3 is replicated in the streaming version and hence generates cross-translation unit calls to permute

my suggested fix is to remove inline_for_extraction noextract from `absorb_inner`, and wait for my other PR to land so that you won't have a performance hit (because as streaming_sha3 and sha3 are merged together into a single file, `absorb_inner` will be marked as `static` and hopefully will happily optimized just as well)

alternatively, you could do `let absorb_inner_non_inline = absorb_inner`, then call `absorb_inner_non_inline` from `Hacl.Hash.SHA3`, which will remove the copy of `absorb_inner` inlined everything in Hacl.Streaming.SHA3, and will allow you to mark permute as private

karthikbhargavan:
@msprotz is there anything other than waiting for the other PR to land?

msprotz (approved):
No that's good for me. Just waiting on Franziskus to approve the other PR. It might impact hacl-packages so I want his green light.

dist/gcc-compatible/internal/Hacl_SHA3.h
msprotz: yeah the file is still there because EverCrypt.Hash repeatedly calls `absorb_inner` in the update_multi loop, and this is a call across translation units, so it generates an internal header

(I forgot about that caller)

msprotz: I'll merge this PR, but if the internal header is bad, and you need loadState to be marked as `static` for perf reasons, open up an issue and I'll fix it

PR #780, opened 2023-02-16
Code quality, bundling, etc.

msprotz:
This one is a slightly more involved change. It aims to fix three things:
- first, there were undesirable cross-algorithm dependencies... for instance, `Hacl_Streaming_MD5.h` used to contain `#include "internal/Hacl_Streaming_SHA2.h"` because they both rely on the same type definition for their internal state, and it so happened that krml would insert the monomorphized type in SHA2 first, then MD5 would depend on it -- this is bad for modularity!
- second, code was scattered across various files, without a consistent naming convention... I know that @pnmadelaine and others are trying to fix this, so I fixed many of those little inconsistencies, while I was it... notably
  - md5 is now consolidated into a single file, Hacl_Hash_MD5, which contains the streaming API + the one-shot API, but hides the internal block API
  - same deal for sha1, sha3
- third, and this is a drive-by change, but I got rid of Hacl_SHA2_Types.h (now there's only the internal header)

Note: sha2 remains split between Hacl_Hash_SHA2 (legacy, still used by EverCrypt_Hash, see #756), and Hacl_Streaming_SHA2 (modern, eventually will be renamed into Hacl_Hash_SHA2).

The streaming state types are now in their own file: https://github.com/hacl-star/hacl-star/pull/780/files#diff-ccb48f7ad898e93510fca542c058834a92fc421a92c66cc372d36a1e684716fa

**Just to be clear**: this changes file names and function names.

This is a somewhat more involved change, so in addition to the usual reviews, I would love some feedback from @pnmadelaine and @franziskuskiefer 

Thanks!


dist/gcc-compatible/Hacl_Hash_SHA1.c
```diff
@@ -26,6 +26,7 @@
 #include "internal/Hacl_Hash_SHA1.h"
 
 #include "internal/Hacl_Krmllib.h"
+#include "internal/Hacl_Hash_MD5.h"
```
msprotz, line 29: this should not be here and is currently breaking the Mozilla build

msprotz: Fixed via https://github.com/FStarLang/karamel/pull/336

msprotz:
This should be green once the krml fix is in. I'll wait until then to hit "update branch".

polubelova (approved):
Thanks!

franziskuskiefer (approved):
Thanks. This is a great step in the right direction.
Here's a PR on hacl-packages to adapt to these changes https://github.com/cryspen/hacl-packages/pull/363

PR #784, opened 2023-02-23
Remove bindings

pnmadelaine:
## Proposed changes

This PR removes the OCaml and JS bindings since they were moved to `cryspen/hacl-packages`.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] I have edited [CHANGES.md](https://github.com/hacl-star/hacl-star/blob/main/CHANGES.md) (if appropriate)


doc/Obtaining.rst
```diff
@@ -88,43 +88,12 @@ and other HACL* users have also developed bindings for other programming languag
...
-``opam install hacl-star-raw``.
-
-From source, the low-level bindings can be installed by running
-``make install-hacl-star-raw`` in ``dist/gcc-compatible``. Then, the high-level
-wrapper can be installed by running ``dune build && dune install`` in
-``bindings/ocaml``.
-
-Documentation for the high-level API is available `here
-<https://hacl-star.github.io/ocaml_doc/>`_.
+OCaml bindings were moved to the `cryspen/hacl-packages` repository.
```
franziskuskiefer, line 91: There's https://tech.cryspen.com/hacl-packages/hacl-js/index.html and https://tech.cryspen.com/hacl-packages/hacl-ocaml/index.html.
Can you add documentation that you think makes sense there?

dist/gcc-compatible/configure
```diff
@@ -227,7 +227,6 @@ show_help () {
   printf "                    * aarch64-apple-ios\n"
   printf "                    * x86_64-apple-ios-simulator\n"
   printf "    --disable-bzero Do not use explicit_bzero (binary will work with an old GLIBC)\n"
-  printf "    --disable-ocaml Disable OCAML bindings\n"
```
franziskuskiefer, line 230: I'm not sure whether some folks would prefer keeping the option to build the raw package here.
If yes, the `hacl-star-raw.opam` should be thinned out to make clear that it's not the real one and we should keep the infrastructure here.
But I'd be fine with removing this here as you've done.

franziskuskiefer (approved):
Thanks, this generally looks good to me with an open question and a follow up.

doc/Obtaining.rst
pnmadelaine: I think the links you provided should be the entry points to the documentation, and we should add anything relevant there. Leaving anything in the hacl-star repository would require maintenance.

franziskuskiefer: Yeah I agree. I just thought there might be interesting information here that you remove and that you could add to the two linked pages.

Makefile
```diff
@@ -93,8 +93,8 @@ endif
 all: all-staged
 
 all-unstaged: compile-gcc-compatible compile-msvc-compatible \
-  compile-portable-gcc-compatible \
-  dist/wasm/package.json obj/libhaclml.cmxa
+  compile-portable-gcc-compatible compile-wasm \
```
msprotz, line 96: this is going to call gcc, isn't it? see rule for `compile-%`

why not as before and take a dependency on dist/wasm/Makefile.basic, to force generation, but not compile in C?

pnmadelaine: Thanks for the note, I replaced it with `dist/wasm/Makefile.basic`

dist/gcc-compatible/configure
pnmadelaine: I don't know, @msprotz what do you think?

msprotz: I don't have a strong opinion either way... my hunch is if I'm working on the bindings I'll almost certainly end up with hacl-star + hacl-packages side-by-side with my testing loop doing build hacl, update hacl in packages, and build packages... so I don't think it makes sense to maintain an old build system in the hacl-star repository 

let's just make sure we're all very good about catching breakages... if the situation changes, we can always reintroduce some amount of build in HACL*

msprotz (approved):
lgtm 👍

PR #795, opened 2023-03-27
clean up specs/tests

polubelova:


msprotz (approved):
Thanks. At some point I'd be curious to know if there's any way to share code more aggressively between these tests. But agree that this cleanup is a good first step. Thanks!

PR #797, opened 2023-04-03
Fix endianness issue in AIX

ayappanec:
AIX (always a big-endian) doesn't have `__BYTE_ORDER__` defined in any of its header files. GCC for this reason defines `__BYTE_ORDER__` in its own header files for AIX. But with XLC compiler, it's not the case. 
The issue is exposed while compiling the latest python --> https://github.com/python/cpython/pull/102206/

### DIFF
--- a/dist/karamel/include/krml/lowstar_endianness.h
+++ b/dist/karamel/include/krml/lowstar_endianness.h
@@ -96,7 +96,8 @@
 #  define le64toh(x) (x)
 
 /* ... generic big-endian fallback code */
-#elif defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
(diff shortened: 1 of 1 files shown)


msprotz:
Thanks for the good catch!

This file comes from an upstream project. Can you submit a PR for this file? https://github.com/FStarLang/karamel/blob/master/include/krml/lowstar_endianness.h I'll approve it immediately and then we can make it trickle down to hacl-star.

Thanks so much!

ayappanec:
> Thanks for the good catch!
> 
> This file comes from an upstream project. Can you submit a PR for this file? https://github.com/FStarLang/karamel/blob/master/include/krml/lowstar_endianness.h I'll approve it immediately and then we can make it trickle down to hacl-star.
> 
> Thanks so much!

I just created a PR in the upstream project.

msprotz:
Fixed upstream.

PR #800, opened 2023-04-11
Addition of all Keccak variants to agile hash specs; addition of all SHA3 variants to low-level implementations

msprotz:
This PR adds all Keccak variants (SHA3 and Shake) to the agile hash infrastructure. This requires:
- extending the hash_alg type with five more variants
- adding a notion of variable length, which for now is leveraged by shake only -- to minimize the diff, many definitions continue to operate on what I dubbed "const hashes" (constant output length) and more general variants are added with a suffix for those few parts of the code that can truly deal with a variable-length hash function
- adapting the rest of the specs.

Adding shake into the agile hash infrastructure allows us to prove trivially (by generalizing the previous sha3-256 proof) that shake obeys the hash laws (which will be useful for my next PR).

This PR adds SHA3 variants (not Shake) to the low-level code, meaning that all SHA3 flavors are now available in EverCrypt hash and could be conceivably used anywhere else hashes are currently used (such as HMAC, HKDF, or whatever else).

This PR does *not* extend the streaming API to deal with more SHA3 variants. This will be the subject of a forthcoming PR with a dedicated streaming module that can deal with Keccak in a general way.

As a drive-by change, I identified a proof performance issue in HPKE, and Aymeric (many thanks!) helped me finish a proof that drives the verification time of Hacl.Meta.HPKE from 20 minutes down to 7 on my laptop. Other setups might differ.

This is the first half of the work for getting SHA3 into Python, see https://github.com/python/cpython/issues/99108


dist/gcc-compatible/EverCrypt_Hash.h
```diff
@@ -85,6 +85,24 @@ EverCrypt_Hash_Incremental_update(
   uint32_t len
 );
 
+void
+EverCrypt_Hash_Incremental_finish_sha3_256(
```
polubelova, line 89: why do we need to expose these functions? aren't they covered by `EverCrypt_Hash_Incremental_finish`?

msprotz: good catch, I neglected to add some `private` qualifier... fixed by my latest commit

specs/Spec.Hash.Definitions.fst
```diff
@@ -62,6 +72,7 @@ let sha2_alg = a:hash_alg { is_sha2 a }
 let sha3_alg = a:hash_alg { is_sha3 a }
 let blake_alg = a:hash_alg { is_blake a }
 let md_alg = a:hash_alg { is_md a }
+let const_alg = a:hash_alg { not (is_shake a) }
```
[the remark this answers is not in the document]
polubelova: i don't know, maybe `fixed_len_alg`?

polubelova (approved):
Thanks!

code/hash/Hacl.Hash.Definitions.fst
```diff
@@ -19,22 +19,22 @@ open FStar.Mul
 inline_for_extraction noextract
 let m_spec (a:hash_alg) : Type0 =
   match a with
-  | MD5 | SHA1 | SHA2_224 | SHA2_256 | SHA2_384 | SHA2_512 | SHA3_256 -> unit
   | Blake2S | Blake2B -> Blake2.m_spec
+  | _ -> unit
```
R1kM, line 23: Did you see proof performance improvements with this change?
I usually like the exhaustive patterns instead of relying on wildcards as it makes it easier to track required changes when adding a new algorithm, but this is a matter of personal taste.

code/streaming/Hacl.Streaming.MD.fst
```diff
@@ -70,12 +70,19 @@ open Hacl.Hash.Definitions
 module D = Hacl.Hash.Definitions
 module Agile = Spec.Agile.Hash
 
+let _: squash (inversion hash_alg) = allow_inversion hash_alg
+
+let is_supported = function
```
R1kM, line 75: No need for noextract on this?

dist/gcc-compatible/Hacl_Hash_Base.h
```diff
@@ -47,6 +47,8 @@ uint32_t Hacl_Hash_Definitions_hash_len(Spec_Hash_Definitions_hash_alg a);
 
 typedef uint8_t *Hacl_Hash_Definitions_hash_t;
 
+typedef Spec_Hash_Definitions_hash_alg Hacl_Hash_Definitions_const_impl;
```
R1kM, line 50: Maybe worth marking this definition as inline_for_extraction noextract in Hacl.Hash.Definitions?

specs/Spec.Hash.Definitions.fst
```diff
@@ -37,15 +37,25 @@ type hash_alg =
   | Blake2S
   | Blake2B
   | SHA3_256
+  | SHA3_224
```
R1kM, line 40: Could you add a comment along the lines of "For backwards-compatibility with existing C clients when only SHA3_256 was available, we put the SHA3_224 constructor after SHA3_256" to ensure that we do not change the order in future refactorings?

specs/Spec.Hash.Definitions.fst
```diff
@@ -154,35 +165,48 @@ let word_t: hash_alg -> Tot inttype = function
...
+  | SHA3_224 | SHA3_256 | SHA3_384 | SHA3_512 | Shake128 | Shake256 -> U64
 
 inline_for_extraction
 let row (a:blake_alg) = lseq (uint_t (word_t a) SEC) 4
 
 inline_for_extraction
 let word (a: hash_alg) = match a with
-  | MD5 | SHA1 | SHA2_224 | SHA2_256 | SHA2_384 | SHA2_512 | SHA3_256 -> uint_t (word_t a) SEC
   | Blake2S | Blake2B -> row a
+  | _ -> uint_t (word_t a) SEC
```
R1kM, line 176: It might be useful to avoid wildcard uses in specs to reduce the risk of mistakes in the future.

R1kM (approved):
Looks good to me, modulo a few very minor comments.
Thanks!

code/hash/Hacl.Hash.Definitions.fst
msprotz: I did not notice any proof improvements. No problem with making this pattern (and others below) exhaustive. Fixed in 2ce8329cd6 which is part of the subsequent PR.

code/streaming/Hacl.Streaming.MD.fst
msprotz: I think the definition is eliminated via bundling (and also happens to be Low*). I added it for good measure. Fixed in 2ce8329cd6 which is part of the subsequent PR.

dist/gcc-compatible/Hacl_Hash_Base.h
msprotz: Fixed in 2ce8329cd6 which is part of the subsequent PR.

specs/Spec.Hash.Definitions.fst
msprotz: Fixed in 2ce8329cd6 which is part of the subsequent PR. Also fixed other wildcards below.

specs/Spec.Hash.Definitions.fst
msprotz: Fixed in 2ce8329cd6 which is part of the subsequent PR.

PR #803, opened 2023-04-14
Keccak streaming API

msprotz:
Hot on the heels of #800, this PR adds a new instance of the streaming functor specifically for algorithms of the Keccak family.

## Background

The streaming functor currently comes in several flavors of agility (the ability to parameterize the code over a choice of algorithm or implementation).
- Static agility: the instance is generic, but must be specialized at compile-time before extraction. Examples: Poly1305 (generic of the choice of implementation: 32, 128 and 256), Blake2 (generic over both algorithm and implementation: Blake2b/32, Blake2b/256, Blake2s/32, Blake2s/128), and MD (generic over the choice of algorithm: MD5, SHA1, SHA2). Concretely, you have: 
```
let hacl_md (a:alg)// : block unit =
  =
  Block
```
and you do:
```
let update_512 = F.update (hacl_md SHA2_512) (G.hide ()) ((state_t SHA2_512).s ()) (G.erased unit)
```
Things are done this way because SHA2_512 and SHA2_256 have different state representations, so you can't extract to Low* a type of the form `lbuffer (if SHA2_256 then uint32 else if SHA2_512 uint64)` -- you have to pick either one at compile time and rely on F* to reduce it to either `lbuffer uint32` or `lbuffer uint64` before sending it over to krml.

- Runtime agility: the state is an inductive, indexed by the agility parameter `type state a = | SHA2_256: lbuffer uint32 ... -> state SHA2_256 | SHA2_512: lbuffer uint64 ... -> state SHA2_512`. This style compiles to C (as a tagged union), but creates more representation overhead and an extra level of indirection because now you have a reference to a `state a` that contains an `lbuffer`. No need to do compile-time specialization: the user picks at run-time which algorithm they want to run with.

## This PR

This PR introduces a new style that is somewhat in-between the two, relying on the observation that all algorithms from the Keccak family share the same state representation.

For this streaming functor, the state becomes `sha3_alg & lbuffer uint64 25`, i.e. no need to have an inductive with different types under each branch.

Making it work required quite some delicate surgery, notably:
- using a singleton type for the left component of the pair to tie it to the index, in order to justify that operations like `copy` really *do* operate on the same algorithm
- judicious use of erasure, to enforce statically the fact that the instance of the streaming functor never requires the user to provide the algorithm at run-time (since it's already stored internally)
- a few plays on types in Hacl.Hash.SHA3 to explain to F* that a Low*-compatible, monomorphic type that talks about `lbuffer uint64 25` is really a refinement of the generic type expected by callers
- a few touch-ups to make sure quality code is generated for the various functions
- and many more little things.

I'm probably going to push a few cleanups but wanted to get this out to start the work of getting it green and gathering feedback.

This is the second half of the work required for adding SHA3 to python/cpython#99108


code/streaming/Hacl.Streaming.Keccak.fst
```diff
@@ -0,0 +1,242 @@
...
+      ())
+    (* frame_freeable: *) (fun #_ l (_, s) h0 h1 ->
+      ())
+    (* alloca: *) (fun (a: alg) ->
+      a, B.alloca (Lib.IntTypes.u64 0) 25ul)
+    (* create_in: *) (fun a r ->
+      a, B.malloc r (Lib.IntTypes.u64 0) 25ul)
+    (* free: *) (fun _ (_, s) ->
+      B.free s)
+    (* copy: *) (fun _ (a, s_src) (a', s_dst) ->
```
msprotz, line 110: this is where the singleton type is needed: because the two types share the same index (as imposed by the type signature of copy), we can conclude that the two values are exactly the same, meaning that a and a' are identical

code/hash/Hacl.Hash.SHA3.fsti
```diff
@@ -9,22 +9,61 @@ module Hacl.Hash.SHA3
...
+
+module B = LowStar.Buffer
+module ST = FStar.HyperStack.ST
+
+/// This is a variant of Hacl.Hash.SHA3.finish that takes an optional output
+/// length. This is used in the Keccak streaming functor, wherein this signature
+/// eventually becomes exposed in the C API. As a result, we cannot have `l` be
+/// an indexed type (e.g. if is_shake then size_t else unit) because that would
+/// not be extractable to C. So, we contend with a suboptimal contract, which
+/// is: "if a is a shake algorithm, then the length is ignored".
```
R1kM, line 56: Isn't it the opposite, i.e. if a is *not* a shake algorithm, then the length is ignored?
Also, why do you need the length precondition as a refinement instead of in the requires clause?

code/streaming/Hacl.Streaming.Keccak.fst
```diff
@@ -0,0 +1,242 @@
...
+    (ensures (
+      let input = S.append input1 input2 in
+      S.length input % U32.v (D.block_len a) = 0 /\
+      prevlen2 % U32.v (D.block_len a) = 0 /\
+      update_multi_s a (update_multi_s a acc prevlen1 input1) prevlen2 input2 ==
+        update_multi_s a acc prevlen1 input)) =
+  Spec.Hash.Lemmas.update_multi_associative a acc input1 input2
+#pop-options
+
+let singleton #t (x: t) = y:t { y == x }
```
R1kM, line 81: noextract/inline_for_extraction?

R1kM (commented):
The F* changes look reasonable so far. I'll have a look at the generated C code once you're done with the clean-up.

code/hash/Hacl.Hash.SHA3.fsti
msprotz: Good catch thanks

code/streaming/Hacl.Streaming.Keccak.fst
msprotz: thanks

msprotz:
@R1kM I think the C code is where I want it to be. I'm still wrestling with CI.

dist/gcc-compatible/Hacl_Hash_SHA3.c
```diff
@@ -25,73 +25,266 @@
...
+  uint8_t *blocks,
+  uint32_t n_blocks
+)
+{
+  for (uint32_t i = (uint32_t)0U; i < n_blocks; i++)
+  {
+    uint8_t *block = blocks + i * block_len(a);
+    Hacl_Impl_SHA3_absorb_inner(block_len(a), block, s);
+  }
+  uint8_t *last = blocks + n_blocks * block_len(a);
```
R1kM, line 105: Why do we need this last statement?

dist/gcc-compatible/Hacl_Hash_SHA3.c
```diff
@@ -25,73 +25,266 @@
...
+  }
+  KRML_CHECK_SIZE(sizeof (uint8_t), len);
+  uint8_t nextBlock[len];
+  memset(nextBlock, 0U, len * sizeof (uint8_t));
+  nextBlock[len - (uint32_t)1U] = (uint8_t)0x80U;
+  Hacl_Impl_SHA3_loadState(len, nextBlock, s);
+  Hacl_Impl_SHA3_state_permute(s);
+}
+
+typedef struct st2_s
```
R1kM, line 166: Could we have a more informative name for this?

dist/gcc-compatible/Hacl_Hash_SHA3.c
```diff
@@ -25,73 +25,266 @@
...
 uint32_t
-Hacl_Streaming_SHA3_update_256(Hacl_Streaming_MD_state_64 *p, uint8_t *data, uint32_t len)
+Hacl_Streaming_Keccak_update(Hacl_Streaming_Keccak_state *p, uint8_t *data, uint32_t len)
 {
-  Hacl_Streaming_MD_state_64 s = *p;
+  Hacl_Streaming_Keccak_state s = *p;
+  Hacl_Streaming_Keccak_st block_state = s.block_state;
   uint64_t total_len = s.total_len;
-  if ((uint64_t)len > (uint64_t)18446744073709551615U - total_len)
+  Spec_Hash_Definitions_hash_alg i = block_state.fst;
```
R1kM, line 251: Nit: `i` seems to suggest an array offset instead of an algorithm.

dist/gcc-compatible/Hacl_Hash_SHA3.h
```diff
@@ -35,25 +35,38 @@ extern "C" {
 #include "krml/lowstar_endianness.h"
 #include "krml/internal/target.h"
 
-#include "Hacl_Streaming_Types.h"
+#include "Hacl_Spec.h"
```
R1kM, line 38: This is independent from this PR, but Hacl_Spec is not the greatest name for this bundle. Maybe Hacl_Hash_Types?

dist/gcc-compatible/Hacl_Hash_SHA3.c
msprotz: https://github.com/hacl-star/hacl-star/blob/main/lib/Lib.Buffer.fst#L351

Not sure how to get rid of it. Maybe another peephole optimization in krml?

dist/gcc-compatible/Hacl_Streaming_Poly1305_128.c
```diff
@@ -26,7 +26,7 @@
 #include "Hacl_Streaming_Poly1305_128.h"
 
 Hacl_Streaming_Poly1305_128_poly1305_128_state
-*Hacl_Streaming_Poly1305_128_create_in(uint8_t *k)
+*Hacl_Streaming_Poly1305_128_create_in(uint8_t *k1)
```
R1kM, line 29: Why the change into k1?

R1kM (commented):
Thanks! I left a few comments about the shape of the C code.

dist/gcc-compatible/Hacl_Hash_SHA3.c
msprotz: yeah it's `i` for index (the streaming code is generic over an index)... since this is meta-programmed, unclear whether we can pick a meaningful name that won't stop making sense for other cases

dist/gcc-compatible/Hacl_Hash_SHA3.c
```diff
@@ -100,51 +293,49 @@ Hacl_Streaming_SHA3_update_256(Hacl_Streaming_MD_state_64 *p, uint8_t *data, uin
   }
   else if (sz == (uint32_t)0U)
   {
-    Hacl_Streaming_MD_state_64 s1 = *p;
-    uint64_t *block_state1 = s1.block_state;
+    Hacl_Streaming_Keccak_state s1 = *p;
+    Hacl_Streaming_Keccak_st block_state1 = s1.block_state;
     uint8_t *buf = s1.buf;
     uint64_t total_len1 = s1.total_len;
+    Spec_Hash_Definitions_hash_alg i1 = block_state1.fst;
```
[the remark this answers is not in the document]
msprotz: yeah touched up a few functions to avoid this and a couple others

dist/gcc-compatible/Hacl_Hash_SHA3.c
msprotz: done

dist/gcc-compatible/Hacl_Hash_SHA3.h
msprotz: there are potentially more useful types in there beyond hashes, the name (obviously) comes from the fact that this groups together all the Spec.* files... any thoughts?

dist/gcc-compatible/Hacl_Streaming_Poly1305_128.c
msprotz: the conflict avoidance mechanism of krml is sometimes conservative

I think a module somewhere defines a top-level constant named k

msprotz: also fixed

dist/gcc-compatible/Hacl_Hash_SHA3.c
msprotz: the reason unused variable elimination doesn't kick in is because we don't know that block_len is pure, and therefore can be eliminated

dist/gcc-compatible/Hacl_Hash_SHA3.h
msprotz: fixed in per Aymeric and I's discussion [a41202d](https://github.com/hacl-star/hacl-star/pull/803/commits/a41202dc7378b6e41c609d6b2f8f8b8e10581827)

msprotz:
@franziskuskiefer bringing your attention to this as it changes the SHA3 streaming API to a more general one for all Keccak constructions -- you now need to pass your choice of algorithm at initialization-time, and the finish function now has two modes, depending on whether the initial choice of algorithm is shake (variable length output) or not (fixed length output solely dependent on the variant of sha3 picked)$:
- finish(st, dst, 0) when `not (is_shake st)` -- the user MUST pass 0 to indicate that they are aware that this is a non-variable-length algorithm
- finish(st, dst, l) whwen `is_shake st` -- the user MUST pass a non-zero length `l`

The API is defensive and will return a non-zero error code to indicate failure if the user fails to obey the MUST clauses.

The alternative design is to have two functions, one called `finish`, and another one called `finish_shake`, where only the latter takes a length argument. I don't have strong feelings either way, so opinions welcome.

R1kM (approved):

dist/gcc-compatible/Hacl_Hash_SHA3.c
msprotz: Aymeric and I talked about this, and the conclusion is that we'd need repeat_blocks_multi (with a suitable precondition that `length input % block_length == 0`). I believe we don't have this combinator yet, but @polubelova can confirm.

polubelova: that's correct, we don't have `loop_blocks_multi` in `Lib.Buffer.fst`

msprotz: ok no problem let me add it it's not a big deal

franziskuskiefer:
> @franziskuskiefer bringing your attention to this as it changes the SHA3 streaming API to a more general one for all Keccak constructions

Looks like didn't actually respond here 😬  Here my thoughts.

1. I think the implicit switch in the `finish` through the argument is not great.
2. Why do you need an argument in the `finish` function? When you made your choice in the `init`, this shouldn't be necessary, i.e. you can put in the length there too.
3. Using 0/1 arguments are always bad, because the caller doesn't know what they stand for. The better approach would be to define a `SHAKE` and `SHA` (or whatever else), that should be used.
4. Looking at the previous points, adding a separate function might be the easiest though.

msprotz:
I've implemented the requested changes. Can either one of @franziskuskiefer or @karthikbhargavan mark this as approved? I'd like to land this soon, it's been a while

karthikbhargavan (approved):
Thanks, much needed changes.

franziskuskiefer (approved):
Thanks :shipit:

PR #806, opened 2023-04-18
Update usage of --split_queries

mtzguido:
This is a fix needed to merge https://github.com/FStarLang/FStar/pull/2776. I will update this later with an everest green including both changes.

## Proposed changes

An upcoming PR in F* will require an argument for `--split_queries`, either `no`, `on_failure` or `always`. The default is `on_failure` (same as not passing the flag before). The old behavior is obtained with `--split_queries always`. So, this just replaces uses of the option to the `always.`

## Types of changes

## Checklist

- [x] I have edited [CHANGES.md](https://github.com/hacl-star/hacl-star/blob/main/CHANGES.md) (if appropriate)

### DIFF
--- a/code/meta/example/Device.fst
+++ b/code/meta/example/Device.fst
@@ -9,7 +9,7 @@ module ST = FStar.HyperStack.ST
 open FStar.HyperStack.ST
 open LowStar.BufferOps
 
-#set-options "--z3rlimit 100 --fuel 1 --ifuel 1 --__no_positivity --record_options --split_queries"
+#set-options "--z3rlimit 100 --fuel 1 --ifuel 1 --__no_positivity --record_options --split_queries always"
 
 (*** Linked lists *)
 noeq


mtzguido:
Everest green here https://github.com/project-everest/everest/actions/runs/4727379343

R1kM (approved):
Thanks Guido!
Just to be sure, it seems split_queries can already take an option with F* master, so we could merge this as is before merging the F* PR, is that correct? As far as I can see, the verification CI came back green on this change.

mtzguido:
> Just to be sure, it seems split_queries can already take an option with F* master, so we could merge this as is before merging the F* PR, is that correct? As far as I can see, the verification CI came back green on this change.

I was wondering about that... actually F* does *not* support optional arguments (hence this is a breaking change). I'm surprised it's green, is this file being verified? A quick grep shows it's not opened anywhere at least.

In any case, I'm totally fine with merging this before, all the better!

msprotz:
I don't think this file is under CI. But we should have other occurrences of split queries elsewhere, no?

R1kM:
I actually do not see any other occurences by grepping in the repo. I think what's happening is that all the split_queries that we are seeing during verification happen when a query fails, they are not explicitly annotated in the files.

mtzguido:
> I think what's happening is that all the split_queries that we are seeing during verification happen when a query fails

Yes, most likely. That's both the current and new default behavior, but you'll now be able to turn it off with `--split_queries no`.

R1kM:
Sounds good. Especially since this file is not under CI, I'd be happier merging this once the F* PR lands, but this can be done easily afterwards.
Thanks again for the fix!

mtzguido:
Sounds good! Just merged it on F*.

PR #802, opened 2023-04-14
Improve the performance of P-256

polubelova:
This PR is a first round of improvements for P-256:
- introducing a new succinct specification (fixes https://github.com/hacl-star/hacl-star/issues/567)
- using a generic bignum library for finite field arithmetic
- using a generic exponentiation for elliptic curve scalar multiplication
- using complete formulas for point addition and doubling (https://eprint.iacr.org/2015/1060.pdf)
- using addition chains for field and scalar inversion (https://briansmith.org/ecc-inversion-addition-chains-01#p256_field_inversion)
- exposing more safe API (closes https://github.com/hacl-star/hacl-star/pull/251)
- updating the documentation for `Hacl_P256.h`

Here are some performance numbers in CPU cycles (on `11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz`):
```
|              | this branch | this branch | main    | main     |
|    	       | gcc-9       | clang-10    | gcc-9   | clang-10 |
|--------------|---------------------------|--------------------|
| ecdsa-sign   | 413635      | 236492      | 1281928 | 775434   |
| ecdsa-verify | 1021866     | 528342      | 2873185 | 1790256  |
| ecdh         | 848661      | 413552      | 1268367 | 810400   |

gcc-9:
R_sign = 1281928/413635 = 3.09
R_verify = 2873185/1021866 = 2.81

clang-10:
R_sign = 775434/236492 = 3.27
R_verify = 1790256/528342= 3.38
```

PR #808, opened 2023-04-21
secp256k1: add ecdh and rm extra functions

polubelova:
This PR is also one step towards a generic spec & implementation for ECDSA and ECDH. 
The code structure is getting ever more similar to #802.


msprotz (approved):
Whatever brings the APIs closer with the same naming convention is a win. Thanks!

karthikbhargavan (approved):
Is this code specific to uint64_t? What would it take to make it also work efficiently on uint32_t? 

The code appears to use a 52-radix representation? Is this standard? Would 51 work just as well? 

Is there value in having an optimized Vale implementation of some of the field arithmetic?

Other than these questions, the changes look fine to me.

polubelova:
Thanks for the review!

> Is this code specific to uint64_t? What would it take to make it also work efficiently on uint32_t?

Yes, it's specific to uint64_t. I think we first need to add a generic spec & code for ECDSA and ECDH (https://github.com/hacl-star/hacl-star/issues/807), and then probably store precomputed tables in bytes to share them between 32-bit and 64-bit versions.

> The code appears to use a 52-radix representation? Is this standard? Would 51 work just as well?

[The libsecp256k1 library](https://github.com/bitcoin-core/secp256k1) uses a 2^52-radix representation for 64-bit platforms and a 2^26-radix representation for 32-bit platforms. Using a 2^51-radix representation wouldn't be so efficient (it takes one more limb to represent a field element).

> Is there value in having an optimized Vale implementation of some of the field arithmetic?

Yes, we could use an optimized Vale implementation for a 2^64-radix representation similar to the one used for curve25519 (https://github.com/hacl-star/hacl-star/tree/marina_secp256k1_64).

PR #812, opened 2023-05-02
tweaks for bignum extraction

polubelova:
This is just a small change for bignum extraction, namely,  we no longer inline the `bn_get_bits` function.
This PR also cherry-picks the commit from #543 which makes the post-condition of `bn_field_init` more accurate.


msprotz (approved):
Looks good to me, it generates more compact C code and should have no performance hit because the function is static inline. Thanks!

R1kM (approved):

PR #818, opened 2023-05-17
Improve error codes in streaming APIs

msprotz:
Since we are now returning possible error codes for finish and squeeze (invalid algorithm), this PR also aligns `update` to return an error code instead of 0/1 (more descriptive).

This probably calls for a documentation update (@pnmadelaine ?).


R1kM (approved):
Looks ilke a nice improvement, thanks! Could you also include a CHANGES.md entry here?

PR #830, opened 2023-06-11
EverCrypt.Hash: use @@strict_on_arguments for efficiency

mtzguido:
The extraction of EverCrypt.Hash.Incremental has a peak memory usage of around 11GB, and takes around 10 minutes (on my setup). Some cursory inspection of F* debug output showed many normalization calls involving matches to destroy the hash state type, as is done in these functions.

Unfolding these functions into the match is (very) useful if the matched-on argument is a literal constructor, so the match can reduce, but a waste that can lead to an exponential explosion otherwise.

Hence, use the strict_on_arguments attribute to make F* 1) eagerly reduce the arguments to these functions before unfolding, and 2) only unfold if the specified arguments are concrete (i.e. the head is
   a constructor).

Re measuring showed that memory usage falls from 11084 MB to 3596 MB when extracting EverCrypt_Hash_Incremental.krml. Runtime decreases from 691s to 112s (again on my setup, measured in isolation).

A full build passed with this patch, and the `dist` directory ended up exactly the same on both runs. Also the extracted `krml` file is equal byte-for-byte.

There likely are more functions that can be benefit from this attribute.

## Proposed changes

## Types of changes

## Checklist

## Further comments

Hi, we first noticed this file was taking tons of memory here https://github.com/FStarLang/FStar/pull/2734. I finally looked into it.

Here are some comparisons of a full build running before and after this patch, see how the memory usage peaks are lower (especially the second one). You can also see how `obj/EverCrypt_Hash_Incremental.krml.end` appears earlier.

hacl_nf1 ramon
hacl_nf2 ramon

Also the bottom line of extracting this file for both before and after:
```
group.total          692.154s
group.utime          686.489s
group.stime          5.665s
group.mempeak        11084MB
group.pidpeak        5
status               exited
exitcode             0
walltime             691.669s
loadavg              1.00
```
```
group.total          114.156s
group.utime          112.439s
group.stime          1.716s
group.mempeak        3596MB
group.pidpeak        5
status               exited
exitcode             0
walltime             114.140s
loadavg              1.00
```

(These results are from a tool that I am writing on and off, I hope to get it good enough to replace runlim and use it everywhere in our builds.)

Please run your tests though! I only did `make`.

### DIFF
--- a/providers/evercrypt/fst/EverCrypt.Hash.fst
+++ b/providers/evercrypt/fst/EverCrypt.Hash.fst
@@ -131,6 +131,7 @@ let invert_state_s (a: alg): Lemma
 =
   allow_inversion (state_s a)
 
+[@@strict_on_arguments [1]]
 inline_for_extraction
 let impl_of_state #a (s: state_s a): i:impl { alg_of_impl i == a } =
   match s with
@@ -155,6 +156,7 @@ let impl_of_state #a (s: state_s a): i:impl { alg_of_impl i == a } =
 //   ...) so as not to repeat redundant information at run-time
 // - hope that we can get away with returning dependent pairs only when needed.
 // We're going for a third one in this module, which is more lightweight.
+[@@strict_on_arguments [1]]
 inline_for_extraction
 let p #a (s: state_s a): Hacl.Hash.Definitions.state (impl_of_state s) =
   match s with
@@ -215,6 +217,7 @@ let frame_invariant #a l s h0 h1 =
   assert (repr_eq (repr s h0) (repr s h1))
 
 inline_for_extraction noextract
+[@@strict_on_arguments [0]]
 let alloca a =
   let s: state_s a =
     match a with
@@ -253,6 +256,7 @@ let alloca a =
   in
   B.alloca s 1ul
 
+[@@strict_on_arguments [0]]
 let create_in a r =
   let h0 = ST.get () in
   let s: state_s a =


msprotz:
Thanks Guido, this is excellent. I, too, have noticed quite the slowdown when extracting that file but I never looked into where it came from. I would love to see a tutorial or wiki page on this magical tool as soon as it comes out :-).

Does the numbering of strict_on_arguments start at 0 or 1? I was expecting these functions to be normalized when the first (i.e. 0-th) argument is a constant.

mtzguido:
> I would love to see a tutorial or wiki page on this magical tool as soon as it comes out :-).

Will do!

> Does the numbering of strict_on_arguments start at 0 or 1? I was expecting these functions to be normalized when the first (i.e. 0-th) argument is a constant.

Starts at zero, and all arguments (implicit or explicit) are counted. So on `impl_of_state` and `p` and  I make the function strict on `s`, and on `alloca` and `create_in` strict on `a`.

msprotz (approved):
Ok so as a general guideline if something is marked inline_for_extraction, and is (essentially) a `match` on the i-th argument, then it's a good idea to make it also strict on that argument.

I wonder if there's an easy way to scan for these in the HACL* codebase.

In any case, it's a huge win. Thanks!

karthikbhargavan (approved):
Thanks!

mtzguido:
> I wonder if there's an easy way to scan for these in the HACL* codebase.

Just tried a [stupid patch](https://gist.github.com/mtzguido/ed43662a1afc9321e6429560bd297ca5) to print all `inline_for_extraction` definitions that contain a `match` somewhere and ran dependency analysis in HACL*. It spat out this list of 227 definitions https://gist.github.com/mtzguido/cdcf757f7696fe4b5032dd8a11413f34.

msprotz:
That's super helpful. @R1kM what do you think of annotating some of these especially in Lib to see if it helps with extraction performance? these things have been known to cause massive slowdowns in the past

R1kM:
That sounds like a great idea. Do we already have a reliable tool in place (e.g., under CI) to track variation in extraction time and memory usage though?

mtzguido:
> That sounds like a great idea. Do we already have a reliable tool in place (e.g., under CI) to track variation in extraction time and memory usage though?

There's the runlim scripts, so running with `RESOURCEMONITOR=1` should generate the files, and then you can summarize with `res_summary.sh` or compare two directories with `runlim_diff.py`, both in `FStar/.scripts`.

I could also run my tool, but it doesn't do this kind of thing yet :). I might have that working in a few days, or I could also just measure after the fact.

PR #836, opened 2023-06-21
Makefile: a nit, put file at the end of the F* invocation

mtzguido:
Hi, this is totally unimportant, but it may be nice to keep the file that F* is called on as the last argument. This allows to just prepend `emacs -f fstar-debug-invocation` to the relevant `.cmd` to get an interactive session on that file.

A hack like this also works to just reuse the Makefile... but it's probably too ugly to live.
```diff
 %.checked: FSTAR_FLAGS=
 %.checked: | hints
        $(call run-with-log,\
+         $(if $(EMACS), emacs -f fstar-debug-invocation,) \
          $(FSTAR) $(FSTAR_FLAGS) \
            --hint_file hints/$(notdir $*).hints \
            $< \
```

----

This makes it slightly more convenient to launch F* on a file with the `emacs -f fstar-debug-invocation` functionality: just prepend that the command line and that is it.

## Proposed changes

## Types of changes

## Checklist


msprotz (commented):
Totally worth fixing, thanks.

Perhaps one way to achieve what you're trying to do is simply to use the `.cmd` file which contains the F* invocation, logged in the obj directory. I'm not sure I understand precisely what emacs -f does, but would this work?

```
emacs -f ... $(cat obj/My.File.fst.checked.cmd)
```
?

msprotz (approved):

mtzguido:
> Totally worth fixing, thanks.
> 
> Perhaps one way to achieve what you're trying to do is simply to use the `.cmd` file which contains the F* invocation, logged in the obj directory. I'm not sure I understand precisely what emacs -f does, but would this work?
> 
> ```
> emacs -f ... $(cat obj/My.File.fst.checked.cmd)
> ```

I didn't think of that :-) but... turns out it doesn't work: the file actually contains two commands joined by a `&&` (the actual F* call, and the `touch -c`), and apparently bash will take all of them as arguments to emacs.

But copying the command by hand and prepending the `emacs -f fstar-debug-invocation` is not too bad, so not doing anything special is fine.

(This `fstar-debug-invocation` thing is an emacs function defined in fstar-mode.el. It allows to take an fstar invocation of shape `/some/path/fstar.exe --opt1 ..... --optn File.fst` and open an interactive mode with that binary, those options, and that file; but it requires the order to be just that.)

R1kM (approved):

PR #844, opened 2023-08-18
Improve the treatment of Lib.Memzero following the latest changes in krml

msprotz:
See https://github.com/FStarLang/karamel/pull/364 for rationale


msprotz:
does anyone know why the build has not started? @pnmadelaine

R1kM (approved):

karthikbhargavan (approved):

franziskuskiefer:
@msprotz this breaks hacl-packages because the header has a different name than the corresponding C file.
We have to rename one of them.

msprotz:
so Lib_Memzero0.h is gone, replaced by lib_memzero0.h and Lib_Memzero0.c remains (notice the case).

What would you like to see?

msprotz:
yeah the casing encodes whether it's a hand-written file or an auto-generated one

no issues with renaming libmemzero0.h to lib_memzero0.h, would that help?

franziskuskiefer:
> no issues with renaming libmemzero0.h to lib_memzero0.h, would that help?

I can try to update the build script to see if that works. But it's not ideal ...

PR #845, opened 2023-08-28
Miscellaneous proof taming

mtzguido:
## Proposed changes

Hi. I'm working on switching the default Z3 version to the latest master and making sure our projects still verify. This is part of the tweaks I had to do to HACL* to get it to work with the newest Z3.

I have some more locally, but some of them are slightly suspect and I would like to confirm they are not Z3 regressions. The ones here however seems like all around improvements, and still verify with Z3 4.8.5.

## Types of changes

## Checklist

### DIFF
--- a/code/hpke/Hacl.Meta.HPKE.fst
+++ b/code/hpke/Hacl.Meta.HPKE.fst
@@ -2,6 +2,8 @@ module Hacl.Meta.HPKE
 
 friend Hacl.Impl.HPKE
 
+#push-options "--z3rlimit 20"
+
 %splice[
   hpke_setupBaseS_higher;
   hpke_setupBaseR_higher;
--- a/code/k256/Hacl.Impl.K256.Verify.fst
+++ b/code/k256/Hacl.Impl.K256.Verify.fst
@@ -94,12 +94,14 @@ let ecdsa_verify_avoid_finv p r =
   QA.store_qelem r_bytes r;
   load_felem r_fe r_bytes;
   let h1 = ST.get () in
+  assert (modifies (loc r_fe) h0 h1);
   //assert (inv_fully_reduced h1 r_fe);
   //assert (as_nat h1 r_fe == qas_nat h1 r);
 
   let h2 = ST.get () in
   fnormalize tmp_x x;
   let h3 = ST.get () in
+  assert (modifies (loc tmp_x) h2 h3);
   BL.normalize5_lemma (1,1,1,1,2) (as_felem5 h2 x);
   //assert (inv_fully_reduced h3 tmp_x);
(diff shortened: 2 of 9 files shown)


msprotz (commented):
Thanks Guido. Are these regressions caused by changes in upstream F*? I'm wondering what caused these proofs to regress.

Do we have any understanding of what's happening? I know these things are complicated, but it would be good if we could get some advice from the F* team as to how proofs should be authored in a way that makes them more robust and future-proof.

mtzguido:
Hi Jonathan. I think it's just that the new Z3's heuristics are slightly different, and then proofs that are "close to edge" can just fail, and need some tweaking. (The F* I'm using for the SMT upgrade is exactly the current master, just flipping the default Z3 version.) Most of this PR is just rlimit tweaks and some sanity assertions.

Some good practices I try to follow when stabilizing are:
- Never use `#set-options`/`#reset-options`, use `#push-options`/`#pop-options` instead. Unless it's something that must apply to the whole file, if so `#set-options` is OK. This is just to make it easier to follow the current rlimit/fuel/smtopt settings.
- Keep `--hint_info` on to see what's happening with the queries.
- If a definition is generating hundreds of queries, the auto split may have kicked in to report a better error location. I find it's usually better to keep `--split_queries no` or `--split_queries always`. (Some related info [here](https://github.com/FStarLang/FStar/pull/2776#issuecomment-1502397472))
- Use `--quake 1/10/k` while proof taming to get an estimate of stability, it will try every query 10 times and report the success rate, accepting it if at least one succeeded.
- Using that and the "sliding admit style", it's usually easy to find the hard assertions in a proof. For those, attempt `assert_spinoff` to try to prove them in a separate query, and `assert_norm`.
- If one particular assertion requires more rlimit/fuels, set them for that assertion only using a tactic, like this (from this PR):
```fstar
        assert (state_eq_S true s1_orig s2_orig /\ state_eq_S true s1 s2)
            by (let open FStar.Tactics in
                set_rlimit 100;
                ());
```
The "global" rlimit should ideally remain small.
- If all else fails, using `--retry 5` can help advance.

Not really deep insights... but it's most of the procedure I use to make stuff more robust.

I'm thinking of adding a quake variant that will create a logfile with some robustness info, which we can then pick up and summarize for each project (similar to the resource monitoring stuff). Maybe we can have a nightly/weekly job that tries everything 20 times and reports the least stable proofs.

mtzguido:
> I'm a bit worried about the additions of `--retry 5`, it seems like unstable proofs would fly under the radar more easily.

Yes, totally reasonable. If you prefer, we can drop that commit from this PR and wait for the next set of changes for the new Z3.

mtzguido:
To be superseded soon :-)

PR #849, opened 2023-09-07
Karthik p256 small edits

karthikbhargavan:
## Proposed changes

This patch adds a few calles to `LowStar.Ignore.ignore` in cases where KaRaMeL leaves unused variables in the C code. This is needed to pass strict C compilation with  `-Werror -Wall`

## Types of changes

- [ X] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [X] I have read the [CODE_OF_CONDUCT.md](https://github.com/hacl-star/hacl-star/blob/main/CODE_OF_CONDUCT.md) doc
- [ ] I have read and agree to submit my changes under the [LICENSE](https://github.com/hacl-star/hacl-star/blob/main/LICENSE)


msprotz:
can you change dist/Makefile.tmpl to override the defaults from here: https://github.com/hacl-star/hacl-star/blob/main/dist/gcc-compatible/Makefile.basic, concretely, append the following to CFLAGS:

-Wunused-variable -Wunused-but-set-variable -Wunused-function -Wunused-parameter

karthikbhargavan:
This PR only fixes the warnings in the P-256 code. I tried fixing more errors but there are too many to do by hand. Especially in the Bignum code. We need a KaRaMeL patch to do this properly. For now, let's get this patch in so that at least the code that is fixed is in main.

karthikbhargavan:
I have now checked in fixes for more code, including Hash, HMAC, and Bignum.

code/ed25519/Hacl.Impl.SHA512.ModQ.fst
```diff
@@ -62,9 +64,12 @@ let sha512_pre_pre2_msg hash prefix prefix2 len input =
...
   let st = Hash.alloca_512 () in
-  let _ = Hash.update_512 st prefix 32ul in
-  let _ = Hash.update_512 st prefix2 32ul in
-  let _ = Hash.update_512 st input len in
+  let err0 = Hash.update_512 st prefix 32ul in
+  let err1 = Hash.update_512 st prefix2 32ul in
+  let err2 = Hash.update_512 st input len in
+  LowStar.Ignore.ignore err0; // Why do we ignore this error?
+  LowStar.Ignore.ignore err1; // Why do we ignore this error?
+  LowStar.Ignore.ignore err2; // Why do we ignore this error?
```
msprotz, line 72: because the post-condition is crafted in such a way that you can eliminate the error statically -- here, you know you won't exceed the maximum length, so the function will never error out. (In essence, the error condition is 100% characterized by the post-condition so that it can be statically eliminated via a case analysis.)

msprotz (approved):
Good for me, but please fix the comment. Thanks

karthik-cryspen:
Ok, this now passes verification with the fresh hints.

Let me summarize status: we have eliminated a bunch of unused variables and unused functions from the generated code, but not yet all. The annotations added in this PR provide both a guide to what one may need to do for other primitives, and what we may want to ask KaRaMeL to inject upstream. (A quick grep for LowStar.Ignore will yield these instances.)

In a future PR, we should eliminate all such warnings and make `-Wunused-*` as default in `Makefile.tmpl`.

R1kM (approved):

PR #851, opened 2023-09-07
CI: use a self-hosted github runner

pnmadelaine:
Because of the upsurge of problems with Hydra, I propose to switch to a self-hosted github runner.
The new workflow will directly call nix from the same server that hosted the Hydra instance.
Using self-hosted runners on public repositories is usually discouraged because of adversarial pull requests.
Here, I argue that the risk is strongly mitigated thanks to the nix sandbox, as long as we only self-host simple nix commands.
(in fact, the risk stays essentially the same as with the Hydra instance, since an adversarial PR could already build arbitrary nix derivations)

### DIFF
--- a/.github/workflows/hintsanddist.yml
+++ b/.github/workflows/hintsanddist.yml
@@ -7,43 +7,17 @@ on:
 
 jobs:
   regenerate-hints-and-dist:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
     steps:
       - name: checkout
         uses: actions/checkout@v3
 
-      - name: delete old files
-        run: git rm -r hints dist/*/*
-
-      - name: download new files
+      - name: regenerate hints and dist
         run: |
-          STATUS_SUCCEEDED=0 # from https://github.com/NixOS/hydra/blob/cf9f38e43fd81f9298e3f2ff50c8a6ee0acc3af0/hydra-api.yaml#L927-L941
-          baseUrl="https://everest-ci.paris.inria.fr"
-          latestFinishedEval=$(curl -sLH 'Content-Type: application/json' "$baseUrl/jobset/hacl-star/branch-main/latest-eval")
-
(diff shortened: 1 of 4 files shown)


pnmadelaine:
> I don't see any output for the build of HACL* here: https://github.com/hacl-star/hacl-star/actions/runs/6112384249/job/16589736509?pr=851 is this normal?

that's expected when one forgets to pass `-L` to `nix build`... thanks for noticing!

msprotz (approved):
Looks good to me, thanks!

pnmadelaine:
For the record, I think I was completely wrong about the risk created by self-hosted runners. I failed to realize that an adversarial PR can create new workflows or modify existing ones, that are going to be run on the server without the PR being merged.
Thus we are not protected by the Nix sandbox at all.
But, I think we can use these [settings](https://docs.github.com/en/enterprise-cloud@latest/actions/hosting-your-own-runners/managing-self-hosted-runners/managing-access-to-self-hosted-runners-using-groups#changing-which-workflows-can-access-a-runner-group) to only allow specific workflows, and thus make me right.
The process is a bit painful, since we need to update the list anytime there is a change to one of our workflows.
@msprotz can you take a look at these? I don't have access to the organization settings.
The workflows to allow are currently:

- .github/workflows/nixflakeupdate.yml in hacl-nix
- .github/workflows/nix.yml in hacl-star
- .github/workflows/hintsanddist.yml in hacl-star

msprotz:
This is the setting I put in, please confirm it's ok:

```
hacl-star/hacl-nix/.github/workflows/nixflakeupdate.yml@main,
hacl-star/hacl-star/.github/workflows/nix.yml@main,
hacl-star/hacl-star/.github/workflows/hintsanddist.yml@main
```

pnmadelaine:
lgtm, thanks!

PR #852, opened 2023-09-07
ci: fix hints and dist regeneration

pnmadelaine:
Fixing an error I introduced in my last PR: the build fails if hints are deleted beforehand.
I also took the opportunity to compress build products at the time of generation, and upload them as github artifacts.
Finally I enabled the Nix CI on all branches, not just main, except branches starting with `_` to match the behaviour of the old CI.


msprotz (approved):
It would be good to add a comment here to explain what the logic is. For instance, I'm thinking "what are the hints being removed from the build". Thanks.

msprotz:
@R1kM ping?

R1kM (approved):

PR #860, opened 2023-09-25
ci: update and fix `hintsanddist.yml`

pnmadelaine:
The new regeneration workflow is broken: it commits the flake lock file and a link to the store.
I did not realize that https://github.com/peter-evans/create-pull-request is designed to commit any uncommitted changes before creating a PR.
The old regeneration workflow did not write anything on disk beside the intended changes, so I never noticed.
This should fix the issue by not writing the lock file and cleaning the repository before running `create-pull-request`.

### DIFF
--- a/.github/workflows/hintsanddist.yml
+++ b/.github/workflows/hintsanddist.yml
@@ -10,12 +10,12 @@ jobs:
     runs-on: self-hosted
     steps:
       - name: checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: regenerate hints and dist
         run: |
           # build the new artifacts (in ./result)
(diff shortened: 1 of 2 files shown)


msprotz:
Other reviewers: @pnmadelaine informed me on slack that this still needs a small tweak before we can merge it (other than that, looks good).

Question: won't the call to `git restore` also revert the hints that we intend to be part of the commit?

pnmadelaine:
@msprotz don't you confuse it with `git reset`? `git restore .` is going to restore modified files to the state of `HEAD`, which is the commit we create in the previous step.

msprotz (approved):

msprotz:
@R1kM ? review?

R1kM (approved):
lgtm, thanks!

PR #861, opened 2023-09-29
Make code compile with all warnings enabled

msprotz:
I was moved by Karthik's heroic efforts to systematically get rid of all unused variables, and thought I'd try to revive my previous attempts to make it work *for realz* this time.

This is with the corresponding branch on krml

Note that i) the code compiles with all warnings removed from Makefile.basic and ii) krml now inserts IGNOREs for unused function *parameters* too

Thoughts, comments, etc. @karthikbhargavan @R1kM @polubelova @franziskuskiefer

### DIFF
--- a/dist/gcc-compatible/EverCrypt_AEAD.c
+++ b/dist/gcc-compatible/EverCrypt_AEAD.c
@@ -46,6 +46,8 @@ The state may be reused as many times as desired.
 */
 bool EverCrypt_AEAD_uu___is_Ek(Spec_Agile_AEAD_alg a, EverCrypt_AEAD_state_s projectee)
 {
+  KRML_HOST_IGNORE(a);
+  KRML_HOST_IGNORE(projectee);
   return true;
 }
 
@@ -108,8 +110,8 @@ create_in_aes128_gcm(EverCrypt_AEAD_state_s **dst, uint8_t *k)
     uint8_t *ek = (uint8_t *)KRML_HOST_CALLOC((uint32_t)480U, sizeof (uint8_t));
     uint8_t *keys_b = ek;
     uint8_t *hkeys_b = ek + (uint32_t)176U;
-    uint64_t scrut = aes128_key_expansion(k, keys_b);
-    uint64_t scrut0 = aes128_keyhash_init(keys_b, hkeys_b);
+    KRML_HOST_IGNORE(aes128_key_expansion(k, keys_b));
+    KRML_HOST_IGNORE(aes128_keyhash_init(keys_b, hkeys_b));
     EverCrypt_AEAD_state_s
     *p = (EverCrypt_AEAD_state_s *)KRML_HOST_MALLOC(sizeof (EverCrypt_AEAD_state_s));
     p[0U] = ((EverCrypt_AEAD_state_s){ .impl = Spec_Cipher_Expansion_Vale_AES128, .ek = ek });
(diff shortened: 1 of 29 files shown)


msprotz:
Oh and there's a little bit of noise in this PR because hints & dist haven't run in a while.

All changes of the form

```
KRML_HOST_IGNORE((void *)(uint8_t)0U);
```

are covered by another fix in krml that has been merged already.

franziskuskiefer:
Thanks, looks like a lot of warnings are gone, that's great!

There are a bunch more though that would be good to squash.

There are still non-portable include paths
```
warning: non-portable path to file '"Lib_Memzero0.h"'; specified path differs in case from file name on disk [-Wnonportable-include-path]
warning: #include "lib_memzero0.h"
```

There are a ton of implicit, lossy conversions (using `-Wconversion`), e.g.
```
krml/internal/types.h:99:
warning: implicit conversion loses integer precision: 'uint128_t' (aka 'unsigned __int128') to 'uint64_t' (aka 'unsigned long long') [-Wimplicit-int-conversion]
```

Compiling `EverCrypt_AEAD.c` on a machine without AES-NI and all the required features gives unused parameter warnings on all the AES functions, e.g.
```
warning: unused parameter 'k' [-Wunused-parameter] create_in_aes256_gcm(EverCrypt_AEAD_state_s **dst, uint8_t *k)
```

karthik-cryspen:
I have never been happier to see my painful work being obsoleted. Looks good overall, but yes, let's do a little bit more and squash all the warnings we can. It is much easier to do this with tooling.

msprotz:
> There are still non-portable include paths

Is that a leftover old file that hasn't been removed? I thought Lib_Memzero0.h was gone for good after I renamed it for you into lib_memzero0.h ... is that in dist/gcc-compatible?

msprotz:
> There are a ton of implicit, lossy conversions (using -Wconversion), e.g.

The one you posted is in a hand-written file, independent from this PR: https://github.com/FStarLang/karamel/blob/master/krmllib/c/fstar_uint128_gcc64.h -- if you know how to fix this, you can submit PRs over there directly and the next refresh of the code will pick them up.

msprotz:
> Compiling EverCrypt_AEAD.c on a machine without AES-NI and all the required features gives unused parameter warnings on all the AES functions, e.g.

Ah, yes, that's a tricky one because the use is under an #ifdef -- that one might require some refactoring, or a manual call to LowStar.Ignore in the meanwhile

franziskuskiefer:
> Is that a leftover old file that hasn't been removed? I thought Lib_Memzero0.h was gone for good after I renamed it for you into lib_memzero0.h ... is that in dist/gcc-compatible?

It's in a bunch of files, e.g.
https://github.com/hacl-star/hacl-star/pull/861/files#diff-a0d77c72d84485650f708e3c8920bee10bfc074dec13a9367c00ad179b9eaa99L29
I see it in at least `Hacl_Hash_Blake2s_128.c`, `Hacl_Hash_Blake2.c`, and `Hacl_Ed25519.c` as well.

> The one you posted is in a hand-written file, independent from this PR: https://github.com/FStarLang/karamel/blob/master/krmllib/c/fstar_uint128_gcc64.h -- if you know how to fix this, you can submit PRs over there directly and the next refresh of the code will pick them up.

There are these two. I'm not sure if there's more https://github.com/FStarLang/karamel/pull/382

> Ah, yes, that's a tricky one because the use is under an #ifdef -- that one might require some refactoring, or a manual call to LowStar.Ignore in the meanwhile

Yeah that's a little harder. Another option would be to have the `ifdef` on the outside and a second function that ignores all the arguments.

msprotz:
Regarding your warning:

```
warning: non-portable path to file '"Lib_Memzero0.h"'; specified path differs in case from file name on disk [-Wnonportable-include-path]
warning: #include "lib_memzero0.h"
```

where is `Lib_Memzero0.h` coming from? I don't see it in HACL*

```
jonathan@absinthe:~/Code/everest/hacl-star (main) $ find . -name 'Lib_Memzero0.h'
```

returns nothing. The warning says `#include "lib_memzero0.h"` finds `Lib_Memzero0.h` on an OSX machine, but we don't have `Lib_Memzero0.h` anywhere in HACL*.

franziskuskiefer:
> where is `Lib_Memzero0.h` coming from? I don't see it in HACL*

Ah, I guess these files don't get updated properly on Mac because they are considered identical. Going in manually and deleting everything seems fine. So this shouldn't be an issue.

msprotz:
#863 has a more recent version of the diff, closing this one

PR #789, opened 2023-03-10
API update

pnmadelaine:
This PR updates the API of various functions in HACL*.
See #675 and https://github.com/cryspen/hacl-packages/issues/255.
To merge with https://github.com/cryspen/hacl-packages/pull/377, https://github.com/hacl-star/merkle-tree/pull/13 and https://github.com/project-everest/mitls-fstar/pull/259.

## Blake2b

| Current                                                            | New                                                                                         |
|--------------------------------------------------------------------|---------------------------------------------------------------------------------------------|
| Hacl_Blake2b_32_blake2b(nn, output, ll, d, kk, k)                  | Hacl_Hash_Blake2b_hash_with_key(output, output_len, input, input_len, key, key_len)         |
| Hacl_Streaming_Blake2_blake2b_32_state                             | Hacl_Hash_Blake2b_state_t                                                                   |
| Hacl_Streaming_Blake2_blake2b_32_no_key_create_in(void)            | Hacl_Hash_Blake2b_malloc(void)                                                              |
| Hacl_Streaming_Blake2_blake2b_32_no_key_update(p, data, len)       | Hacl_Hash_Blake2b_update(state, chunk, chunk_len)                                           |
| Hacl_Streaming_Blake2_blake2b_32_no_key_finish(p, dst)             | Hacl_Hash_Blake2b_digest(state, output)                                                     |
| Hacl_Streaming_Blake2_blake2b_32_no_key_init(s1)                   | Hacl_Hash_Blake2b_reset(state)                                                              |
| Hacl_Streaming_Blake2_blake2b_32_no_key_free(s1)                   | Hacl_Hash_Blake2b_free(state)                                                               |
| Hacl_Blake2b_256_blake2b(nn, output, ll, d, kk, k)                 | Hacl_Hash_Blake2b_Simd256_hash_with_key(output, output_len, input, input_len, key, key_len) |
| Hacl_Streaming_Blake2b_256_blake2b_256_state                       | Hacl_Hash_Blake2b_Simd256_state_t                                                           |
| Hacl_Streaming_Blake2b_256_blake2b_256_no_key_create_in(void)      | Hacl_Hash_Blake2b_Simd256_malloc(void)                                                      |
| Hacl_Streaming_Blake2b_256_blake2b_256_no_key_update(p, data, len) | Hacl_Hash_Blake2b_Simd256_update(state, chunk, chunk_len)                                   |
[...]

PR #864, opened 2023-10-05
Add support for parameters beyond key/output length in Blake2

R1kM:
## Proposed changes

This PR adds support for parameters beyond key and hash length in Blake2, as requested by https://github.com/python/cpython/issues/99108 .
The specification of the parameter block can be found in https://www.blake2.net/blake2.pdf, Section 2.8.
One minor difference between the official specification and the F* specification is that, to xor the parameter block with the IV, the parameter block array needs to contain machine integers of the same type (i.e., uint32 for Blake2S, uint64 for Blake2B) instead of bytes. The serialization is done manually in the spec, using shifts and xors.

This PR also adapts the Low* code to support generic parameters in the future, however, at the moment, it does not expose an API to call Blake2 with non default parameters (hence the draft status). It also does not propagate these improvements to the streaming functor, this will be done in a further PR. In the current state, feedback on the specification and generated code would be most useful.

Note: There seems to be regressions on Hacl.Spec.BignumQ.Lemmas again.. It shouldn't have changed compared to main, and the file verifies from Emacs, but fails from the CLI and on CI.


code/blake2/Hacl.Impl.Blake2.Generic.fst
```diff
@@ -549,7 +549,221 @@ let blake2_init_st  (al:Spec.alg) (ms:m_spec) =
...
+      let s5 = salt_u64.[1] in
+      let personal_u64: lseq uint64 2 = uints_from_bytes_le (as_seq h0 (get_personal p)) in
+      let s6 = personal_u64.[0] in
+      let s7 = personal_u64.[1] in
+      [@inline_let]
+      let l = [s0; s1; s2; s3; s4; s5; s6; s7] in
+      assert_norm (List.Tot.length l == 8);
+
+      // There seems to be something not triggering with createL, requiring the
+      // following lemma calls, and assert_norms to relate List.index to the
```
msprotz, line 740: there's supposed to be helpers like seq_of_list that do this automatically

code/blake2/Hacl.Impl.Blake2.Core.fst
```diff
@@ -31,6 +31,34 @@ let row_v #a #m h r =
 
 let row_v_lemma #a #m h0 h1 r1 r2 = ()
 
+let create_default_params a salt personal =
+  match a with
+  | Spec.Blake2S ->
+    Mkblake2s_params
```
msprotz, line 37: so I this done to avoid issues with field overloading, maybe there's a better issue... @mtzguido I thought you added something to help disambiguate field names when they're overloaded across two different types in scope?

msprotz (approved):

R1kM:
After discussion, exposing a Blake2 API with support for parameters without touching to the streaming functor and the rest of the scaffolding seems tricky, and to lead to lots of code duplication (e.g., to have a type of blake2_init that does not take a parameter, and a type for blake parameterized by this new init type).

Hence, all of this will be handled generically in a future PR, that will propagate parameters to hashes and streaming, while also exposing at the top-level APIs that do not take parameters, and that are instantiated with default_params.

R1kM:
Note: This is to be merged after #863 , which also refreshes dist/ (which will reduce noise here), and requires some fixes to the JS bindings.

code/blake2/Hacl.Impl.Blake2.Core.fst
R1kM: Indeed, if I add an extra type ascription, I can use the standard record syntax. I pushed an update.

msprotz:
Do we have any test vectors that exercise the new API?

code/blake2/Hacl.Impl.Blake2.Core.fst
mtzguido: Sorry for the delay! I was away and just seeing this now. I didn't write that feature but yes, types can be used to disambiguate overloaded record fields.

dist/gcc-compatible/Hacl_Hash_Blake2.c
```diff
@@ -500,16 +501,72 @@ void Hacl_Blake2b_32_blake2b_init(uint64_t *hash, uint32_t kk, uint32_t nn)
   r3[1U] = iv5;
   r3[2U] = iv6;
   r3[3U] = iv7;
-  uint64_t kk_shift_8 = (uint64_t)kk << (uint32_t)8U;
-  uint64_t iv0_ = iv0 ^ ((uint64_t)0x01010000U ^ (kk_shift_8 ^ (uint64_t)nn));
+  uint8_t salt[16U] = { 0U };
+  uint8_t personal[16U] = { 0U };
+  Hacl_Impl_Blake2_Core_blake2b_params
```
msprotz, line 506: can this be inlined for extraction?

msprotz: maybe [@inline_let]

code/blake2/Hacl.Impl.Blake2.Generic.fst
R1kM: I looked into this, but, as far as I understand (I might have missed a different lemma), the lemmas should be triggered automatically when using createL. See the comment below, I'm not sure why this is not the case here.

dist/gcc-compatible/Hacl_Hash_Blake2.c
R1kM: This one is a bit unfortunate, but all the functions/stack allocations are in the stack effect, including packing all elements in the params struct (to ensure that the struct satisfies some invariants after initialization, i.e., the salt and personal arrays are live). Hence, inline_let does not work here.

R1kM:
Thanks for the review. We do not have yet any test vectors for new features of the spec (i.e., specific params). The official RFC does not actually specify the parameters block, and instead hardcodes a serialization for the default parameters: https://datatracker.ietf.org/doc/html/rfc7693#section-2.5
Test vectors in the reference source code (https://github.com/BLAKE2/BLAKE2/blob/master/testvectors/blake2-kat.json) also only focus on default parameters.

The new spec in this PR passes the tests for default params that we currently have, which seems to indicate it is at least correct for existing use cases. While not ideal, I'll try to generate some test cases from the Python implementation for non-default params, to try to get some coverage.

karthikbhargavan (approved):
This looks pretty good, thanks!

R1kM:
@polubelova suggested today that we split the Blake2 specification into Spec.Blake2.Parameters, containing all definitions related to parameters, and Spec.Blake2, similarly to the separation we have between Spec.Hash.Definitions and Spec.Agile.Hash. I will update the PR accordingly.

R1kM:
> @polubelova suggested today that we split the Blake2 specification into Spec.Blake2.Parameters, containing all definitions related to parameters, and Spec.Blake2, similarly to the separation we have between Spec.Hash.Definitions and Spec.Agile.Hash. I will update the PR accordingly.

Done in [047b1ed](https://github.com/hacl-star/hacl-star/pull/864/commits/047b1ed28c2003c0745dc4e3d4edad768d9b0910).

msprotz:
Ok anything preventing this from being merged, then? What are the CI failures?

R1kM:
CI failures are related to stabilized proofs in #879 .

We were hoping to merge #789 first to avoid conflicts, but maybe this is not worth it. Let me add the spec tests for extra parameters, and this PR will be ready for merge.

R1kM:
The last two commits added spec tests for Blake2 parameters. As mentioned in the code, the handling of the parameters is not specified in the official RFC, but only in the Blake2 paper. Hence, I could not find official test vectors for parameters; the test vectors in this PR were manually crafted, and the expected results were obtained using the current implementation in the Python 3 hashlib library

PR #865, opened 2023-10-05
Updated codegen from krml branch protz_ignore_ifdef

msprotz:
As a point of discussion, to confirm that the warnings are gone from AEAD.


msprotz:
The JS bindings should also be good, now.

msprotz:
@franziskuskiefer I could use your help I don't know why the PR on the packages side is not picking up the changes in the krml files...

msprotz:
Overall, this disables all of the unused warnings. There are a couple things to note -- thoughts, comments, suggestions welcome.
- This is a little bit verbose.
- It also conflates two things: one is, discarding the return value of a stateful call (e.g. `ignore(expand_key...)`) and generating a spurious use of a variable in order to defeat the unused variable warning (e.g. `(void)x`).

One possible way to have more idiomatic code-gen would be to never insert KRML_HOST_IGNORE, but instead:
- just sequence value-returning calls normally, i.e. `expand_key(...); ...` and never name the return value -- this is standard C practice and C compilers don't budge if you do that, unless the function itself has a nodiscard attribute
- for variables that may be unused, generate `KMRL_MAYBE_UNUSED` like we do for static functions that may be unused (see EverCrypt_Poly1305 for an example).

This is of course a little bit of work, but maybe not the end of the world -- I'm not sure.

franziskuskiefer:
> @franziskuskiefer I could use your help I don't know why the PR on the packages side is not picking up the changes in the krml files...

Good question ... running the update locally again I'm seeing the diff. Looks like the update on CI isn't fully working? @pnmadelaine can you have a look what's going on here?

```
--- a/karamel/include/krml/internal/target.h
+++ b/karamel/include/krml/internal/target.h
@@ -57,6 +57,14 @@
 #  define KRML_HOST_IGNORE(x) (void)(x)
 #endif
 
+#ifndef KRML_MAYBE_UNUSED
+#  if defined(__GNUC__)
+#    define KRML_MAYBE_UNUSED __attribute__((unused))
+#  else
+#    define KRML_MAYBE_UNUSED
+#  endif
+#endif
+
 #ifndef KRML_NOINLINE
 #  if defined(_MSC_VER)
 #    define KRML_NOINLINE __declspec(noinline)
```

R1kM (approved):

R1kM:
Regarding wrapping function calls in KRML_HOST_IGNORE, I'm personally the proposed (C idiomatic) alternative of silently dropping the return value by not naming it. However, I see the point that wrapping a function call in something called "IGNORE" might be confusing.

What would you think of KRML_IGNORE_RETVAL for this specific usecase?

msprotz:
Closing this, since changes have been merged upstream in krml.

PR #869, opened 2023-10-10
Include `alloca.h` in `Hacl_Bignum_Base.h` for mozilla dist

### LINKED ISSUE #868: Missing alloca.h include for dist/mozilla
This is a followup to #857.
Compilation fails on some systems when the `alloca.h` include is missing.
We can either add it during extraction, or modify `builtin.h` to include it. I'd vote to add it to the extraction to get a quick fix and then consider adding it to the `builtin.h`.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1857190 for the original bug report.

I don't know how to reproduce but it's on Solaris

franziskuskiefer:
I didn't test it, but something like this should fix the issue.

Fixes #868

### DIFF
--- a/dist/package-mozilla.sh
+++ b/dist/package-mozilla.sh
@@ -99,6 +99,7 @@ $sed -i -z 's/\n\n\/\*\n  As per[^/]*\*\///g' mozilla/Hacl_P256.c
 
 # Add an include for "builtin.h" to Hacl_Bignum_Base.h
 sed -i -z 's!\(#include.*types.h"\)!\1\n#include "krml/internal/builtin.h"!g' mozilla/internal/Hacl_Bignum_Base.h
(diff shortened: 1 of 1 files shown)


msprotz:
What is the issue? On which build platforms? 

We have this: https://github.com/FStarLang/karamel/blob/master/include/krml/internal/builtin.h which is perhaps what needs to be replicated?

franziskuskiefer:
> What is the issue? On which build platforms?

See the linked issue #868.

msprotz:
let me try to upstream the fix it should take just as long as this PR -- see fstarlang/karamel#386

msprotz:
(The reason I want to do it this way is that alloca.h is not available on windows, so this would likely break your windows build.

franziskuskiefer:
👍🏻 sure, fixing it properly right away is better.

msprotz:
Fixed via #870

PR #876, opened 2023-11-02
blake2: fix a typo in a specification

polubelova:
This MR fixes a typo in the Blake2b specification.

When hashing with a key, the maximum message length should still be less than 2^128 but without one block (for a key), i.e., 2^128 - 128.

Let's wait for the CI to see if this change breaks things or not.


msprotz (approved):

R1kM:
Good catch, thanks Marina!

R1kM (approved):

PR #879, opened 2023-11-09
Vale.Def.Words.Seq: stabilize a tricky proof

mtzguido:
## Proposed changes

This is just stabilizing a proof, `nat_to_four_to_nat`, that I've seen fail a couple of times already. It's about the decomposition of a U32 into chunks four of U8 being unique. The proof is pretty hardcoded to both the 4 components and the 'size' of 256, which I think makes it more complex than needed. Instead I wrote a more generic lemma about any base representation being unique (modulo left zeroes).

I'm not sure this is the right place for this lemma, or whether something like this exists... but it does make the module a lot more robust. Perhaps a similar idea can be used for the analogous `four_to_nat_to_four_8` lemma, but it seems to work just fine.

## Types of changes

## Checklist

### DIFF
--- a/code/ed25519/Hacl.Spec.BignumQ.Lemmas.fst
+++ b/code/ed25519/Hacl.Spec.BignumQ.Lemmas.fst
@@ -897,7 +897,7 @@ let aux (a b c:int)
   : Lemma (requires 0 <= b /\ c < a)
           (ensures 0 <= a + b -c) = ()
 
-#push-options "--z3rlimit 50 --fuel 0 --ifuel 0"
+#push-options "--z3rlimit 100 --fuel 0 --ifuel 0 --split_queries always"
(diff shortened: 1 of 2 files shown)


mtzguido:
Missed the fact that the Makefile disables NL arith and sets low fuels, fixed now. This passes --quake 10 without hints.

R1kM (commented):
Thanks! This looks good to me, but CI seem to be failing because of an unrelated flaky proof in Hacl.Spec.BignumQ.Lemmas.
Could you merge the branch `afromher_fix` into yours, which hopefully should stabilize it?

mtzguido:
Thanks! Somehow that file didn't fail locally.. merged now.

R1kM (approved):

msprotz (approved):

PR #881, opened 2023-11-14
Hacl.Spec.BignumQ.Lemmas: trying to stabilize proof

mtzguido:
## Proposed changes

This proof is failing for me again after some changes to F* that shouldn't really impact it. It looks like the hint fails to replay (see below), and the solver fails to prove the full query when hints are being used on the file. Running without hints works just fine. We've seen other instances of this elsewhere, and I would think this is caused by some random internal state in the solver that is slightly different in either case.. but I'm not sure.

For now, it seems like restarting the solver does the trick. The hint replay fails, but the full query does work from the "clean" state. Even if hint hashes are disabled, which I suppose could be the trigger of this breaking.
```
Z3 says: unknown
(/home/guido/r/fstar/3089/everest-TEMP/hacl-star/code/ed25519/Hacl.Spec.BignumQ.Lemmas.fst(968,9-968,15))       Query-stats (Hacl.Spec.BignumQ.Lemmas.lemma_barrett_reduce', 54)        failed {reason-unknown=unknown because canceled} (with hint) in 8660 milliseconds with fuel 0 and ifuel 0 and rlimit 250
Z3 says: unsat
(/home/guido/r/fstar/3089/everest-TEMP/hacl-star/code/ed25519/Hacl.Spec.BignumQ.Lemmas.fst(968,9-968,15))       Query-stats (Hacl.Spec.BignumQ.Lemmas.lemma_barrett_reduce', 54)        succeeded in 1537 milliseconds with fuel 0 and ifuel 0 and rlimit 250                                                             
```

## Types of changes

## Checklist

### DIFF
--- a/code/ed25519/Hacl.Spec.BignumQ.Lemmas.fst
+++ b/code/ed25519/Hacl.Spec.BignumQ.Lemmas.fst
@@ -897,6 +897,13 @@ let aux (a b c:int)
   : Lemma (requires 0 <= b /\ c < a)
           (ensures 0 <= a + b -c) = ()
 
+// The proof below fails to replay hints. And, when hints have been
+// used for the file, the query ends up failing, yet it succeeds when
+// hints are disabled. This is puzzling but something that's been
+// observed elsewhere, probably just due to Z3 randomness. In any case,
+// restarting the solver seems to avoid the problem...
+#restart-solver
+
 #push-options "--z3rlimit 100 --fuel 0 --ifuel 0 --split_queries always"
 let lemma_barrett_reduce' x =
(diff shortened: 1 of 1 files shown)


R1kM:
Thanks Guido. This proof has been flaky for a while, we really should try to rewrite it when we have time.
Even if I'm not a big fan of this solution in general, should be just use a retry/quake instead of a restart-solver for this specific lemma to try to be a bit more future-proof? What do others think?

mtzguido:
I think a `--retry` here wouldn't be too bad... it can be set like this for only this subgoal:
```fstar
Tactics.set_options "--retry 5";
```

R1kM:
Subsumed by #882, closing this

PR #882, opened 2023-11-15
Hacl.Spec.BignumQ.Lemmas: Try to stabilize proof

R1kM:
The Barrett reduction proof in Ed25519 is notoriously flaky; while we should ultimately spell it out more to stabilize it, this is a stopgap measure where the faulty queries are annotated with --retry 5 to unblock CI.

Subsumes #881


polubelova (approved):
Thanks!

PR #897, opened 2023-12-14
Vectorized SHA-3 spec and implementation

mamonet:
This patch adds specification and implementation modules for vectorized SHA-3 that support single-lane and 4-lanes.

WIP: The patch is missing equivalence lemmas with the original SHA3 spec.


code/sha3/Hacl.Impl.SHA3.Vec.fst
```diff
@@ -0,0 +1,1503 @@
...
+  let b1 = create 256ul (u8 0) in
+  let b2 = create 256ul (u8 0) in
+  let b3 = create 256ul (u8 0) in
+  let b = ntup4 (b0, (b1, (b2, b3))) in
+  let h0 = ST.get() in
+  Lib.NTuple.eq_intro (as_seq_multi h0 b) (next_block_seq_zero m);
+  //let h1 = ST.get() in
+  //assert (stack_allocated b0 h0' h1 (Seq.create 256 (u8 0)));
+  //assert (stack_allocated b1 h0' h1 (Seq.create 256 (u8 0)));
+  admit(); //Can't prove (stack_allocated_multi b h0 h1 (Seq.create 256 (u8 0)))
```
msprotz, line 762: is this still an issue?

mamonet: I forget about this one, I will update the relative functions shortly.

mamonet: The issue is fixed, admits removed with the last commit.

Makefile.common
```diff
@@ -183,7 +183,9 @@ HASH_BUNDLE= \
 # therefore Hacl.Impl.SHA3 cannot be put on the right-hand side of the bundle
 # (which would eliminate internal helpers not used otherwise, such as absorb or
 # squeeze)
-SHA3_BUNDLE=-bundle Hacl.Streaming.Keccak+Hacl.SHA3+Hacl.Impl.SHA3=Hacl.Hash.SHA3[rename=Hacl_Hash_SHA3,rename-prefix]
+SHA3_BUNDLE=-bundle Hacl.Streaming.Keccak+Hacl.SHA3+Hacl.Impl.SHA3=Hacl.Hash.SHA3[rename=Hacl_Hash_SHA3]
+SHA3_SCALAR_BUNDLE=-bundle Hacl.Hash.SHA3.Scalar=Hacl.Spec.SHA3.Vec.*,Hacl.Impl.SHA3.Vec,Hacl.Hash.SHA3.Scalar
```
msprotz, line 187: when `Hacl.Hash.SHA3.Scalar` is already on the left-hand side of the `=`, you do not need to repeat it on the right-hand side, so please remove `Hacl.Hash.SHA3.Scalar` at the end of the line... same for the line below

code/sha3/Hacl.Hash.SHA3.Scalar.fst
```diff
@@ -0,0 +1,176 @@
...
+  -> inputByteLen:size_t{v inputByteLen == length input}
+  -> Stack unit
+     (requires fun h ->
+       live h input /\ live h output /\ disjoint input output)
+     (ensures  fun h0 _ h1 ->
+       modifies (loc output) h0 h1 /\
+       as_seq h1 (output <: lbuffer uint8 outputByteLen) ==
+       S.shake128 (v inputByteLen) (as_seq h0 (input <: lbuffer uint8 inputByteLen)) (v outputByteLen))
+let shake128 output outputByteLen input inputByteLen =
+  admit();
```
msprotz, line 39: to facilitate the review of this PR, can you either:
- turn this admit into an assume, so that we understand what is missing
- add a comment describing what is missing?

code/sha3/Hacl.Hash.SHA3.Simd256.fst
```diff
@@ -0,0 +1,353 @@
...
+module S = Spec.SHA3
+module V = Hacl.Spec.SHA3.Vec
+module ST = FStar.HyperStack.ST
+module B = LowStar.Buffer
+module M = LowStar.Modifies
+
+#reset-options "--z3rlimit 50 --max_fuel 0 --max_ifuel 0 --using_facts_from '* -FStar.Seq'"
+
+val shake128:
+  output0:buffer_t MUT uint8
```
msprotz, line 29: why not but outputByteLen first then use `lbuffer` type for the output parameters? same with input

code/sha3/Hacl.Hash.SHA3.Simd256.fst
```diff
@@ -0,0 +1,353 @@
...
+       internally_disjoint4 output0 output1 output2 output3)
+     (ensures  fun h0 _ h1 ->
+       modifies (loc output0 |+| loc output1 |+| loc output2 |+| loc output3) h0 h1 /\
+       as_seq h1 output0 == S.sha3_512 (v inputByteLen) (as_seq h0 (input0 <: lbuffer uint8 inputByteLen)) /\
+       as_seq h1 output1 == S.sha3_512 (v inputByteLen) (as_seq h0 (input1 <: lbuffer uint8 inputByteLen)) /\
+       as_seq h1 output2 == S.sha3_512 (v inputByteLen) (as_seq h0 (input2 <: lbuffer uint8 inputByteLen)) /\
+       as_seq h1 output3 == S.sha3_512 (v inputByteLen) (as_seq h0 (input3 <: lbuffer uint8 inputByteLen)))
+let sha3_512 output0 output1 output2 output3
+        input0 input1 input2 input3 inputByteLen =
+  admit();
```
msprotz, line 255: same remark about this admit and the ones above

code/sha3/Hacl.Hash.SHA3.Simd256.fst
```diff
@@ -0,0 +1,353 @@
...
+  let ib = ntup4 (input0,(input1,(input2,input3))) in
+  let rb = ntup4 (output0,(output1,(output2,output3))) in
+  keccak #SHA3_512 #M256 576ul (* 1024ul *) inputByteLen ib (byte 0x06) 64ul rb
+
+val state_malloc:
+    r:rid
+  -> ST.ST (s:buffer uint64 { length s = 100 })
```
msprotz, line 262: use lbuffer?

why is this helper needed?

code/sha3/Hacl.Hash.SHA3.Simd256.fst
```diff
@@ -0,0 +1,353 @@
...
+       as_seq h1 output2 == S.sha3_512 (v inputByteLen) (as_seq h0 (input2 <: lbuffer uint8 inputByteLen)) /\
+       as_seq h1 output3 == S.sha3_512 (v inputByteLen) (as_seq h0 (input3 <: lbuffer uint8 inputByteLen)))
+let sha3_512 output0 output1 output2 output3
+        input0 input1 input2 input3 inputByteLen =
+  admit();
+  let ib = ntup4 (input0,(input1,(input2,input3))) in
+  let rb = ntup4 (output0,(output1,(output2,output3))) in
+  keccak #SHA3_512 #M256 576ul (* 1024ul *) inputByteLen ib (byte 0x06) 64ul rb
+
+val state_malloc:
```
msprotz, line 260: this should probably be marked `inline_for_extraction`

msprotz (commented):
Mostly looks good to me. Thanks!
- Let's try to fix those signatures with casts everywhere. They're pretty unreadable.
- Let's at least document all of the admits so that we understand what remains to be done.

Makefile.common
mamonet: Removed both `Hacl.Hash.SHA3.Scalar` and `Hacl.Hash.SHA3.Simd256` on the right-hand.

code/sha3/Hacl.Hash.SHA3.Scalar.fst
mamonet: The implementation of these functions is not complete yet and it requires equivalent lemmas with original spec to prove post-condition and certain asserts to fulfill "multi-buffer" requirement of `keccak` function, I am working on equivalent lemmas and will update the functions once done.

code/sha3/Hacl.Hash.SHA3.Simd256.fst
mamonet: It's the same parameter order that used in the original implementation with same casting technique, our team asked for this order specifically for what we thought it fits C standard convention of API argument order and former SHA3 API.

code/sha3/Hacl.Hash.SHA3.Simd256.fst
mamonet: Using lbuffer will throw `(Error 331) This name is being ignored` and `(Error 19) Subtyping check failed`. Removing the helper will throw the first error.

code/sha3/Hacl.Hash.SHA3.Simd256.fst
msprotz: yes absolutely

code/sha3/Hacl.Hash.SHA3.Simd256.fst
msprotz: what syntax are you using? I am thinking
```
val state_malloc: r:rid -> ST.ST (lbuffer uint64 100)
```

(or whatever the order of the arguments to lbuffer is)

code/sha3/Hacl.Hash.SHA3.Simd256.fst
msprotz: yes, it's good to remain consistent... here's another style you might want to consider to have a more "HACL-idiomatic" style, yet keep the same order of parameters:

```
let same_as x = y:size_t { x == y }
```

val shake128:
  g_outputByteLen:Ghost.erased size_t ->
  output0: lbuffer uint8 g_outputByteLen ->
  ...
  output3: lbuffer uint8 g_outputByteLen ->
  outputByteLen: same_as g_outputByteLen
  ```
  
  this might remove a bunch of casts -- up to you to decide if you think it's worth it

code/sha3/Hacl.Hash.SHA3.Simd256.fst
mamonet: I tried `val state_malloc: r:rid -> ST.ST (lbuffer uint64 100ul)`, it throws `(Error 331) This name is being ignored`

code/sha3/Hacl.Hash.SHA3.Simd256.fst
mamonet: It's more good-looking style for sure. I will adapt this signature with Scalar/Simd256 functions in the next update.

code/sha3/Hacl.Hash.SHA3.Simd256.fst
msprotz: but which name is being ignored? you should have a precise location for the error

msprotz: also, I'm wondering if adding a type ascription at call-site, such as:
```
let b: lbuffer uint64 100 = malloc ...
```
will help at al

karthikbhargavan:
Spec equivalence is mostly done, it needs some help from HACL* experts to push it over the line.

msprotz (commented):
@mamonet can you give a precise list of the admits you need help with?

there's this one: https://github.com/hacl-star/hacl-star/pull/897/files#diff-cb79561e4bcb46900172e2c56ca32bfd38dc6c5bd5019ccbc2bb2f4e874407a1R676 which you wrote about on Slack, but are the other `admit`s temporary ones that will disappear once the admit above is fixed?

msprotz:
First look at this: Hacl.Impl.SHA3.Vec.fst does not type-check on my machine.

```
Verified module: Hacl.Impl.SHA3.Vec
56 errors were reported (see above)
```

is this expected? @mamonet 

for the record I'm running with 3x the rlimit

msprotz (commented):
This is great and I can't wait to see the old code go! Congrats on an ambitious PR that touches many parts of HACL*. You seem to have become more proficient with Low*/HACL* over the past several months and this is a significant contribution.

Some high-level thoughts.
- Please move the signatures into Hacl.Impl.SHA3.Vec.fsti -- this will make verification faster and more modular.
- Please remove the #a parameter from `absorb_inner` (and fix all of the other functions accordingly). The behavior of `absorb_inner` does not depend on your choice of algorithm. (The original Hacl.Impl.SHA3 does not an #a for absorb_inner.) If I follow the series of functions that are called (load_state, load_ws, load_blocks, set wsi_1_4, set_wsi), I see calls to helper functions such as `word_t a` but `word_t a == U64`, always, for any Keccak algorithm, so you did not need the `a` in the first place for any of these helpers.
- There is an API mismatch for absorb_inner. The original one takes a block whose length is rateInBytes, but you impose a block length of 256. Why? This means that a function that called the old API now needs to allocate a buffer of size 256 on the stack and then copy the input block into the 256-sized buffer before calling your new absorb_inner. This is an API mismatch that will cost performance. Can you fix it?
- In order to finish the work, not much is needed once you fix the above. I sketched for you what needs to happen in Hacl.Hash.SHA3. It's quite simple: any call to Hacl.Impl.SHA3 needs to be replaced with a call to Hacl.Vec.SHA3, followed by the corresponding lemma form Hacl.Spec.SHA3.Equiv. If you remove that admit (and fix other calls), then you don't need to do anything else and all of the other modules will now call your new implementation.

In particular, this means that you don't need to look into streaming as everything is properly abstracted behind Hacl.Hash.SHA3.fsti!

I'll let you work to address these review comments. And here's the sketch of what I was describing above:

```
diff --git a/code/hash/Hacl.Hash.SHA3.fst b/code/hash/Hacl.Hash.SHA3.fst
index 730c255d60..1ddb3f0f4c 100644
--- a/code/hash/Hacl.Hash.SHA3.fst
+++ b/code/hash/Hacl.Hash.SHA3.fst
@@ -11,6 +11,8 @@ open Lib.IntTypes
 
 module LB = Lib.Buffer
 
+module V = Hacl.Impl.SHA3.Vec
+
 friend Spec.Agile.Hash
 
 #push-options "--fuel 0 --ifuel 0 --z3rlimit 20"
@@ -65,13 +67,36 @@ let update_multi_sha3_st (a:keccak_alg) =
       B.as_seq h1 s ==
         Spec.Agile.Hash.update_multi a (B.as_seq h0 s) () (B.as_seq h0 blocks)))
 
+// A small wrapper that has the original signature (a scalar absorb_inner) but
+// rewired to call the vec implementation with width=1
+val absorb_inner:
+    rateInBytes:size_t{0 < v rateInBytes /\ v rateInBytes <= 200}
+  -> block:LB.lbuffer uint8 rateInBytes
+  -> s:Hacl.Impl.SHA3.state
+  -> ST.Stack unit
+    (requires fun h0 -> LB.(live h0 s /\ live h0 block /\ disjoint s block))
+    (ensures  fun h0 _ h1 ->
+      LB.(modifies (loc s) h0 h1 /\
+      as_seq h1 s ==
+      Spec.SHA3.absorb_inner (v rateInBytes) (as_seq h0 block) (as_seq h0 s)))
+
+let absorb_inner rateInBytes block s =
+  let h0 = ST.get () in
+  admit ();
+  Hacl.Impl.SHA3.Vec.absorb_inner #SHA3_256 #Hacl.Spec.SHA3.Vec.Common.M32 rateInBytes block s;
+  Hacl.Spec.SHA3.Equiv.absorb_inner_lemma_l #SHA3_256 #Hacl.Spec.SHA3.Vec.Common.M32
+    rateInBytes (LB.as_seq h0 block) (LB.as_seq h0 s) 0;
+  ()
+
+
 /// This function needs to be type-checked monomorphically (not with agile
 /// types) so as to extract to Low* as-is.
 let update_multi_sha3 (a: keccak_alg): update_multi_sha3_st a = fun s () blocks n_blocks ->
   [@inline_let]
   let spec_f = Spec.SHA3.absorb_inner (Spec.Hash.Definitions.rate a/8) in
   let h0 = ST.get () in
-  Lib.Buffer.loop_blocks_multi (block_len a) n_blocks blocks spec_f (Hacl.Impl.SHA3.absorb_inner (block_len a)) s
+  Lib.Buffer.loop_blocks_multi (block_len a) n_blocks blocks spec_f (absorb_inner (block_len a)) s
 
 /// There is a proof going here, that if your algorithm is of the keccak family,
 /// then the monomorphic, Low*-compatible signature above, is a refinement of
```

Let's use our regular sync-up tomorrow to go over this, or in 8 days, whatever you prefer. Thanks!

mamonet:
Thanks for your review.

> First look at this: Hacl.Impl.SHA3.Vec.fst does not type-check on my machine.
> 
> ```
> Verified module: Hacl.Impl.SHA3.Vec
> 56 errors were reported (see above)
> ```
> 
> is this expected? @mamonet

No, I updated FStar repo in order to get type-check failing. I am looking at the errors now and trying to make more reliable code.

> Some high-level thoughts.
> 
> * Please move the signatures into Hacl.Impl.SHA3.Vec.fsti -- this will make verification faster and more modular.

I did that and I got a lot of errors about unsolved implicit arguments, we may look at that together because I don't seem to figure it out. `sha2-mb` uses a similar but slightly different measure, it keeps the signatures in the same module, but it splits it to two modules `Hacl.Impl.SHA2.Core.fst` and `Hacl.Impl.SHA2.Generic.fst` let me know if it's an alternative way for what you've suggested.

> * Please remove the #a parameter from `absorb_inner` (and fix all of the other functions accordingly). The behavior of `absorb_inner` does not depend on your choice of algorithm. (The original Hacl.Impl.SHA3 does not an #a for absorb_inner.) If I follow the series of functions that are called (load_state, load_ws, load_blocks, set wsi_1_4, set_wsi), I see calls to helper functions such as `word_t a` but `word_t a == U64`, always, for any Keccak algorithm, so you did not need the `a` in the first place for any of these helpers.

I will remove redundant parameters in the functions you mentioned.

> * There is an API mismatch for absorb_inner. The original one takes a block whose length is rateInBytes, but you impose a block length of 256. Why? This means that a function that called the old API now needs to allocate a buffer of size 256 on the stack and then copy the input block into the 256-sized buffer before calling your new absorb_inner. This is an API mismatch that will cost performance. Can you fix it?

The buffer length has to be a multiple of 32 in order to fit in 256-bit vector registers, the buffer is padded with zeros in all cases. I can expose a new `absorb_inner` function that takes a block of size `rateInBytes` but I will allocate 256-byte buffer internally and copy that block into, same cost. The best I can do is to reduce the buffer length to 224-byte but not sure if that would do any better.

> * In order to finish the work, not much is needed once you fix the above. I sketched for you what needs to happen in Hacl.Hash.SHA3. It's quite simple: any call to Hacl.Impl.SHA3 needs to be replaced with a call to Hacl.Vec.SHA3, followed by the corresponding lemma form Hacl.Spec.SHA3.Equiv. If you remove that admit (and fix other calls), then you don't need to do anything else and all of the other modules will now call your new implementation.
> 
> In particular, this means that you don't need to look into streaming as everything is properly abstracted behind Hacl.Hash.SHA3.fsti!
> 
> I'll let you work to address these review comments. And here's the sketch of what I was describing above:

Thanks for the sketch, I will look at for sure.

> 
> Let's use our regular sync-up tomorrow to go over this, or in 8 days, whatever you prefer. Thanks!

I have dental appointments for the rest of this week, I can adjust the schedule but I am not able to speck very well :) let's stay in touch on Slack and I will let you know when I can do meetings.

msprotz:
> No, I updated FStar repo in order to get type-check failing. I am looking at the errors now and trying to make more reliable code.

So these were all working with a previous F* version and now broke? I hope this is only a helper function somewhere. Off the top of my head, I don't know of any deep changes in F* in that timeframe, except perhaps some stuff related to seq_of_list, which may or may not be relevant for you.

> I did that and I got a lot of errors about unsolved implicit arguments, we may look at that together because I don't seem to figure it out. 

Let's look at it together.

> The buffer length has to be a multiple of 32 in order to fit in 256-bit vector registers, the buffer is padded with zeros in all cases. I can expose a new absorb_inner function that takes a block of size rateInBytes but I will allocate 256-byte buffer internally and copy that block into, same cost. 

That sounds good. Let's try that, and then let's discuss more if the performance is not good.

> I have dental appointments for the rest of this week, I can adjust the schedule but I am not able to speak very well :) let's stay in touch on Slack and I will let you know once I can do meetings.

No worries! I'm on Slack.

mamonet:
> * Please move the signatures into Hacl.Impl.SHA3.Vec.fsti -- this will make verification faster and more modular.

Done!

> * Please remove the #a parameter from `absorb_inner` (and fix all of the other functions accordingly). The behavior of `absorb_inner` does not depend on your choice of algorithm. (The original Hacl.Impl.SHA3 does not an #a for absorb_inner.) If I follow the series of functions that are called (load_state, load_ws, load_blocks, set wsi_1_4, set_wsi), I see calls to helper functions such as `word_t a` but `word_t a == U64`, always, for any Keccak algorithm, so you did not need the `a` in the first place for any of these helpers.

Done!

> * There is an API mismatch for absorb_inner. The original one takes a block whose length is rateInBytes, but you impose a block length of 256. Why? This means that a function that called the old API now needs to allocate a buffer of size 256 on the stack and then copy the input block into the 256-sized buffer before calling your new absorb_inner. This is an API mismatch that will cost performance. Can you fix it?

`absorb_inner` is internal function, can you tell me the use cases of this function?
Modifying the signature requires restructure of higher-level functions and this is a considerable amount of changes that could also lead to spec mismatch.

msprotz:
Ok this is starting to look pretty good! For me, everything in code/sha3 is now good.

Let's focus on the changes required in code/hash, which I believe are rather small. Your goal is now to express SHA3 as an algorithm that fits the specifications in Spec.Agile.Hash.fst. This module captures all hash algorithms under a single specification. This means one can build other layers (streaming APIs, HMAC, HKDF, signature hashing, etc.) in HACL* once and for all for any hash algorithm, since they all obey the same laws and signatures from Spec.Agile.Hash.

Re-exporting SHA3 as one such hash algorithm is done in Hacl.Hash.SHA3, which is the last remaining file to fix before you can delete the old SHA3 implementation!

Now finally getting to your question, the reason I mentioned absorb_inner is because the "update_multi" (generic processing of N blocks by a hash function) specification is currently written as a `repeat_blocks` of `absorb_inner`. See https://github.com/hacl-star/hacl-star/blob/main/specs/Spec.Agile.Hash.fst#L47. This is why Hacl.Hash.SHA3.update_multi (processing N blocks with SHA3, exported with the signature of generic hashes) https://github.com/hacl-star/hacl-star/blob/main/code/hash/Hacl.Hash.SHA3.fst#L74 currently calls old SHA3's absorb_inner via Loops.loop_blocks.

In your specific case, you now need to fix Hacl.Hash.SHA3.update_multi to call Hacl.Impl.SHA3.Vec.absorb_nblocks. Then, you have two options:
- prove that absorb_nblocks is equivalent (in specification) to `repeat_blocks (absorb_inner ...)`
- change the specification in Spec.Agile.Hash to use V.absorb_inner_blocks, but then prove in Spec.Hash.Lemmas.fst that V.absorb_inner_blocks satisfies the hash laws.

code/sha3/Hacl.Impl.SHA3.Vec.fst
```diff
@@ -984,6 +984,7 @@ let get_multiblock_t (m:m_spec{is_supported m}) =
 inline_for_extraction noextract
 val get_multiblock_1: #m:m_spec{lanes m == 1} -> get_multiblock_t m
 
+#push-options "--max_ifuel 1 --max_fuel 1"
```
msprotz, line 987: fuel is for unrolling definitions in the smt solver; if you know that both of these are necessary, then the recommended style is to use `--fuel 1 --ifuel 1` to avoid first trying with 0/0, then failing, then trying with 1/1

mamonet: I replaced it with more proper options, my observation is that it's more reliable to use basic arithmetics of FStar than the ones in `Lib.IntTypes` e.g. for `get_multiblock_1` function we can use `size (v i * v rateInBytes)` instead of `i *! rateInBytes` that makes the verification process more stable and faster even when assisting the resolver with the following assert `assert (v (i *! rateInBytes) == v i * v rateInBytes)` and that could go for other operations where possible.

mamonet:
I utilized the new vectorized functions within `Hacl.Hash.SHA3` and managed to extract C files for SHA3 modules.

msprotz:
Here's what remains to be done.
- @karthikbhargavan we need your review. There seems to be excessive inlining in Hacl_Hash_SHA3.c -- please assess and advise Mamone.
- Missing primitive from libintvector.h:
```
Hacl_Hash_SHA3_Simd256.c:75:14: error: call to undeclared function 'Lib_IntVector_Intrinsics_vec256_load64_le'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    ws[0U] = Lib_IntVector_Intrinsics_vec256_load64_le(b0);
    ```
@karthikbhargavan please assess, and either patch hacl-star/lib/c/libintvector.h or advise Mamone
- wasm code-gen bug in krml: that one is on me

thanks

File: dist/gcc-compatible/Hacl_Hash_SHA3.h [hunk 9]
```diff
@@ -77,6 +77,8 @@ uint32_t Hacl_Hash_SHA3_hash_len(Hacl_Hash_SHA3_state_t *s);
 
 bool Hacl_Hash_SHA3_is_shake(Hacl_Hash_SHA3_state_t *s);
 
+void Hacl_Hash_SHA3_absorb_inner_32(uint32_t rateInBytes, uint8_t *b, uint64_t *s);
```
msprotz, line 80: right that's the only unpleasant bit from the C code change... if we wanted to fix it we could mark this function as private in Hacl.Impl.SHA3.Vec.fst and have Hacl.Hash.SHA3 `friend` Hacl.Impl.SHA3.Vec... not sure it's worth the trouble.

msprotz:
@mamonet please regenerate dist/test/c/Hacl_Test_SHA3.c and push

msprotz:
(after you upgrade krml to the latest version, I pushed a fix for your issue)

msprotz:
```
* Error 19 at /build/source/code/sha3/Hacl.Impl.SHA3.Vec.fst(1201,42-1201,52):
hacl-star>   - Assertion failed
hacl-star>   - See also prims.fst(420,83-420,96)
```

@mamonet can you please strengthen this proof?

dist/gcc-compatible/Hacl_Hash_SHA3.h
msprotz: coming back to this, I actually don't know why this ends up in the header... because the function definitely seems private so I would've expected it to end up in a private header (in `internal/`)... but let's fix this in a followup PR there's been enough happening in this one already

msprotz:
once this proof is fixed, the code will be ready to merge for me too

@karthikbhargavan when you have a chance can you try to formally approve this PR? we need two reviews to merge... thanks!

msprotz:
```
Error 19 at /build/source/code/sha3/Hacl.Spec.SHA3.Equiv.fst(852,2-852,10):
hacl-star>   - Assertion failed
hacl-star>   - The SMT solver could not prove the query. Use --query_stats for more
hacl-star>     details.
hacl-star>   - See also /build/source/lib/Lib.Sequence.fsti(87,56-87,80)
```

@mamonet

code/sha3/Hacl.Spec.SHA3.Equiv.fst
```diff
@@ -849,19 +864,23 @@ let storeState_inner_unroll_sub_12_18 #m s i =
   let b24 = Spec.storeState_inner s 24 b23 in
   let block = storeState_inner_unroll s in
   //s.[12]
```
msprotz, line 866: don't know what rlimit you're running with here, but FYI I wouldn't find it shocking if the default value for the whole file was to be 100

msprotz:
Congrats! This is green. Please ask @karthikbhargavan for a second review and a merge :-)

karthikbhargavan (approved):
Great work, thanks Mamone and Jonathan!

PR #903, opened 2023-12-22
Update HACL to deal with a code-gen change in krml

msprotz:
A change that recently landed on krml concerns external, type-polymorphic functions.
- A long while ago, these were not supported, and the only way to handle them was to have dedicated support in krml. There were two such functions: Lib.Memzero.memzero (for any array), and LowStar.Ignore.ignore (for any value).
- About a couple months ago, type-polymorphic external declarations became supported in the internal AST, meaning that the custom support for memzero and ignore went away. The patch made it so that a call to a type-polymorphic external function is emitted as a regular function call, with an extra argument for the type parameter. The idea is to implement these external functions by hand, with a macro that does something useful with the type. Example: a call to `val memzero: #a:Type -> b:buffer a -> l:uint32 { l == len b } -> ST unit` becomes `memzero(b, l, <<concrete value of type a; e.g. uint64_t>>)`, which in turn allows for an implementation such as `#define memzero(b, l, t) (secure_memzero0(b, l * sizeof t))`
- As I started expanding the usefulness of these type-polymorphic external declarations, it became clear that one extra thing was needed: the macro needs to *receive* the return type as well. Consider `val foobar: #a -> state a -> option a` to be implemented as a macro (because, say, `state` is a system data structure and `foobar` is part of the glue to interact with such a system). In order to implement this macro, `foobar` needs to *know* the name of the return type, e.g. `#define foobar(s, t, return_t) (check(s) ? ((return_t){ .tag = Some, .val = take(s) }) : ((return_t){ .tag = None })`. Hence the krml patch to add the return type to the macro invocation.

(Astute readers might suggest that `return_t` can be derived by token concatenation in the C preprocessor, e.g. by doing `Pervasives_option_##t`, but this doesn't work because 1. the prefix might not contains the `Pervasives_` part if the specific instance of option is not part of the public header, and 2. as soon as `t` is not a base type, the macro will emit e.g. `t*`, generating an invalid token.)

Anyhow, this is a pretty minor change but just wanted to give a rationale for why we have so many lines in HACL affected. Thanks.


msprotz:
Oh and the Everest build won't be green until this is merged. Heads-up @mtzguido and @R1kM

mtzguido (approved):

karthikbhargavan (approved):
The invocations of `memzero` always appear to use `void*` as a return type, and this type is ignored in the macro, so it looks a bit weird. If we could use a more meaningful return type here, even if it is a macro called "default" it may look slightly better.

msprotz:
Ok I can submit a followup, I'll just merge this in order to unblock the build.

Void * is to represent values of type unit. It could conceivably be emitted as KRML_UNIT

PR #912, opened 2024-02-02
Update some proofs for changes to F*'s Seq

mtzguido:
## Proposed changes

These are fixes for FStarLang/FStar#3203. 

For the SHA files, the proof was relying on the normalization of `seq_of_list` to consume the list and hence obtain the length. Since now `seq_of_list` does not have an exposed definition, this fails. Instead, we normalize the *list( to compute its length, and then the refinement on the result of `seq_of_list` (i.e. that the lenghts match) allow us to recover this assertion.

The last diff is just pointing to `seq_of_list` in its new location, and removing a now-useless reveal_opaque.

## Types of changes

## Checklist

### DIFF
--- a/code/sha2-mb/Hacl.Spec.SHA2.Equiv.fst
+++ b/code/sha2-mb/Hacl.Spec.SHA2.Equiv.fst
@@ -66,7 +66,8 @@ val seq_of_list_is_create8: #a:Type -> x0:a -> x1:a -> x2:a -> x3:a -> x4:a -> x
 
 let seq_of_list_is_create8 #a x0 x1 x2 x3 x4 x5 x6 x7 =
   let rp = Seq.seq_of_list [x0; x1; x2; x3; x4; x5; x6; x7] in
-  assert_norm (length rp == 8);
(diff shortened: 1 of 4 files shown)


msprotz (approved):
Looks good to me. Thanks so much Guido!

Will this optimized implementation of seq_of_list also benefit the normalizer?

R1kM (approved):

mtzguido:
> Will this optimized implementation of seq_of_list also benefit the normalizer?

In a way : )... seq_of_list and seq_to_list are now opaque (behind an interface), so the normalizer will not ever unfold them or reduce them.

Previously `seq_of_list [1;2;3]` would reduce to something like `create 1 1 @| create 1 2 @| create 1 3 @| empty`, which is not very good IMO. Now, that does not happen, but the proofs still go through since we have lemmas saying e.g. `seq_of_list (x::xs) == append (create 1 x `append` seq_of_list xs)`.

Since `seq` is an abstract type (externally) I think we also cannot provide a sensible reduction rule. But definitely let me know if this is a problem and we can revisit.

PR #917, opened 2024-02-26
CI: update action `create-pull-request`

pnmadelaine:
The update is required as v5 still uses Node.js 16 which was deprecated


R1kM (approved):
Thanks for the quick fix!

msprotz (approved):

PR #918, opened 2024-03-01
Afromher rs

msprotz:
## Proposed changes

Preliminary support for Rust extraction of HACL*. This is joint work with @R1kM.

Enough work has gone into this that I would like to see things being put under CI so as to make sure nothing regresses. I have no intention of advertising this as things are pretty rough still.

## Types of changes

This leverages the work of the past several months on the Rust backend of KaRaMeL and enables it within HACL*. Notably, this creates:
- a new subdirectory, dist/rs
- puts under version control a complete extraction of HACL* to Rust, modulo KaRaMeL failures for some of the functions
- adds some hand-written files for the library modules (hacl.rs, lib.rs, etc.), the machine integers (eventually should be extracted, but hand-written for now)
- adds targets in the main Makefile to extract and build with cargo

## What to look for in this review

- Many small, non-threatening changes had to be enacted on the original F* source in order to comply with the restrictions of the Rust borrow-checker. As such, the first thing to review is the diff for dist/gcc-compatible; I would need some performance benchmarks to make sure there are no regressions.
- Then, one can look at the shape of the generated Rust code, the build, and debate whether this is an acceptable layout, etc.

## Limitations

- Many of these modules compile, but have run-time errors. @R1kM has a plan to fix this once and for all, but for now, only the addition of tests and/or fuzzing will help us figure out which modules have issues.
- More modules need to be rewritten to observe a buffer ownership pattern that is compatible with Rust. HMAC, HKDF, Ed25519 all requires modifications in the source.
- More modules need to be rewritten to obey aliasing restrictions and avoid in-place APIs (Ed25519).
- There are likely many performance issues, the chief one that I'm aware of being the fact that we take mutable pointers to globally-allocated constants, which generates a copy of the constant per the corresponding Rust warning. This is quite concerning, and I don't know if this will be optimized at the LLVM level. In any case, this is something I'd like to fix. While we could try to port the source to leverage LowStar.ConstBuffer more, this is likely going to be a herculean task, and I'd rather implement a const-analysis phase in krml in order to automatically decorate all pointers (and function signatures) with const whenever possible. It's still a fair amount of work, but not as bad as fixing the source.


msprotz:
@pnmadelaine can you add `uucp` to the build? I thought I had fixed the karamel flake properly but I must've done something wrong...?

msprotz:
@franziskuskiefer bringing your attention to this, as it might impact performance on HACL packages

msprotz:
@pnmadelaine @twal @W95Psp 

hey Nix gurus, I added cargo here: https://github.com/hacl-star/hacl-star/pull/918/commits/d354afe29fb33877da9066b9475ba986b03dd6b6 but I'm still getting "cargo: command not found" in the build logs

can someone help me on the path to Nix enlightenment?

thanks!!

goffrie:
> ```
> hacl-star> error: could not find `Cargo.toml` in `/build/source/dist/rs` or any parent directory
> ```

the source files that make it into the Nix build are filtered here: https://github.com/hacl-star/hacl-star/blob/0dfadb6fb997020c372d2201c200f260b5c812c0/.nix/hacl.nix#L47-L73

msprotz:
> the source files that make it into the Nix build are filtered here:
> 

ugh... the directory is a mix of hand-written files and autogenerated files, do I really need to manually list all the files that are hand-written?

msprotz:
FYI CI is now green, and includes extracting, building and testing the Rust code.

msprotz:
HACL packages CI is now green, which should allow us to get a perf comparison between:
- HACL/main and HACL/afromher_rs, in C -- how many of these source changes slow down the C code?
- HACL/afromher_rs, in C vs. HACL/afromher_rs, in Rust -- how much slower are the Rust versions?

@franziskuskiefer sounded like he would give this a stab, since eventually we want all of this to go into libcrux.

pnmadelaine:
Sorry, I missed the notifications about the Nix questions!

> ugh... the directory is a mix of hand-written files and autogenerated files, do I really need to manually list all the files that are hand-written?

I know it is a bit tiresome, but this is really useful to avoid rebuilding when only the generated files change!

pnmadelaine:
@msprotz it seems like the source filter can be written a little bit more compactly, see https://github.com/hacl-star/hacl-star/tree/pnmadelaine/afromher_rs

msprotz:
Awesome, can you submit a PR?

spitters:
What is the status of this? We'd like to use it here: https://github.com/cryspen/libcrux/issues/225 for ElectionGuard integration.

R1kM:
@msprotz and myself are working this week on significant improvements to the codebase (especially related to the pervasive use of mutable borrows). The version currently on the branch is still considered experimental, however, modules that are not commented out in hacl.rs compile and pass tests, although with variable performance depending on the algorithms.

spitters:
Thanks! This solves the issue we had with compiling to wasm https://github.com/cryspen/libcrux/issues/225
Are you planning to release this as a separate crate?

msprotz:
Right now we have isolated the bignums as a standalone crate. @marshray said that this should work for him to integrate into electionguard. how about concordium? 

If the standalone bignum crate works for you, we could contemplate splitting up more algorithms as separate crates.

msprotz:
- would you be ok consuming the entire HACL* library as Rust? or do you need hmac in a separate sub-crate like bignum?
- do you need it published on crates.io or are you ok consuming directly from a branch of HACL*?

spitters:
@bshvass made a version of EG with HACL* hmac here:
https://github.com/bshvass/electionguard-rust/tree/hacl-star-rs

@marshray may want to wait with including this until @msprotz has made a release.

The work above does not include the HACL* bignums yet.

msprotz:
Is there any feedback on the integration? Happy to take comments, suggestions for improvement...

PR #919, opened 2024-03-09
Import AES-CTR32 modules from pnmadelaine_aes branch and add NI proofs

mamonet:
This PR imports AES-CTR32 modules from `pnmadelaine_aes` branch and adds proofs for NI functions.
M32 (CoreBitSlice) module is added without proofs yet, CoreNI and Generic modules are fully proven.


dist/Makefile.tmpl
```diff
@@ -49,6 +49,9 @@ include Makefile.basic
 
 CFLAGS		+= -Wno-parentheses -Wno-deprecated-declarations -Wno-\#warnings -Wno-error=cpp -Wno-cpp -g -std=gnu11 -O3
 
+CFLAGS_AES_NI = -maes -msse4.1
```
msprotz, line 52: probably doesn't work for ARM at all -- you probably need to add further changes into `dist/configure` and do something like

```
compile_aesni=true
```

(in `detect_x64`)

then 

```
if ! $compile_aesni; then
  echo "$CC $CROSS_CFLAGS cannot compile NI, disabling AES-NI"
  echo "BLACKLIST += $(ls *AES*NI*.c | xargs)" >> Makefile.config
else
  echo "#define HACL_CAN_COMPILE_AESNI 1" >> config.h
fi
```

msprotz (commented):
Can you show us the C code, too?

This is only when the CPU has NI support, correct?

dist/Makefile.tmpl
msprotz: and running your test would then be guarded in tests/Makefile just like, say, the vec128 test, for instance

Makefile.common
```diff
@@ -225,6 +225,7 @@ BIGNUM_BUNDLE= \
   -bundle Hacl.Bignum.Base,Hacl.Bignum.Addition,Hacl.Bignum.Convert,Hacl.Bignum.Lib,Hacl.Bignum.Multiplication[rename=Hacl_Bignum_Base] \
   -static-header Hacl.Bignum.Base,Hacl.Bignum.Addition,Hacl.Bignum.Convert,Hacl.Bignum.Lib,Hacl.Bignum.Multiplication \
   -bundle Hacl.Bignum,Hacl.Bignum.*[rename=Hacl_Bignum]
+AES_CTR32_BUNDLE=-bundle Hacl.AES_128.CTR32.NI=Hacl.Impl.AES.* -bundle Hacl.AES_256.CTR32.NI=Hacl.Impl.AES.*
```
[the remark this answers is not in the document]
mamonet: I tried `AES_CTR32_BUNDLE=-bundle Hacl.Impl.AES.*[rename=Hacl_AES_Common]` but it yields to the following warrning
```
Warning 18: After bundling, two C files are named Hacl_AES_Common
Warning 18 is fatal, exiting.

File: dist/Makefile.tmpl (hunk 1 above)
Comment (mamonet) [reply dist/Makefile.tmpl]: I made this change, however I am not able test this function https://github.com/mamonet/hacl-star/blob/aes-ctr32/dist/configure#L203 I would like for you to review the change and test this function to make sure everything's ok

File: Makefile.common (hunk 2 above)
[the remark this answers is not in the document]
Comment (msprotz) [reply Makefile.common:228]: yeah just remove the second occurrence of Hacl.Impl.AES.* -- it's just confusing since this is not how bundles work

File: dist/Makefile.tmpl (hunk 1 above)
Comment (msprotz) [reply dist/Makefile.tmpl]: on aarch64 (raspberry pi linux):

```
jonathan@raspberrypi:/tmp/hacl-star/dist/gcc-compatible $ ./configure
bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
... detected ARM platform
...  supports 128-bit NEON
... cc  can cross-compile to ARM64 with SIMD
...  supports cryptographic instructions
... cc  can cross-compile to ARM64 with crypto extension
 does not support x64 assembly, disabling Curve64
 does not support legacy vale stubs
 does not support 256-bit arithmetic
... glibc is recent enough for explicit_bzero
```

on apple m3 (osx)

```
... detected ARM platform
...  supports 128-bit NEON
... cc  can cross-compile to ARM64 with SIMD
...  supports cryptographic instructions
... cc  can cross-compile to ARM64 with crypto extension
 does not support x64 assembly, disabling Curve64
 does not support legacy vale stubs
 does not support 256-bit arithmetic
 ```

that doesn't mean the raspberry pi can *run* aesni but at least it can compile it

so this part looks good to me

code/aes/Hacl.Impl.AES.Generic.fst
```diff
@@ -0,0 +1,1800 @@
...
+  (ensures (fun h0 b h1 -> modifies (loc ctx) h0 h1 /\
+    (let n = LSeq.sub (as_seq h1 ctx) 0 (v (nlen m)) in
+    let kex = LSeq.sub (as_seq h1 ctx)
+      (v (nlen m)) ((Spec.num_rounds a+1) * v (klen m)) in
+    let state = Spec.aes_ctr32_init_LE a (as_seq h0 key)
+      (as_seq h0 nonce) in
+    nonce_to_bytes m n == state.block /\
+    keyx_to_bytes m a kex == state.key_ex)))
+
+let aes_init #m #a ctx key nonce =
```
msprotz, line 1327: I take it that a toolchain supports either building both variants, or none at all?

code/aes/Makefile
```diff
@@ -0,0 +1,35 @@
+HACL_HOME=$(realpath ../..)
```
msprotz, line 1: does this work? has this been tested a little bit?

msprotz: the whole local Makefile, I mean

dist/configure
```diff
@@ -416,6 +439,9 @@ if detect_x64; then
     echo "... $build_target supports our inline ASM"
     compile_inline_asm=true
   fi
+  echo "... $build_target supports AES/PCLMUL instruction set extensions"
+  compile_aesni=true
+  echo "CFLAGS_AESNI = -maes -msse4.1" >> Makefile.config
```
msprotz, line 444: do all toolchains know these build flags? I suppose so... @karthikbhargavan do we support ancient redhats still for Mozilla? maybe irrelevant for the purposes of our own CI

specs/Spec.AES.fst
```diff
@@ -423,3 +436,179 @@ let aes128_ctr_key_block1 key n_len n =
...
+  -> msg:bytes{length msg / 64 <= max_size_t} ->
+  Tot (ciphertext:bytes{length ciphertext == length msg})
+
+let aes_ctr32_encrypt_bytes_LE v key nonce ctr0 msg =
+  let st0 = aes_ctr32_init_LE v key nonce in
+  aes_ctr32_encrypt_stream_LE v st0.key_ex
+    st0.block ctr0 msg
+
+let aes_ctr32_decrypt_bytes_LE v key nonce ctr0 msg =
+  aes_ctr32_encrypt_bytes_LE v key nonce ctr0 msg
```
msprotz, line 614: where is the proof that encrypt (decrypt x) == x?

tests/aes128-ctr32-ni-test.c
```diff
@@ -0,0 +1,105 @@
...
+  }
+  b = cpucycles_end();
+  t2 = clock();
+  clock_t tdiff1 = t2 - t1;
+  cycles cdiff1 = b - a;
+
+  printf("AES-NI PERF:\n");
+  printf("cycles for %" PRIu64 " bytes: %" PRIu64 " (%.2fcycles/byte)\n",count,(uint64_t)cdiff1,(double)cdiff1/count);
+  printf("time for %" PRIu64 " bytes: %" PRIu64 " (%.2fus/byte)\n",count,(uint64_t)tdiff1,(double)tdiff1/count);
+  printf("bw %8.2f MB/s\n",(double)count/(((double)tdiff1 / CLOCKS_PER_SEC) * 1000000.0));
```
msprotz, line 102: your main is not returning a value

is this checking that the results are correct?

msprotz (commented):
Various tidbits to be fixed. I would really like some thorough testing to make sure this is producing the correct results, with a variety of test cases, on arm and x64. The fact that the arguments to the (unverified) macro glue were in the wrong order makes me... slightly nervous

I would also like to see the incremental API being tested out in the C file.

Makefile.common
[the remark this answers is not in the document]
mamonet: Done!

code/aes/Hacl.AES_128.CTR32.NI.fst
```diff
@@ -0,0 +1,158 @@
...
+
+module ST = FStar.HyperStack.ST
+module LSeq = Lib.Sequence
+module Spec = Spec.AES
+
+let aes_ctx = aes_ctx MAES Spec.AES128
+let skey = skey Spec.AES128
+
+[@ CInline ]
+val create_ctx: unit ->
```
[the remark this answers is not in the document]
mamonet: Right, I replaced `create_ctx` with functions that allocate and free buffer on the heap.

code/aes/Hacl.AES_128.CTR32.NI.fst
```diff
@@ -0,0 +1,158 @@
...
+[@ CInline ]
+val aes128_init:
+    ctx: aes_ctx
+  -> key: skey
+  -> nonce: lbuffer uint8 12ul ->
+  Stack unit
+  (requires (fun h -> live h ctx /\ live h nonce /\ live h key /\ disjoint ctx key /\
+    as_seq h ctx == Seq.create (size_v (ctxlen MAES Spec.AES128)) (elem_zero MAES)))
+  (ensures  (fun h0 _ h1 -> modifies1 ctx h0 h1 /\
+    (let n = LSeq.sub (as_seq h1 ctx) 0 (v (nlen MAES)) in
```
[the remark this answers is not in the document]
mamonet: Done!

code/aes/Hacl.AES_128.CTR32.NI.fst
```diff
@@ -0,0 +1,158 @@
...
+
+
+[@ CInline ]
+val aes128_set_nonce:
+    ctx: aes_ctx
+  -> nonce: lbuffer uint8 12ul ->
+  Stack unit
+  (requires (fun h -> live h ctx /\ live h nonce))
+  (ensures  (fun h0 _ h1 -> modifies1 ctx h0 h1 /\
+    nonce_to_bytes MAES (LSeq.sub (as_seq h1 ctx) 0 (v (nlen MAES))) ==
```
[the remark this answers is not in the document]
mamonet: Done!

code/aes/Makefile
mamonet: It does work, we used the local Makefile to test the implementation but since we added support for aes test in tests/Makefile we can remove this one.

code/aes/Hacl.AES_128.CTR32.NI.fst
```diff
@@ -0,0 +1,158 @@
...
+    let n = LSeq.sub (as_seq h0 ctx) 0 (v (nlen MAES)) in
+    as_seq h1 kb == u8_16_to_le (Spec.aes_encrypt_block Spec.AES128
+      (keyx_to_bytes MAES Spec.AES128 k) (Spec.aes_ctr32_set_counter_LE
+      (nonce_to_bytes MAES n) counter)))))
+
+[@ CInline ]
+let aes128_key_block kb ctx counter = aes_key_block #MAES #Spec.AES128 kb ctx counter
+
+
+inline_for_extraction noextract
```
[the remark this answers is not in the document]
mamonet: I don't see a use case for this function, I think we can remove it.

code/aes/Hacl.AES_128.CTR32.NI.fst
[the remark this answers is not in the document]
msprotz: you'll still need the variant that allocates on the stack when you write aes-gcm and want to allocate aes state on the caller's stack

code/aes/Hacl.AES_128.CTR32.NI.fst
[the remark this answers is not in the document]
msprotz: 👍

code/aes/Hacl.Impl.AES.Core.fst
```diff
@@ -0,0 +1,87 @@
...
+module LSeq = Lib.Sequence
+
+
+#set-options "--z3rlimit 50"
+
+
+inline_for_extraction noextract
+let create_state (#m:m_spec) =
+  match m with
+  | MAES -> Hacl.Impl.AES.CoreNI.create_state()
```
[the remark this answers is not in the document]
msprotz: 👍

code/aes/Makefile
msprotz: keep the Makefile, I just wanted to make sure we didn't check in a Makefile that hadn't been tested before

tests/aes128-ctr32-ni-test.c
mamonet: It checks that the results are correct but only print it, I made the main function exits with a value.

msprotz: ok, there are some helpers in the test directory, so please use test_helpers.h instead of redefining all of these macros again (not your fault: this was probably done like that in the original branch)

notably, please use compare_and_print

mamonet:
> The fact that the arguments to the (unverified) macro glue were in the wrong order makes me... slightly nervous
> 
The order of arguments were flipped in the implementation side of F* code, but it doesn't meet the argument order in spec. To make the order consistence, I could either change the order for impl and spec or in libintvector.h I went for the later.

specs/Spec.AES.fst
mamonet: that proof should be applied for `aes_ctr32_decrypt_bytes_LE` function?

code/aes/Hacl.AES_128.CTR32.NI.fst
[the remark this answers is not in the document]
mamonet: Restored and marked as inline_for_extraction noextract.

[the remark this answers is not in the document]
msprotz: great, FYI, we've been calling the functions `alloca` in other modules, so please use that name for consistency

tests/aes128-ctr32-ni-test.c
mamonet: Done!

specs/Spec.AES.fst
msprotz: yes -- it should be pretty trivial in this case but it's a good "sanity check" to make sure you can prove that?

code/aes/.gitignore
msprotz, line 1: is this empty file needed?

code/aes/AUTHORS.md
```diff
@@ -0,0 +1 @@
+This code was primarily written by Karthikeyan Bhargavan (INRIA).
```
msprotz, line 1: @karthikbhargavan can you find a phrasing that incorporates @mamonet's contributions?

specs/Spec.AES.fst
mamonet: Now I'm thinking about it, there shouldn't have been `aes_ctr32_decrypt` function. GCM uses the forward version of AES cipher for encryption and decryption. I checked Vale AES/GCTR spec and there's no decryption function. See https://github.com/hacl-star/hacl-star/blob/main/vale/specs/crypto/Vale.AES.AES_s.fst
https://github.com/hacl-star/hacl-star/blob/main/vale/specs/crypto/Vale.AES.GCTR_s.fst

mamonet: Vale has no such lemma, in fact Vale doesn't have decryption function for AES/GCTR spec. See  https://github.com/hacl-star/hacl-star/blob/main/vale/specs/crypto/Vale.AES.AES_s.fst
https://github.com/hacl-star/hacl-star/blob/main/vale/specs/crypto/Vale.AES.GCTR_s.fst
GCM uses the forward version of AES cipher for encryption and decryption, so I think it's good to have `aes_ctr32_decrypt` unlike Vale but a lemma for encrypt (decrypt x) == x is to verify the spec itself and not the implementation.

code/aes/.gitignore
mamonet: yes, I removed that empty file.

code/aes/AUTHORS.md
karthikbhargavan: Done

dist/configure
karthikbhargavan: Let's just make sure it runs on our CI and on hacl packages. NSS has its own constraints and we will deal with them over there.

code/aes/Hacl.AES_128.CTR32.NI.fst
```diff
@@ -0,0 +1,164 @@
...
+    disjoint out inp))
+  (ensures (fun h0 _ h1 -> modifies (loc out) h0 h1 /\
+    as_seq h1 out == Spec.aes_ctr32_encrypt_bytes_LE Spec.AES128
+    (as_seq h0 k) (as_seq h0 n) counter (as_seq h0 inp)))
+
+let aes128_ctr_encrypt len out inp k n c = aes_ctr_encrypt #MAES #Spec.AES128 len out inp k n c
+
+
+[@ CInline ]
+val aes128_ctr_decrypt:
```
msprotz, line 150: @pnmadelaine can you review this naming convention? I thought with your big PR, we were now trying to avoid redundant prefixes, such as `aes128_` or `state_`

dist/gcc-compatible/Hacl_AES_128_CTR32_NI.h
```diff
@@ -0,0 +1,86 @@
...
+#endif
+
+#include <string.h>
+#include "krml/internal/types.h"
+#include "krml/lowstar_endianness.h"
+#include "krml/internal/target.h"
+
+#include "libintvector.h"
+
+typedef Lib_IntVector_Intrinsics_vec128 *Hacl_AES_128_CTR32_NI_aes_ctx;
```
msprotz, line 40: I do not see state_malloc?

dist/gcc-compatible/Hacl_AES_256_CTR32_NI.h
```diff
@@ -0,0 +1,95 @@
...
+#include "krml/lowstar_endianness.h"
+#include "krml/internal/target.h"
+
+#include "libintvector.h"
+
+typedef Lib_IntVector_Intrinsics_vec128 *Hacl_AES_256_CTR32_NI_aes_ctx;
+
+typedef uint8_t *Hacl_AES_256_CTR32_NI_skey;
+
+void
```
msprotz, line 44: same here, where is malloc/free?

dist/gcc-compatible/Hacl_AES_128_CTR32_NI.h
mamonet: I updated C files in gcc-compatible/
I also renamed `state_malloc/free` to `context_malloc/free`

dist/gcc-compatible/Hacl_AES_256_CTR32_NI.h
mamonet: Likewise.

franziskuskiefer:
@msprotz @karthikbhargavan what's the state here?

code/aes/Hacl.AES_128.CTR32.NI.fst
pnmadelaine: indeed, this should be named `ctr_decrypt`

dist/gcc-compatible/Hacl_AES_128_CTR32_NI.h
```diff
@@ -0,0 +1,106 @@
...
+
+  decrypt number of bytes in AES-CTR32 mode.
+  
+  `counter` is the initial value of counter state.
+  
+  Decryption uses the forward version of AES cipher
+*/
+void
+Hacl_AES_128_CTR32_NI_ctr_decrypt(
+  uint32_t len,
```
msprotz, line 93: same

dist/gcc-compatible/Hacl_AES_128_CTR32_NI.h
```diff
@@ -0,0 +1,106 @@
...
+/**
+Free AES-128 context buffer
+*/
+void Hacl_AES_128_CTR32_NI_context_free(Lib_IntVector_Intrinsics_vec128 *s);
+
+/**
+Initiate AES-128 context buffer with key expansion and nonce
+*/
+void
+Hacl_AES_128_CTR32_NI_init(Lib_IntVector_Intrinsics_vec128 *ctx, uint8_t *key, uint8_t *nonce);
```
msprotz, line 58: It's good to have a comment, but it's rephrasing what's already visible from the name of the function and the arguments. The questions that should be answered by the comment are:
- can I call this after encryption?
- does it *reset* the state to an initial state? should it be called reset? @pnmadelaine 
- does it reset the nonce to zero?

tests/Makefile
```diff
@@ -59,6 +59,13 @@ TARGETS += uint128-intrinsics-test.exe
 endif
 endif
 
+# AESNI
+ifneq ($(COMPILE_AESNI),)
+CFLAGS += -DHACL_CAN_COMPILE_AESNI
+else
+TARGETS := $(filter-out aes128-ctr32-ni-test.exe, $(TARGETS))
```
msprotz, line 66: nit: just put `%-ni-test.exe` here so that future tests are filtered out too

msprotz (commented):
Good to me once the comments, parameter names, and the little tidbit in the test file are fixed. 

I want @pnmadelaine to look closely at the refreshed C header files and confirm whether the naming convention is correct or not. Let's merge only after PNM confirms we're good. Thank you.

tests/Makefile
mamonet: Done!

dist/gcc-compatible/Hacl_AES_128_CTR32_NI.h
mamonet: AES-CTR32 API doesn't maintain the state of the counter, that state is supposed to be handled by the caller who is supposed to pass the counter though parameters for Encrypt/Decrypt functions (Proper doc is added to describe that) and this is by design. TBH I don't see what's the point since GCM is the only spec that utilizes this algorithm `AES-CTR32`, and basically it's implemented after GCM spec.

dist/gcc-compatible/Hacl_AES_128_CTR32_NI.h
```diff
@@ -0,0 +1,106 @@
...
+/**
+Initiate AES-CTR32-128 context with key and nonce, and
+
+  encrypt number of bytes in AES-CTR32 mode.
+  
+  `counter` is the initial value of counter state.
+*/
+void
+Hacl_AES_128_CTR32_NI_ctr_encrypt(
+  uint32_t len,
```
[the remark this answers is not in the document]
mamonet: Done!

dist/gcc-compatible/Hacl_AES_128_CTR32_NI.h
mamonet: Done!

PR #923, opened 2024-03-23
Extend implementation of Blake2 to support extended parameters

R1kM:
## Proposed changes

This PR is a follow-up to #864, which extended the Blake2 specs to include extra parameters. This PR propagates the changes to the implementation, and allows the use of parameters in C.
Parameters only have an impact at initialization time. To support them, this PR extends the streaming functor,
and generates new parameter-specific init/malloc/reset functions to avoid changing existing APIs.

This is needed for https://github.com/python/cpython/issues/99108.
I'm marking this PR as draft until we confirmed that the new exposed APIs are sufficient for integration into Python's hashlib.

## Types of changes


R1kM:
Subsumed by #930, closing

PR #926, opened 2024-03-27
Improve handling of Blake2 parameters in Spec

R1kM:
## Proposed changes

The previous extension of Blake2 parameters spec (cf #864) relied on a reference implementation corresponding to BLAKE2x (https://www.blake2.net/blake2x.pdf). The current Blake2 draft (https://www.blake2.net/blake2.pdf), as implemented by, e.g., Python, has several differences, including no support for extendable output functions (XOF), which in turn induces larger parameters for tree-hashing.

This PR fixes the spec to match parameter-handling from the current Blake2 standard, and removes the xof_length field from parameters.


karthikbhargavan (approved):
Looks good. Do the Blake2 authors or someone else have some more comprehensive test vectors somewhere?

PR #927, opened 2024-03-28
Blake2, Part 1-3/5: Extract with key support

msprotz:
As part of the blake2 improvements for python, the first step is generating a streaming API that can accept keys, optionally, without duplicating the entire codebase.


msprotz:
The main visible change is that the key length is now stored at runtime. This is necessary to successfully implement the `copy` function.

The main change in the proofs is that:
- the if-then-else over the key type is gone, this was causing a lot of difficulties
- the blake2 block class is no longer parametric over the key length: this removes a runtime argument that was going to show up unconditionally in the `update` function (and others), and simplifies the proofs

msprotz:
EverCrypt.Poly1305.fst started misbehaving (no clue why). I had to use a different z3seed, but CI may be of a different opinion since it's OSX z3 vs. Linux z3.

code/streaming/Hacl.Streaming.Blake2.Common.fst
```diff
@@ -139,77 +147,45 @@ let stateful_blake2 (a : alg) (m : m_spec) : I.stateful unit =
...
-let stateful_key (a : alg) (kk : key_size a) :
-  I.stateful unit =
+let stateful_key (a : alg):
+  I.stateful (key_size_t a) =
   I.Stateful
-    (fun _ -> stateful_key_t a kk)
+    (fun kk -> stateful_key_t a kk)
     (* footprint *)
-    (fun #_ h s -> if kk = 0 then B.loc_none else B.loc_addr_of_buffer (s <: B.buffer uint8))
+    (fun #kk h s -> if kk = 0ul then B.loc_none else B.loc_addr_of_buffer ((snd s) <: B.buffer uint8))
```
R1kM, line 164: Since we now always have a buffer in the state, do we need the disjunction here ?

dist/gcc-compatible/Hacl_Hash_Blake2b.c
```diff
@@ -679,46 +679,138 @@ void Hacl_Hash_Blake2b_finish(uint32_t nn, uint8_t *output, uint64_t *hash)
   Lib_Memzero0_memzero(b, 64U, uint8_t, void *);
 }
 
-/**
-  State allocation function when there is no key
-*/
-Hacl_Hash_Blake2b_state_t *Hacl_Hash_Blake2b_malloc(void)
+uint8_t *FStar_Pervasives_Native_snd__uint32_t__uint8_t_(K___uint32_t__uint8_t_ x)
```
R1kM, line 682: Nit: Could we use a struct instead of a pair to avoid these names ?

dist/gcc-compatible/Hacl_Hash_Blake2b.c
```diff
@@ -679,46 +679,138 @@ void Hacl_Hash_Blake2b_finish(uint32_t nn, uint8_t *output, uint64_t *hash)
...
+  uint32_t i = fst__uint32_t__uint64_t_____uint64_t_(block_state);
+  KRML_MAYBE_UNUSED_VAR(i);
+  uint32_t kk1 = key.fst;
+  uint8_t *key1 = key.snd;
+  if (!(kk1 == 0U))
+  {
+    uint8_t *sub_b = buf + kk1;
+    memset(sub_b, 0U, (128U - kk1) * sizeof (uint8_t));
+    memcpy(buf,
+      FStar_Pervasives_Native_snd__uint32_t__uint8_t_((
```
R1kM, line 758: Nit: Could this be inlined ?

R1kM (approved):

code/streaming/Hacl.Streaming.Blake2.Common.fst
msprotz: Yes because otherwise some lemmas cannot be proven for B.null (loc_in_footprint and stuff like that).

It's a massive pain and it's even worse in my upcoming PR but I don't have any way around that

msprotz: (Maybe it exists -- I just couldn't get it to work)

dist/gcc-compatible/Hacl_Hash_Blake2b.c
msprotz: yeah if it were a dedicated struct with projectors marked inline_for_extraction -- this struct is going to gain more fields in the upcoming PR so I can try to remove that in the final PR

karthikbhargavan (approved):
The changes look good, but I am afraid we may not have enough test vectors for all the cases. Is there a test-vector corpus somewhere we could use?

R1kM:
I could not find existing tests for parameters. The official Blake2 repo has many tests for blake and keyed blake (https://github.com/BLAKE2/BLAKE2/blob/master/testvectors/blake2-kat.json), but unfortunately does not seem to have anything for parameters (in addition to implementing the Blake2x version instead of the Blake2 version for parameters).

The tests I previously added to the spec were directly obtained from the Python implementation. I'm happy to generate more if needed.

dist/gcc-compatible/Hacl_Hash_Blake2b.c
msprotz: as discussed in PR #930, I cannot use structs because it doesn't monomorphize correctly, and I can't hide the names because of an incorrect handling of internal headers and CAbstractStruct in krml -- to be fixed in a followup

dist/gcc-compatible/Hacl_Hash_Blake2b.c
msprotz: this would require adding an alternate definition of fst and snd for HACL* that is marked inline_for_extraction... is it worth the trouble? perhaps

msprotz: or we submit an upstream PR on F*

msprotz:
Why don't we throw in a few more tests and then merge this? Thanks for your offer to help Aymeric :-)

R1kM:
I added 20 new autogenerated tests for parameters, using the current Python3 hashlib as an oracle.
The random generation creates random values for all parameters but the digest length (which is left to the default value),
and the key and key length (which is here left empty).

The script is easy to tweak, so I'm happy to add more/test specific patterns if needed.

msprotz:
@R1kM and I will pursue the final addition of `leaf_node` required for https://github.com/python/cpython/issues/99108 as separate PRs -- there isn't much left to do and we've worked out a plan of action (basically, extend the index with an additional boolean)

PR #928, opened 2024-03-29
Blake2, Part 2/5: Fix redundant (and incorrect) handling of parameters

msprotz:
Right now the spec overrides the key length and output length values stored in the parameters with extra arguments.

This PR fixes this, and comes on top of the other Blake2 PR.

Needs to be merged *after* #927


msprotz:
Probably needs hints or something.

R1kM (approved):

PR #930, opened 2024-04-02
Blake2, Part 3/5: Parameters, in the proper style, all the way to a decent top-level API

msprotz:
This incorporates #923 even though the code here is now in a much different form.

Some high-level summary of this new style:
- the index now covers what needs to be retained at run-time, namely a key length and an output length
- the key type class retains a full set of parameters, a key buffer -- but it's only used for the initialization functions (alloca, malloc, reset) that need to *receive* the key, then after that, the meta-parameter in the functor is set to `Erased`, meaning the key does not hold anything at runtime after that
- this leaves us with the blake2 state to hold important runtime information: the output length, and the key length; those cannot be changed from one instantiation of the algorithm to another

Compared to the previous style where the key length was an *argument* to the entire type class, this is a notable improvement:
- key length does not need to be known at compile-time
- no superfluous runtime argument update, digest, etc.

The proofs were unpleasant, to say the least, with many, many small things I had to tweak to make things work. (There are various comments -- for instance, `let x = y in t[x]` is not the same as `let x = y in t[y]` where `t`is a type that contains `x`or `y`; or, destructuring pairs in types `type t = fun (x, y) -> ...` seems to wreak havoc on unification and much better behavior is observed by using `fst` or `snd`. Anyhow, I've left some comments here and there, and the toplevel API takes care of not exposing the monomorphized pairs that are generated by the streaming functor (at the cost of my sanity, see proof of `malloc_with_key`).

What remains is:
- ~mostly documentation, making clear what are the assumptions on the part of the client behavior regarding these two lengths~
- ~perhaps reimplementing `reset` properly since the tests are written against that API~
- NOT DOING: proper struct for the index so that the type looks decent in the C header file (see comment below)


code/streaming/Hacl.Streaming.Blake2b_256.fst
```diff
@@ -37,49 +38,162 @@ let state_t (kk: G.erased (Common.key_size_t Spec.Blake2B)) =
...
+  let h1 = ST.get () in
+  assert F.(freeable blake2b_256 i h1 s);
+  assert (nn == (Spec.blake2_default_params Spec.Blake2B).digest_length);
+  calc (==) {
+    F.reveal_key blake2b_256 i h1 s;
+  (==) { }
+    blake2b_256.key.v i h0 (p, k);
+  (==) { }
+    Common.key_v i h0 (p, k);
+  (==) { _ by (FStar.Tactics.trefl ()) }
```
msprotz, line 133: like, why?? no clue

code/streaming/Hacl.Streaming.Blake2.Params.fst
```diff
@@ -0,0 +1,101 @@
...
+open Lib.Buffer
+open LowStar.BufferOps
+
+module B = LowStar.Buffer
+module ST = FStar.HyperStack.ST
+module HS = FStar.HyperStack
+
+let _align = ()
+
+// MUST BE A SEPARATE TYPE ABBREVIATION
```
msprotz, line 13: puzzling

code/blake2/Hacl.Impl.Blake2.Core.fsti
```diff
@@ -99,20 +99,16 @@ let blake2_params_v (#a: Spec.alg) (h: mem) (p: blake2_params a): GTot (Spec.bla
...
-      as_seq h salt == Seq.create (Spec.salt_length a) (u8 0) /\
-      as_seq h personal == Seq.create (Spec.personal_length a) (u8 0)
-    )
-    (ensures (fun h0 p h1 ->
-      h0 == h1 /\
-      blake2_params_loc p == loc salt `union` loc personal /\
-      blake2_params_inv h1 p /\
-      blake2_params_v h1 p == Spec.blake2_default_params a))
+inline_for_extraction noextract
+val alloca_default_params: a: Spec.alg -> StackInline (blake2_params a)
```
R1kM, line 103: I can't remember the precise issue I had in the previous versions that prevented me from doing this, but I'm glad to see the extra indirection gone.

code/blake2/Hacl.Impl.Blake2.Generic.fst
```diff
@@ -1002,6 +1022,54 @@ let blake2_update #al #ms blake2_update_key blake2_update_blocks
       else blake2_update_blocks wv hash lb d)
     else blake2_update_blocks wv hash (size_to_limb al 0ul) d
 
+// TODO: this is technically redundant because output_len and key_len are ALREADY present in params. But this is good for regularity... or not?
```
R1kM, line 1025: I'm a bit worried this will lead to mistakes, as this add extra preconditions to sync params with lengths.
I'd be in favor of expressing length constraints based on params.

code/streaming/Hacl.Streaming.Blake2.Common.fst
```diff
@@ -149,43 +149,158 @@ let key_size (a : alg) = kk:nat{kk <= Spec.max_key a}
...
+        let h1 = ST.get () in
+        assert (key_footprint #a #i h1 s == P.footprint h1 p);
+        s
+      else
+        let h0 = ST.get () in
+        let p = P.alloca a i in
+        let h00 = ST.get () in
+        assert P.(invariant h00 p);
+        let h01 = ST.get () in
+        assert P.(invariant h01 p);
```
R1kM, line 235: Duplicated assert

code/streaming/Hacl.Streaming.Blake2.Params.fst
```diff
@@ -0,0 +1,101 @@
...
+// nothing works if inlined underneath B.pointer, below
+let blake2_params a (i: index a) =
```
R1kM, line 15: Needs an inline_for_extraction noextract?

R1kM (approved):

code/streaming/Hacl.Streaming.Blake2.Params.fst
msprotz: sure

msprotz:
https://github.com/FStarLang/karamel/pull/421 contains the build fix but is stuck on an F* build failure, to be diagnosed

msprotz:
last build failure was a verification error in lib.buffer (which incidentally also failed on my machine, and I had to locally bump the rlimit)

msprotz:
Ok we're down to silly errors in tests. I'll fix them.

msprotz:
🕺

karthikbhargavan (approved):
Wow, a monumental effort here!

The changes to the top-level API look fine to me,
time and experiments will tell if people expect something different.

PR #931, opened 2024-04-09
Add magic nix command because I can't be searching in my bash history every time

msprotz:


msprotz:
Hmm looks like this is not how HACL* works. Closing.

PR #936, opened 2024-04-18
Blake2, 5/5: Extend implementation to support last node

R1kM:
Following #934, this PR extends the implementation of Blake2 to support the last_node argument.
To do so, it extends the streaming functor index with a last_node boolean, which is set at initialization time.

To minimize API changes, this flag is only exposed in the malloc_with_params_and_key function, in particular, keyed hashes without parameters set last_node to False by default (as this flag only makes sense in tree-hashing mode, which often requires additional parameters).

The implementation required to extend the update_last function with an additional boolean argument for last_node. This function is not intended to be used directly, it is recommended to use hashes through the streaming functor.


msprotz (approved):
Looks good to me, with a comment to explain that the reset API cannot change whether this is the last node or not.

I'll try an integration into Python shortly.

karthikbhargavan (approved):
I am not sure I understood all the changes, the they look minimal and good.

PR #940, opened 2024-04-24
Fix header inclusion logic and bump XCode version following GitHub Actions OSX image upgrade

msprotz:
This fixes a recent regression following the merge of SHA3/SIMD256.

With this PR, SHA3 depends on Hacl_SHA2_Types.h, since the latter file contains general-purpose definitions (e.g., a quadruple of pointers) that are now used by both SHA2 and SHA3, since SHA3 is now SIMD too.

I could rename Hacl_SHA2_Types.h into Hacl_Multibuf_Types.h or something, but this might break some clients like Python, so I'm not sure how judicious that would be.

In the meanwhile, that fixes the issue reported by Franziskus.


code/sha3/Hacl.Hash.SHA3.Simd256.fst
```diff
@@ -27,6 +27,17 @@ module M = LowStar.Modifies
 
 #reset-options "--z3rlimit 50 --max_fuel 0 --max_ifuel 0 --using_facts_from '* -FStar.Seq'"
 
+val absorb_inner_256:
+    rateInBytes:size_t{v rateInBytes > 0 /\ v rateInBytes <= 200}
+  -> b: Hacl.Impl.SHA2.Types.bufx4
```
msprotz, line 32: this it the important bit that encodes that SHA3.Simd256 now depends on the types (and pretty names) contains in that file -- this instructs krml to use those pretty names when monomorphizing types.

msprotz:
xcode-select: error: invalid developer directory '/Applications/Xcode_13.4.app'

msprotz:
@pnmadelaine @W95Psp would any of you know what's going on with github actions on osx? the latest error seems pretty cryptic

msprotz:
https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md looks like the version of XCode that we select is no longer supported

R1kM (approved):

franziskuskiefer:
Thanks. I can confirm that this fixes the issue locally for me.
hacl-packages CI is failing on this because it needs changes in tests (the API changed).
Here's a draft with fixes https://github.com/cryspen/hacl-packages/pull/461

franziskuskiefer (approved):
I presume the msvc version gets updated accordingly (it is not in here).

Then this looks good. There are some failures in the bindings that need fixing (and msvc doesn't go through obviously), but otherwise CI is happy.

msprotz:
Let me kick another dist refresh

PR #941, opened 2024-04-25
Various refreshes

msprotz:
Not sure what's happening with the build. Trying again to get a green.


msprotz:
The Makefile was incorrect and was passing `-no-prefix Hacl.Test.HMAC.DRBG` instead of `-no-prefix Hacl.Test.HMAC_DRBG`. Previously, owing to another bug in krml, this was ok, but now that I've cleaned things up and remove the krml bug, the HACL Makefile bug started showing.

R1kM (approved):

karthikbhargavan (approved):

PR #948, opened 2024-05-06
Update OCaml builds in Makefile

R1kM:
Following @mtzguido 's suggestion, add `-thread` to the OCAMLOPT/SHARED aliases in Makefiles.
This should also remove a recent warning during builds.


R1kM:
This is blocked on other fixes to HACL* proofs. Closing, to integrate it in existing PRs.

Issue #954, opened 2024-06-14
SHA3 Performance regression


franziskuskiefer:
The new SHA3 code from #897 (avx2 and scalar) lost about 30% performance just before merge.
We have been using the version before merge for a while and noticed this performance regression on updating hacl-packages.

```
Benchmark                    Time             CPU      Time Old      Time New       CPU Old       CPU New
---------------------------------------------------------------------------------------------------------
Hacl_Sha3_224             +0.2821         +0.2819          2792          3580          2792          3580
Hacl_Sha3_256             +0.2699         +0.2698          3213          4081          3213          4080
Hacl_Sha3_384             +0.2709         +0.2709          3992          5073          3991          5073
Hacl_Sha3_512             +0.2816         +0.2817          5698          7302          5697          7302
Hacl_Sha3_256_Streaming   +0.3030         +0.3030          3435          4475          3435          4475
Hacl_Sha3_shake128        +0.2827         +0.2827          2426          3111          2425          3111
Hacl_Sha3_shake256        +0.2890         +0.2891          3216          4145          3215          4145
```

msprotz:
Can @mamonet investigate? Thanks for pinpointing this!

karthikbhargavan:
I addressed this issue in PR #955 

For some reason, there is a regression in the generation of nice C types, and there is a question on the tradeoff between code size and performance, but the immediate perf issue is addressed.

PR #955, opened 2024-06-15
Restore SHA-3 speed by inlining absorb_inner

karthikbhargavan:
## Proposed changes

This PR fixes Issue #954 by restoring the `inline_for_extraction` flags for `absorb_inner`.
The trade-off is that the resulting C code is larger because the code in this function gets copied multiple times.

An alternative fix would be to introduce a KaRaMeL attribute that inserted `__attribute((always_inline))` (and equivalent attributes for different compilers) into the generated C code. That would allow us to generate compact code while keeping
the good performance.

## Types of changes

- [ X ] Bugfix (non-breaking change which fixes an issue)
- [ X ] Improved performance (fix or feature that would improve performance of some algorithm on some platform)

## Checklist

- [ X ] I have read the [CODE_OF_CONDUCT.md](https://github.com/hacl-star/hacl-star/blob/main/CODE_OF_CONDUCT.md) doc
- [ X ] I have read and agree to submit my changes under the [LICENSE](https://github.com/hacl-star/hacl-star/blob/main/LICENSE)


dist/gcc-compatible/Hacl_SHA2_Types.h
```diff
@@ -35,30 +35,17 @@ extern "C" {
 #include "krml/lowstar_endianness.h"
 #include "krml/internal/target.h"
 
-typedef struct Hacl_Hash_SHA2_uint8_2p_s
```
karthikbhargavan, line 38: I don't know why these typedefs changed in the current extraction. 
It seems like a regression since it results in bad type names in the C code.

msprotz: yeah this is blocking because now sha2 depends on sha3 (which is *not* what we want)

these types are poorly named -- they should be named "generic types" or something but let's leave that aside for now

the way to fix this is to figure out *where* in Hacl_SHA3 pairs of pointers to uint8 are used, and have those pairs be references to Hacl.SHA2.Types instead of un-named types (like `buffer uint8 & buffer uint8`)

msprotz (changes requested):
Let's try to fix that issue with type names, because right now we can't merge this as it introduces sideways dependencies from sha2 to sha3

msprotz:
Concerning your remark about inline: we could make this the behavior of CInline by default, since we tend to always want something marked as inline to be always inlined. But we'd have to do careful performance regressions to make sure this is beneficial.

The way to try this is simply to modify `lib/PrintC.ml` to print `krml_inline` instead of `inline` and then define the macro accordingly in target.h or something

PR #956, opened 2024-06-20
Prettify Blake2 dist

R1kM:
Minor change to streaming Blake2: do not mark `stateful_key_t` as inline_for_extraction, and rename it into `params_and_key`.

This does not modify the API, but removes an autogenerated monomorphized type, and also helps with Rust extraction.


msprotz (approved):
Nice

PR #969, opened 2024-08-21
Simplify a bignum mask

R1kM:
The relevant changes are in Hacl.Bignum.Comparison.fst.
The computation of the lt mask on bignums was performing an additional masking step, which is actually not needed.
This PR removes it, along with a proof that it is equivalent to the previous behavior and the refresh of the generated C code.


msprotz (approved):
I presume this PR was prompted by a clippy warning?

PR #974, opened 2024-08-30
Inline init_with_params in Blake2

R1kM:
This PR inlines the calls to `init_with_params` occurring in `malloc_raw` and `reset_raw` in the streaming Blake2 code.
This only impacts internal code, and does not modify user-facing APIs.

This change is needed for HACL-rs: the previous pattern was leading to a move-out, as the Blake2 parameters don't have the copy trait, and were passed as a value to `init_with_params`.
Modifying this function to pass it through a pointer is a much more involved change (if possible due to refinement type mismatches in the streaming functor), which would also impact the user-facing API, hence this smaller change.


msprotz (approved):
If that's what makes hacl-rs work... thanks!

PR #980, opened 2024-09-13
specify inline related attribute for Oracle Studio compiler

vladak:
## Proposed changes

This change avoids warning/error when compiling with Oracle Studio by adding appropriate attributes for the inline enforcement. The `__SUNPRO_C` internal define  is documented in the Studio compiler man page for the `cc` command. The attribute values themselves are documented in https://docs.oracle.com/cd/E77782_01/pdf/E77788.pdf (section 2.10)

## Types of changes

## Checklist

- [x]  I have read the [CODE_OF_CONDUCT.md](https://github.com/hacl-star/hacl-star/blob/main/CODE_OF_CONDUCT.md) doc

## Further comments

While the `__SUNPRO_C` define has a distinct value for every version, for this change I am assuming fairly modern version, i.e. 12.x, so the value is not checked. Let me know if that's okay.

### DIFF
--- a/dist/karamel/include/krml/internal/target.h
+++ b/dist/karamel/include/krml/internal/target.h
@@ -82,6 +82,8 @@
 #    define KRML_NOINLINE __declspec(noinline)
 #  elif defined (__GNUC__)
 #    define KRML_NOINLINE __attribute__((noinline,unused))
+#  elif defined (__SUNPRO_C)
+#    define KRML_NOINLINE __attribute__((noinline))
 #  else
 #    define KRML_NOINLINE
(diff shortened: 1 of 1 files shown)


msprotz:
Super, thanks for the PR. This file lives upstream at https://github.com/FStarLang/karamel/blob/master/include/krml/internal/target.h do you mind submitting a PR there? The update will then trickle down within a week to HACL*.

Out of curiosity, what is your use-case?

vladak:
Okay, created https://github.com/FStarLang/karamel/pull/484

My use case is compiling ECC code from Mozilla NSS on Solaris 11.4.

PR #991, opened 2024-10-11
Make Intel intrinsincs available even on x86

msprotz:
The vectorized operations are available even when compiling for x86 mode. See https://github.com/python/cpython/pull/125244 for some context, and also this was reported privately by @samuel-lee-msft a while back (thanks)


R1kM (approved):
Thanks!

PR #998, opened 2024-10-18
Makefiles/scons: using fstar.exe --ocamlopt to simplify, and -thread for libhaclml.cmxa

mtzguido:
This PR makes use of F*'s --ocamlenv and --ocamlopt to simplify some of the build. The idea is that most clients can avoid remembering the some of the needed flags and the name of the library by just using these modes of fstar.exe. However see the discussion [here](https://fstar.zulipchat.com/#narrow/channel/458710-compiler-devel/topic/RFC.3A.20fstar.2Eexe.20--ocamlopt.2C.20etc), I wonder what your thoughts are.

This also passes -thread to the building of libhaclml.cmxa, which I think is needed (there's a warning about it).

## Types of changes

- [x] Build system cleanup

## Checklist

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1026,7 +1026,7 @@ obj/libhaclml.cmxa: $(filter-out $(HACL_HOME)/obj/Meta_Interface.cmx,$(ALL_CMX_F
 # ocamlfind mklib -o haclml -package fstar.lib -g -I $(HACL_HOME)/obj $(addprefix $(HACL_HOME)/obj/*.,cmo cmx ml o)
 # NB: We cannot use fstar.exe --ocamlopt here as it passes -linkpkg. (Should it not?)
 	$(call run-with-log,\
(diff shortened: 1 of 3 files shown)


msprotz:
My intuition is that I'm wary of taking a dependency on F*, and I'm afraid that I'll eventually end up debugging F*'s logic for building *and* OCaml's build logic should anything go haywire. The hesitation around linkpkg and -a also leads me to think that for anything but the most basic examples, people should just drive OCaml directly.

But this is my personal hunch (I remember basically having to reverse-engineer a lot of ocamlbuild's logic a long while ago, and in the end the purported layer of abstraction was more of a pain than a benefit).

Perhaps this would be mitigated if F* always printed out the invocation it shells out. Does it do that currently?

Curious to hear what @R1kM and also @twal think.

R1kM:
I tend to agree, I'm not sure how much the additional indirection brings, apart from shortening the commands

PR #1003, opened 2024-11-04
Remove KRML_CHECK_SIZE in HMAC

msprotz:
In support of https://github.com/python/cpython/pull/126359


msprotz:
Nix seems to be failing. Can one of our Nix experts look at this? @nadrieril @W95Psp? Thanks

msprotz:
`fstar> sh: line 1: cd: ../../mk: No such file or directory`

looks like something changed on the F* side ... @tahina-pro @mtzguido did something change in the build of HACL* that would require updating its Nix file?

mtzguido:
Ah, sorry, I broke it here https://github.com/FStarLang/FStar/pull/3610. I didn't realize the Nix build failed! Let me take a look.

mtzguido:
Can you retry?

msprotz:
Thanks for helping out Guido! I re-ran but it doesn't seem different, so I'm not sure what's up

mtzguido:
Hm.. I'm not a nix expert at all, but this same line
```
these 3 derivations will be built:
  /nix/store/9cgmfp2ldx7ghmf5c7mwvs4irs8658bc-fstar-2e5c4a48bd3c9af15dd323558fcfbfb3d2225d5d.drv
  [...]
```
appears in all runs, before and after F* was updated, but the fact that this hash does not change makes me think it's picking up the version before the fix in some cache.

Wild idea, if you `git commit --amend` and force-push here, or add an empty commit, that would trigger a run from scratch, and maybe fix this? It seems that locally I can start a nix build on this branch and go further than CI gets to.

R1kM (approved):

msprotz:
Looks like it worked, thanks Guido!

PR #1007, opened 2024-12-03
Incremental (streaming) HMAC

msprotz:
In support of https://github.com/python/cpython/pull/126359, this is an implementation of streaming (incremental) HMAC for all known fixed-length hash algorithms in HACL*.

This went relatively smoothly (insofar as wrestling with Low* goes), with only a tiny modification to the streaming functor.

This is a work in progress, and I still want to clean up the toplevel API, but this verifies, extracts, and compiles.


msprotz:
@picnixz can you take a look at the API? This is not the final form, but it should allow you to confirm this is headed in the right direction. What I have on my list is:
- the key length is repeated between the index and key_and_len types -- the final malloc function will look more "C-native", and will take implementation, key, and key length, and will build the two structs under the hood
- the `malloc` function right now assumes that you never screw up and only ever pass, e.g., Blake2b_256 on platforms that *do* support it -- the final function will be defensive and will return NULL in case you pass something inconsistent
- the C files depend on two defines that need to be computed at build-configure time -- the Python integration will have to mimic what it does already for blake2 and define those two macros via build-time compile flags
- the `reset` function only works if the key length is identical -- just like `malloc`, this will be wrapped in a defensive function that may return an error code if the reset operation failed

There's also a `false_elim` function somewhere that will be replaced by a KRML_ERROR once I get to it.

msprotz:
APIs for `malloc` and `reset` were fixed.

msprotz:
The digest function was missing! It's now in there, with the caveat that the implementation of `digest` has a temporary allocation declared on the stack for each possible choice of algorithm. I don't think it's the end of the world, but if it's a dealbreaker, we can have a variant that heap-allocates the temporary state and thus won't generate this kind of code (or we do a peephole optimization that introduces a true C union on the basis that the allocations all originate from disjoint branches of the code).

dist/gcc-compatible/Hacl_Streaming_HMAC.h
```diff
@@ -0,0 +1,119 @@
...
+#include "Hacl_Streaming_Types.h"
+#include "Hacl_Krmllib.h"
+#include "Hacl_Hash_SHA3.h"
+#include "Hacl_Hash_SHA2.h"
+#include "Hacl_Hash_Blake2s_Simd128.h"
+#include "Hacl_Hash_Blake2s.h"
+#include "Hacl_Hash_Blake2b_Simd256.h"
+#include "Hacl_Hash_Blake2b.h"
+
+#define Hacl_Agile_Hash_MD5 0
```
picnixz, line 47: Out of curiosity, why not using an enumeration instead?

dist/gcc-compatible/Hacl_Streaming_HMAC.h
```diff
@@ -0,0 +1,119 @@
...
+typedef struct Hacl_Streaming_HMAC_Definitions_index_s
+{
+  Hacl_Agile_Hash_impl fst;
+  uint32_t snd;
+}
+Hacl_Streaming_HMAC_Definitions_index;
+
+typedef struct Hacl_Streaming_HMAC_Definitions_key_and_len_s
+{
+  uint8_t *fst;
```
picnixz, line 75: Quick implementation: do you keep the key in clear here? Since HMAC does

```
H(K' ^ opad || H(K' ^ ipad || text))
```

and the underlying hash function admits an update routine, can't you just store the state of the inner and outer layers instead?

picnixz (commented):
Thanks for the initial work!

What I'm lacking a way to clone the state without affecting the input state (not just resetting it). Namely, have a safe `memcpy` that computes the correct lengths etc (I can do it on my side but I'd prefer having it in HACL*). A bit like `Hacl_Hash_MD5_copy`.

dist/gcc-compatible/Hacl_Streaming_HMAC.h
```diff
@@ -0,0 +1,119 @@
...
+
+Hacl_Streaming_Types_error_code
+Hacl_Streaming_HMAC_update(
+  Hacl_Streaming_HMAC_agile_state *state,
+  uint8_t *chunk,
+  uint32_t chunk_len
+);
+
+void
+Hacl_Streaming_HMAC_digest(Hacl_Streaming_HMAC_agile_state *state, uint8_t *output, uint32_t l);
```
picnixz, line 110:
```suggestion
Hacl_Streaming_HMAC_digest(Hacl_Streaming_HMAC_agile_state *state, uint8_t *output, uint32_t digest_size);
```

dist/gcc-compatible/Hacl_Streaming_HMAC.h
msprotz: yes, I could, it would just require a little bit more work (and I'm off next week) -- do you think this is a blocker?

picnixz: And take your time. We have until May for it to be included in 3.14 and if it does not land in 3.14, it'll land in 3.15.

picnixz: Let's ask someone from CPython as well: @gpshead In the Python implementation of HMAC, we drop the key and only keep inner and outer states (well, we just keep the hash objects directly). Do you think this should be the same here ?

dist/gcc-compatible/Hacl_Streaming_HMAC.h
msprotz: enums in C11 are of type int and this might use up more space in structs -- the default behavior is thus to use uint8 and macros

there's an option to generate C code that uses enums (remember that this is all generated C code from the verified source), and I think C23 allows defining narrower enum types, so I could improve the code-gen to use enum *and* use a narrow type, but Python only assumes C11, not C23, so if you wanted enums, those would end up being int

I could do that for HACL* as it might look better, and would most likely not break any clients... @R1kM @karthikbhargavan thoughts?

dist/gcc-compatible/Hacl_Streaming_HMAC.h
msprotz: fix forthcoming (using length as a suffix for consistency with the rest of the codebase)

dist/gcc-compatible/Hacl_Streaming_HMAC.h
picnixz: > C23 allows defining narrower enum 

I thought it was allowed earlier but it's only for C++, my bad. So, I'm ok with those macros (since I can just use them anyway as if they were enumerated fields).

dist/gcc-compatible/Hacl_Streaming_HMAC.h
msprotz: I didn't realize I could perform that optimization, but I agree it's a nice one to have. The optimization would be a little bit more of verification work, I think, but not dramatically so, since most of the groundwork is already laid out. You would just have to incorporate into the invariant that the second hash state, upon initialization, has been fed only the key, and remains unchanged into the call to `digest`, upon which is receives the first hash.

dist/gcc-compatible/Hacl_Streaming_HMAC.h
msprotz: 👍

msprotz:
The lack of copy was just an omission on my part, I just pushed it along with the renaming of `l` into `digest_length`. Thanks for the prompt feedback!

dist/gcc-compatible/Hacl_Streaming_HMAC.c
```diff
@@ -0,0 +1,2204 @@
...
+  if (scrut.tag == SHA2_512_s)
+  {
+    uint64_t *p1 = scrut.case_SHA2_512_s;
+    Hacl_Hash_SHA2_sha512_init(p1);
+    return;
+  }
+  if (scrut.tag == SHA3_224_s)
+  {
+    uint64_t *p1 = scrut.case_SHA3_224_s;
+    memset(p1, 0U, 25U * sizeof (uint64_t));
```
picnixz, line 381: I can only comment on auto-generated code since I don't know how to read F*, but I would have expected `Hacl_Hash_SHA3_sha224_init` (same for other SHA3 branches) here even if it only does a 0 memset.

dist/gcc-compatible/Hacl_Streaming_HMAC.c
```diff
@@ -0,0 +1,2204 @@
...
+    uint64_t *p1 = scrut.case_SHA3_224_s;
+    uint32_t remOut = 28U;
+    uint8_t hbuf[256U] = { 0U };
+    uint64_t ws[32U] = { 0U };
+    memcpy(ws, p1, 25U * sizeof (uint64_t));
+    for (uint32_t i = 0U; i < 32U; i++)
+    {
+      store64_le(hbuf + i * 8U, ws[i]);
+    }
+    memcpy(dst + 28U - remOut, hbuf, remOut * sizeof (uint8_t));
```
picnixz, line 755: Ditto about the lack of `Hacl_Hash_SHA3_sha512_finish` for this one.

dist/gcc-compatible/Hacl_Streaming_HMAC.c
```diff
@@ -0,0 +1,2204 @@
...
+    uint32_t *p_src = scrut.case_MD5_s;
+    Hacl_Agile_Hash_state_s x1 = *s_dst;
+    uint32_t *p_dst;
+    if (x1.tag == MD5_s)
+    {
+      p_dst = x1.case_MD5_s;
+    }
+    else
+    {
+      p_dst = KRML_EABORT(uint32_t *, "unreachable (pattern matches are exhaustive in F*)");
```
picnixz, line 939: Out of curiosity, but does the HACL* design require to always abort and not to return an error code in case of such failures?

msprotz: In the verified code, there is a precondition that says that the caller cannot mix-and-match for the copy function, and can only copy between states that have identical choices of implementation and key length. This means that from the point of view of the verified code, this simply cannot happen.

In reality, however, callers are in C, and the C compiler does not check that precondition, so this line of code could still be reachable. All we have is a comment (TODO) that says don't do that. So the code generation adds this line so that i) there is an actionable error rather than a segfault that requires firing up the debugger, and ii) to avoid compiler warnings along the lines of "this variable is possibly unset".

I could change the verified so that the code first checks at runtime that the choice of implementation and key lengths match, and returns an error otherwise -- as opposed to a precondition that the C code might miss. Thoughts?

dist/gcc-compatible/Hacl_Streaming_HMAC.c
msprotz: Worth fixing, good catch!

dist/gcc-compatible/Hacl_Streaming_HMAC.c
msprotz: Same, also worth fixing. Will do.

dist/gcc-compatible/Hacl_Streaming_HMAC.h
```diff
@@ -0,0 +1,126 @@
...
+#endif
+
+#include <string.h>
+#include "krml/internal/types.h"
+#include "krml/lowstar_endianness.h"
+#include "krml/internal/target.h"
+
+#include "Hacl_Streaming_Types.h"
+#include "Hacl_Krmllib.h"
+#include "Hacl_Hash_SHA3.h"
```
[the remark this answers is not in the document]
msprotz: it's only included in the .c, not the .h -- not sure why, I'll take a look,

dist/gcc-compatible/Hacl_Streaming_HMAC.c
msprotz: Looking at it more closely, my explanation above is wrong, because the code returns a fresh state for the copy, and this is a union selection, so the code won't pick the "wrong branch". (I though the copy function took both source and destination pointers.)

In that case, yes, this could be optimized too.

msprotz: (Note that this is an internal copy function, and that the user-facing Hacl_Streaming_HMAC_copy function returns a fresh state.)

picnixz: Is there a way for `Hacl_Streaming_HMAC_copy` to fail and return NULL without triggering a crash?

picnixz: > So unless I'm missing something, Hacl_Streaming_HMAC_copy cannot trigger an error

For instance, what happens if malloc() returns NULL? (for instance, we're out of memory). We do check when malloc() returns NULL on our side, but I wondered whether you considered this.

msprotz: Right, that we generally do not check for in HACL*. We could, if deeemed important. Let us know!

picnixz: I personally would like to be able to raise a `PyErr_NoMemory()` which would anyway lead to a MemoryError (and generally an interpreter abort but not always I think). So I would prefer raising that exception if possible (and it would make sense that the allocation function behaves like a dedicated malloc() with the same semantics, namely returning "NULL" = no memory or something else happened).

If it's too hard or could break things on your side, it's also fine. (Probably, it'd be good to get Gregory's opinion on this matter as well).

dist/gcc-compatible/Hacl_Streaming_HMAC.h
```diff
@@ -0,0 +1,126 @@
...
+#include "Hacl_Hash_SHA2.h"
+#include "Hacl_Hash_Blake2s_Simd128.h"
```
picnixz, line 42: I'm currently upgrading my CPython PR, but does the SIMD includes be guarded behind `#if HACL_CAN_COMPILE_VEC{128,256}`?

dist/gcc-compatible/Hacl_Streaming_HMAC.h
```diff
@@ -0,0 +1,126 @@
...
+#define Hacl_Agile_Hash_SHA2_224 2
+#define Hacl_Agile_Hash_SHA2_256 3
+#define Hacl_Agile_Hash_SHA2_384 4
+#define Hacl_Agile_Hash_SHA2_512 5
+#define Hacl_Agile_Hash_SHA3_224 6
+#define Hacl_Agile_Hash_SHA3_256 7
+#define Hacl_Agile_Hash_SHA3_384 8
+#define Hacl_Agile_Hash_SHA3_512 9
+#define Hacl_Agile_Hash_Blake2S_32 10
+#define Hacl_Agile_Hash_Blake2S_128 11
```
picnixz, line 58: Ditto

picnixz:
FTR, I've finished implementing HMAC on Python's side: https://github.com/picnixz/cpython/tree/feat/hmac/hacl-99108. You can have a look at each commit as they are really independent. However, I've got some questions:

- Currently, the one-shot HMAC interface only accepts messages and keys whose length fits on 32 bits. Is there a plan to have one-shot **non-agile** generic interface for messages with larger lengths?
- Is there any plan to support keys with length exceeding 32 bits? (currently, we just raise an error but IIRC, OpenSSL takes a size_t for the key length).

Realistically speaking, I don't see in which universe I will use a key that has size > 2^32... so I don't need that. What I need more is one-shot HMAC for messages of length > 2^32 but I'm also fine with not having this if it's too complex.

picnixz:
Just for those subscribed to this pull request. I plan to re-open a fresh one using https://github.com/picnixz/cpython/tree/feat/hmac/hacl-99108 in order to incorporate both streaming and 1-shot HMAC at the same time. However, we can continue the design discussion on this PR until the HACL* part is done.

msprotz:
Thanks for the update, on my end, I plan to implement the two-hash-state solution, along with a variant of malloc that can return NULL to indicate an allocation failure.

We currently do not have an easy way to support hashing sizes > 2^32 (it's a long-standing limitation of HACL* that would require a substantial rewrite), but this can be lifted by writing a small wrapper in C that stack-allocates the correct state (by hand), then calls init / update / digest via the agile incremental API. Would that work?

picnixz:
> but this can be lifted by writing a small wrapper in C that stack-allocates the correct state (by hand), then calls init / update / digest via the agile incremental API

On Python's end, I can live with a simple solution where, if either the message or the key is too long, then I fall back to the agile API instead. It'll be a bit slower since we need to construct the object on our side, but the cost will likely be aborbed by the actual HMAC computation.

By the way, [my previous comment](https://github.com/hacl-star/hacl-star/pull/1007#issuecomment-2545690107) was on the wrong PR. I wanted to add this comment on the CPython's PR not this one :')

msprotz:
> if either the message or the key is too long, then I fall back to the agile API instead

yes, the cost of allocating/freeing will be negligible

I just want to be clear that we have no solution for a key that might be > 2^32, so if this is a must-have, let me know and we'll start looking into "upgrading" enough of the hashes that they can deal with size_t (and not uint32) up to the hmac key

picnixz:
> so if this is a must-have

Honestly, in practice I don't see in which universe you'll use a key with > 2^32 bits (realistically, it's likely that you swapped the message and the key when ushowing the interface). So, unless a downstream user is crying "hey my key is very large and I need HMAC on this", I think we can live without it. It's entirely up to you since you're the one doing the maintenance in the end.

dist/gcc-compatible/Hacl_Streaming_HMAC.c
msprotz: I pushed an update that makes `malloc` now return NULL if any of the underlying malloc calls failed.

Note that for the HMAC API specifically, the NULL return value is also returned in case of an illegal combination of implementation and choice of build target (e.g. 256-bit vectorized implementation on Apple Silicon). You can easily distinguish the two if need be as a caller, but if this is a blocker, I can tweak the API to be of the form:

```c
Hacl_Streaming_Error_Code
Hacl_Streaming_HMAC_malloc(Hacl_Streaming_HMAC_agile_state **dst, Hacl_Agile_Hash_impl impl, uint8_t *key, uint32_t key_length);
```

instead of:

```c
Hacl_Streaming_HMAC_agile_state
*Hacl_Streaming_HMAC_malloc(Hacl_Agile_Hash_impl impl, uint8_t *key, uint32_t key_length);
```

let me know

msprotz:
@R1kM I changed the `stateful` definition to return an `option` for the `create_in` function. This makes the generated C uglier, but I'm not sure I care deeply?

The reason `create_in` cannot use `NULL` to indicate allocation failure is that `stateful` does not necessarily rely on the memory -- it might be a value without any underlying mutable allocation (and we *do* leverage this in a couple places).

The alternative design is to tweak `stateful` as follows:
- add an `allocation_success` predicate to `stateful`
- add a run-time function `allocation_succeeded` that takes a piece of state and returns `either unit (squash (allocation_success s))`
- `alloca` always establishes `stateful` in its post
- `create_in` changes its post to be of the form `if allocation_success s then ... else modifies_none ...`
- instances of `stateful` without allocation failure define those predicates to be `True` and `true` respectively

Let me know what you think, I believe the alternative design would be a little more complicated (though not significantly) and would generate more idiomatic code.

dist/gcc-compatible/Hacl_Streaming_HMAC.c
picnixz: > the NULL return value is also returned in case of an illegal combination of implementation and choice of build target (e.g. 256-bit vectorized implementation on Apple Silicon). 

If the work can be done on your side automatically, I would be happy, so if a status code is returned, that would be perfect. SO thanks in advance.

R1kM:
I'm fine with the `option`, I'm a bit wary of the complexity and stability of proofs with the alternatives

dist/gcc-compatible/Hacl_Streaming_HMAC.c
msprotz: no worries, that's done!

msprotz:
Ok the key is no longer kept at runtime.

We still have one issue which is that mmalloc_partial is unsound (CC @tahina-pro) so we need to remove that function and expose mmalloc_partial_uninitialized. See https://github.com/hacl-star/hacl-star/pull/1007/files#diff-6575620b2a57a4b1ceed2b0a54e8d51f8c4e29225d5e7ac991568dcf8ad251b3R230-R231 -- I'll say more on Monday 

I'll do another round later with the more minor fixes related to enums and excessive inlining.

msprotz:
Ok so the problem is that `malloc` is modeled in Low* as taking an initial value, to be filled for every index of the array. When the initial value is zero, we use `calloc` -- this is fine. But if the initial value is something else, we either perform an assignment (array length 1) or a for-loop based initializer (unknown array length), and then the null-check that the user wrote comes *after* that.

My proposed fix is in https://github.com/FStarLang/karamel/pull/513 and I just pushed a refresh C package that incorporates this fix.

dist/gcc-compatible/Hacl_Streaming_HMAC.h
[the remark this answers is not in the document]
msprotz: fixed in 488c4b1e8f via https://github.com/FStarLang/karamel/pull/525

dist/gcc-compatible/Hacl_Streaming_HMAC.h
msprotz: resolved per discussion above regarding enums (I believe)

dist/gcc-compatible/Hacl_Streaming_HMAC.c
msprotz: fixed via 0cabac04d9

dist/gcc-compatible/Hacl_Streaming_HMAC.c
msprotz: fixed via 0cabac04d9

msprotz:
hey @picnixz I think I've fixed all of your comments and this should be in good shape -- when you get a chance, can you tell me whether this looks good for python? if so, I'll work on fixing the last few tidbits to get a green, then merge... thanks!

picnixz:
Thank you! I am currently travelling and will be back next week. I have some work for the upcoming CRYPTO deadline but I'll definitely look at it after the 15th, if not before.

msprotz:
Just talked about this with @R1kM and the plan is to merge this so that we have consistent null-checking across all those APIs now, then we can do follow-up touch-ups as you see fit if you have further tweaks requested. Cheers.

R1kM (approved):

msprotz:
The build issue is that there is now an incorrect dependency from SHA1 towards Blake2, most likely owing to type monomorphizations. Looking into it.

msprotz:
@R1kM this turned out to be a lot more complicated than I thought: the fact that the streaming functor now relies on the option type means that various monomorphizations of `option` were inserted at the first use-site, leading to the tags (None, Some) being inserted in whichever file used it first (in this case, Blake2) and subsequent algorithms (e.g. SHA2) #include-ing the header for Blake2 just to get the tags None and Some in scope.

What I ended up doing is lifting some of the state type definitions (which, upon being visited, would generate monomorphizations of option) into a separate file, and grouping all type definitions into one file.

The net result is that all types now live in Hacl_Streaming_Types.h, including state types that contain pointers to arrays of vec128/vec256. This still compiles, because libintvector.h is always safe to include (vec256 becomes void* on platforms that don't support it, precisely for this use-case), but it's not ideal. Worse, because I can no longer eliminate unused declarations from this module (because their presence now has a side-effect), I get a C file with the discriminators (is_UU...).

But @mtzguido has given me a secret attribute to remove that last bit so hopefully I can fix that.

PR #1017, opened 2025-01-06
Updates for new F* build and packaging

mtzguido:
## Proposed changes

This PR prepares HACL* for the new F* build and layout. The last patch is for the change in F* layout:
- `fstar.exe --locate_lib` returns the root of the library, but the actual fst/fsti are under `ulib/` in this root. This does not matter for F* since there is an `fstar.include` file, but the Makefile here (and vale-depend) don't follow the fstar.include, so we need to be more explicit.
- Use `fstar.pluginlib` to compile the plugin, this library includes the compiler hooks and etc which are not in the normal application library `fstar.lib`.

NB: an F* installed layout (either binary package, or building from repo, or whatever) looks like:
```
./bin
./bin/fstar.exe
./bin/get_fstar_z3.sh
./lib
./lib/fstar
./lib/fstar/META
./lib/fstar/compiler
./lib/fstar/compiler/fstarcompiler__FStarC_TypeChecker_Generalize.cmx
[...]
./lib/fstar/ulib
./lib/fstar/ulib/ulib/Prims.fst
[...]
```

The first three patches are mostly cleanup and could be merged before F* switches.
- Unifying logic for locating F*, and removing the need for FSTAR_HOME. The Makefile only needs to find a working fstar.exe and it will query it for everything needed (via the --locate family of options). Also the OCAMLPATH is no longer extended globally and `fstar.exe --ocamlenv` is used. (This may need more discussion.)
- Tiny fix in vale-depend.py (fix escape sequence)
- Tiny fix in compilation of libhacl.cmxa (pass -thread)

(This is open for comments of course, and the same comment as https://github.com/FStarLang/karamel/pull/512 for the logic on how to find F* applies. I opened https://github.com/FStarLang/FStar/discussions/3655 for it.)

## Types of changes

- [x] Misc build changes

## Checklist


locate_fstar.mk
```diff
@@ -0,0 +1,27 @@
...
+# If FSTAR_HOME is set, we honor it and pick that F*.
+ifdef FSTAR_HOME
+FSTAR_EXE ?= $(abspath $(FSTAR_HOME)/bin/fstar.exe)
+endif
+
+# Otherwise we try to find it from the PATH.
+
+# Bash's 'type -P' is essentially 'which'. This relies on having bash
+# around, but does not require 'which'.
+FSTAR_EXE ?= $(shell bash -c 'type -P fstar.exe')
```
msprotz, line 15: maybe say why bash is preferable to which?

Makefile.common
```diff
@@ -243,14 +238,9 @@ TAC = $(shell which tac >/dev/null 2>&1 && echo "tac" || echo "tail -r")
...
-  export OCAMLPATH := $(FSTAR_HOME)/lib:$(OCAMLPATH)
-endif
-
 # Warning 8: this pattern-matching is not exhaustive.
 # Warning 20: this argument will not be used by the function.
 # Warning 26: unused variable
-OCAMLOPT = OCAMLFIND_IGNORE_DUPS_IN="`ocamlc -where`/compiler-libs" ocamlfind opt -package fstar.lib -linkpkg -thread -g -I $(HACL_HOME)/obj -w -8-20-26
-OCAMLSHARED = OCAMLFIND_IGNORE_DUPS_IN="`ocamlc -where`/compiler-libs" ocamlfind opt -shared -package fstar.lib -thread -g -I $(HACL_HOME)/obj -w -8-20-26
+export OCAMLFIND_IGNORE_DUPS_IN="`ocamlc -where`/compiler-libs"
+OCAMLOPT = $(FSTAR_EXE) --ocamlenv ocamlfind opt -package fstar.lib -linkpkg -thread -g -I $(HACL_HOME)/obj -w -8-20-26
```
msprotz, line 245: I thought we had had a discussion before on the topic of running  the ocaml tools via F*, and we didn't feel so enthusiastic on the basis that it would be yet one more level of indirection to debug. Maybe @R1kM remembers? Can you say whether there's any intended benefit to doing this?

msprotz (commented):
I have a couple questions, the rest looks good. Thanks!!

Makefile.common
R1kM: I think the relevant discussion was here: https://github.com/hacl-star/hacl-star/pull/998

mtzguido: Right, I understood that discussion as "let's not have F* drive the compilation" (`fstar.exe --ocamlopt`) but I was thinking that `fstar.exe --ocamlenv` is fine, it's just asking F* to extend the environment with a relevant `OCAMLPATH`. I'm fine removing this chunk and just extending OCAMLPATH with `fstar.exe --locate_ocaml` though (instead of $FSTAR_HOME/lib, there's no FSTAR_HOME).

locate_fstar.mk
mtzguido: Actually a good question.. I had `which` before, except that it failed on a Nix build for prosecco-green that did not have `which` in the PATH. Maybe `$(shell which fstar.exe || bash -c 'type -P fstar.exe')` would be preferable? (are you aware of any macos gotchas?)

Makefile.common
msprotz: If you want F* to configure the OCaml environment (which I think is reasonable), a good way might be like other tools do, to have `fstar --env` print something like:

```
export OCAMLPATH=...
export OCAML_IGNORE_DUPS_IN=...
etc.
```

to be `eval`'d in build scripts and in the user's environment. What do you think?

msprotz: that way this is trivially debuggable rather than having to go through the F* source code the minute something gets wrong

msprotz: excellent! if you want to use that, then, I'm fine with it -- this is immediately debuggable and we don't have to rely on F* to fork processes (which introduces additional opportunities for unexpected behavior)

mtzguido:
Updated, also to remove hardcoding the paths of library modules and checked files (https://github.com/hacl-star/hacl-star/pull/1017/commits/407ad83d984282be5d67c3e7330e8a0b407547e7).

Once F* updates to the new build, all that remains to be done is changing `fstar.lib` to `fstar.pluginlib` in `OCAMLSHARED`, I left it out of this PR so it can be merged independently.

mtzguido:
This is now updated with just plainly removing FSTAR_HOME. It worked against current F* master (https://github.com/FStarLang/FStar/commit/e422f42a0a73b0e5c6673985784c09b9ad554a5f ) and the updated krml that does not need FSTAR_HOME (https://github.com/FStarLang/karamel/pull/516, https://github.com/FStarLang/karamel/commit/21ea5a0a0d277630b4b2dddd9b429dd582d53e66).

I think the Nix build will work too after FStarLang/FStar#3662

msprotz (approved):
Also good. If you wanted to make krml work the same way (found in path as opposed to relying on KRML_HOME to simplify makefiles) I would be happy with that too

mtzguido:
Yes absolutely, I plan to do that sometime soon too.

Makefile.common
```diff
@@ -16,36 +16,41 @@ endif
...
     $(error $1 ($(value $(strip $1))) does not exist))
 
-$(call mustexist, FSTAR_HOME)
+# eval $(FSTAR_OCAMLENV) extends the environment so we can
+# find fstar's ocaml libraries.
+FSTAR_OCAMLENV := $(shell $(FSTAR_EXE) --ocamlenv)
+
+# We also set this variable to reduce warnings
+FSTAR_OCAMLENV += OCAMLFIND_IGNORE_DUPS_IN='$(shell ocamlc -where)/compiler-libs';
+FSTAR_OCAMLENV += export OCAMLFIND_IGNORE_DUPS_IN;
```
msprotz, line 45: how about preserving the previous value of OCAMLFIND_IGNORE_DUPS_IN, if any? (I recently started seeing errors about something being both in lib/ and compiler-libs/ and I'm wondering if this might be it)

Makefile.common
```diff
@@ -243,14 +247,8 @@ TAC = $(shell which tac >/dev/null 2>&1 && echo "tac" || echo "tail -r")
...
-else
-  export OCAMLPATH := $(FSTAR_HOME)/lib:$(OCAMLPATH)
-endif
-
 # Warning 8: this pattern-matching is not exhaustive.
 # Warning 20: this argument will not be used by the function.
 # Warning 26: unused variable
-OCAMLOPT = OCAMLFIND_IGNORE_DUPS_IN="`ocamlc -where`/compiler-libs" ocamlfind opt -package fstar.lib -linkpkg -thread -g -I $(HACL_HOME)/obj -w -8-20-26
-OCAMLSHARED = OCAMLFIND_IGNORE_DUPS_IN="`ocamlc -where`/compiler-libs" ocamlfind opt -shared -package fstar.lib -thread -g -I $(HACL_HOME)/obj -w -8-20-26
+OCAMLOPT    = eval "$(FSTAR_OCAMLENV)" && ocamlfind opt -package fstar.lib -linkpkg -thread -g -I $(HACL_HOME)/obj -w -8-20-26
```
msprotz, line 253: also wondering if FSTAR_OCAMLEVEN preserves the values of previous environment variables

msprotz (approved):
Good for me, please merge whenever you feel ready, modulo the two minor clarifications. Thank you!

Makefile.common
mtzguido: It does
```
$ fstar.exe --ocamlenv
OCAMLPATH='/home/guido/r/fstar/master/stage1/out/lib:'; export OCAMLPATH;
$ OCAMLPATH=/foo/bar fstar.exe --ocamlenv
OCAMLPATH='/home/guido/r/fstar/master/stage1/out/lib:/foo/bar'; export OCAMLPATH;
```

msprotz: ah but it does so by reading the path rather than printing `export OCAMLPATH=...;$OCAMLPATH`?

msprotz: I guess that's fine

Makefile.common
mtzguido: Pushed a patch, fwiw it was also overriding the original value before (but the variable was unset anyway, for me). I've also seen those warnings for a long time I think.

Makefile.common
mtzguido: Indeed, I was mimicking what `opam env` does, but I'm not sure why they chose this (maybe to work in a variety of shells I imagine, but `OCAMLPATH='blah':"$OCAMLPATH"` seems robust enough to me?).

It does mean that, in the Makefile `eval $(FSTAR_OCAMLENV)` may override extensions, and they should be set after it. E.g.
```
FSTAR_OCAMLENV = $(shell fstar.exe --ocamlenv)

blah:
        eval "$(FSTAR_OCAMLENV)" && export OCAMLPATH=ggggggggggg:$$OCAMLPATH && env | grep OCAMLPATH
        export OCAMLPATH=ggggggggggg:$$OCAMLPATH && eval "$(FSTAR_OCAMLENV)" && env | grep OCAMLPATH
```
```
$ make blah
eval "OCAMLPATH='/home/guido/r/fstar/master/stage1/out/lib:'; export OCAMLPATH;" && export OCAMLPATH=ggggggggggg:$OCAMLPATH && env | grep OCAMLPATH
OCAMLPATH=ggggggggggg:/home/guido/r/fstar/master/stage1/out/lib:
export OCAMLPATH=ggggggggggg:$OCAMLPATH && eval "OCAMLPATH='/home/guido/r/fstar/master/stage1/out/lib:'; export OCAMLPATH;" && env | grep OCAMLPATH
OCAMLPATH=/home/guido/r/fstar/master/stage1/out/lib:
```
I think it's fine for now..? But something to keep in mind

PR #1025, opened 2025-02-19
Enable more instances of the C Abstract Struct pattern

msprotz:
This is with https://github.com/FStarLang/karamel/pull/542


dist/gcc-compatible/internal/Hacl_Streaming_Types.h
```diff
@@ -30,14 +30,82 @@
...
+
+typedef struct Hacl_Streaming_MD_state_64_s
+{
+  uint64_t *block_state;
+  uint8_t *buf;
+  uint64_t total_len;
+}
+Hacl_Streaming_MD_state_64;
+
+typedef struct Hacl_Streaming_Blake2_Types_two_vec128_s
```
chris-eibl, line 52:
```suggestion
#if defined(HACL_CAN_COMPILE_VEC128)
typedef struct Hacl_Streaming_Blake2_Types_two_vec128_s
```

dist/gcc-compatible/internal/Hacl_Streaming_Types.h
```diff
@@ -30,14 +30,82 @@
...
+{
+  Lib_IntVector_Intrinsics_vec128 *fst;
+  Lib_IntVector_Intrinsics_vec128 *snd;
+}
+Hacl_Streaming_Blake2_Types_two_vec128;
+
```
chris-eibl, line 58:
```suggestion
#endif

```

dist/gcc-compatible/internal/Hacl_Streaming_Types.h
```diff
@@ -30,14 +30,82 @@
...
+typedef struct Hacl_Streaming_Blake2_Types_two_vec256_s
```
chris-eibl, line 59:
```suggestion
#if defined(HACL_CAN_COMPILE_VEC256)
typedef struct Hacl_Streaming_Blake2_Types_two_vec256_s
```

dist/gcc-compatible/internal/Hacl_Streaming_Types.h
```diff
@@ -30,14 +30,82 @@
...
+{
+  Lib_IntVector_Intrinsics_vec256 *fst;
+  Lib_IntVector_Intrinsics_vec256 *snd;
+}
+Hacl_Streaming_Blake2_Types_two_vec256;
+
```
chris-eibl, line 65:
```suggestion
#endif

```

dist/gcc-compatible/internal/Hacl_Streaming_Types.h
```diff
@@ -30,14 +30,82 @@
...
+typedef struct Hacl_Streaming_Blake2_Types_block_state_blake2b_32_s
+{
+  uint8_t fst;
+  uint8_t snd;
+  bool thd;
+  K____uint64_t___uint64_t_ f3;
+}
+Hacl_Streaming_Blake2_Types_block_state_blake2b_32;
+
+typedef struct Hacl_Streaming_Blake2_Types_block_state_blake2b_256_s
```
chris-eibl, line 75:
```suggestion
#if defined(HACL_CAN_COMPILE_VEC256)
typedef struct Hacl_Streaming_Blake2_Types_block_state_blake2b_256_s
```

dist/gcc-compatible/internal/Hacl_Streaming_Types.h
```diff
@@ -30,14 +30,82 @@
...
+{
+  uint8_t fst;
+  uint8_t snd;
+  bool thd;
+  Hacl_Streaming_Blake2_Types_two_vec256 f3;
+}
+Hacl_Streaming_Blake2_Types_block_state_blake2b_256;
+
```
chris-eibl, line 83:
```suggestion
#endif

```

dist/gcc-compatible/internal/Hacl_Streaming_Types.h
```diff
@@ -30,14 +30,82 @@
...
+typedef struct Hacl_Streaming_Blake2_Types_block_state_blake2s_32_s
+{
+  uint8_t fst;
+  uint8_t snd;
+  bool thd;
+  K____uint32_t___uint32_t_ f3;
+}
+Hacl_Streaming_Blake2_Types_block_state_blake2s_32;
+
+typedef struct Hacl_Streaming_Blake2_Types_block_state_blake2s_128_s
```
chris-eibl, line 100:
```suggestion
#if defined(HACL_CAN_COMPILE_VEC128)
typedef struct Hacl_Streaming_Blake2_Types_block_state_blake2s_128_s
```

dist/gcc-compatible/internal/Hacl_Streaming_Types.h
```diff
@@ -30,14 +30,82 @@
...
+{
+  uint8_t fst;
+  uint8_t snd;
+  bool thd;
+  Hacl_Streaming_Blake2_Types_two_vec128 f3;
+}
+Hacl_Streaming_Blake2_Types_block_state_blake2s_128;
+
```
chris-eibl, line 108:
```suggestion
#endif

```

chris-eibl (commented):
Most probably this is generated, but at least I marked the spots to `#ifdef`.

chris-eibl:
I think, `libintvector.h` can now be moved to internal, since the goal was to hide it.
Are there more candidates to be moved?

msprotz:
The windows CI error is a legitimate one -- I need to fix up an abstraction.

msprotz:
Ok at this stage I feel confident that this is in the right shape for Python, modulo my comments above. I intend to merge, and fix things up as they arise during the Python integration process.

Thanks @chris-eibl for the feedback.

The Nix CI will come back red until I merge https://github.com/FStarLang/karamel/pull/542

chris-eibl:
> Regarding the #ifdef: I think this doesn't work, because I still need these types in scope for Hacl_Streaming_HMAC.c to compile, since its agile type contains references to those types in any case (but with the new libintvector-shim.h that should not be a problem).

If my analysis is correct, this means, that
- even though in the Python example `blake2module.c` does no longer need a compiler that supports the intrinsics without compiling for an `AVX` architecture (since the abstraction now hides the "problematic" intrinsics)
- the intrinsics are needed for all `HACL_*_.c` files that include `internal/Hacl_Streaming_Types.h`, e.g. `Hacl_Hash_SHA1.c`. This was not the case, when Python last downstreamed (https://github.com/hacl-star/hacl-star/tree/f218923ef2417d963d7efc7951593ae6aef613f7)
- IOW, a binary that does run on hosts which do not support `AVX`, can no longer be created with such compilers

Hopefully, I am mistaken?

msprotz:
> the intrinsics are needed for all HACL_*_.c files that include internal/Hacl_Streaming_Types.h, e.g. Hacl_Hash_SHA1.c. This was not the case, when Python last downstreamed

ok I did not realize about this side-effect, thanks for pointing it out

in my opinion then, the best option is to split the Hacl_Streaming_Types.h header into three files Hacl_Streaming_Types{,_Simd128,_Simd256} -- and then inclusion is more precise and does not have that unpleasant side effect

what do you think?

thanks again for your detailed review

chris-eibl:
> in my opinion then, the best option is to split the Hacl_Streaming_Types.h header into three files Hacl_Streaming_Types{,_Simd128,_Simd256} -- and then inclusion is more precise and does not have that unpleasant side effect

Sounds even better than my `#ifdef` approach. I just thought that would have been the easiest solution :)

> thanks again for your detailed review

Thanks for all your efforts!

chris-eibl:
Just a thought: include files with the same name only in different paths (e.g. `Hacl_Streaming_Types.h` and `internal/Hacl_Streaming_Types.h`) can be a root of trouble. Since most of the code is generated, would it be easily possible to e.g. postfix the internal ones with `_internal` (or whatever deems more appropriate)?

If this introduces churn or is not easily doable in the code generators - just forget about it :)

msprotz:
@chris-eibl just pushed a change to avoid the issue you described with SHA1 -- the SIMD types are now in their own separate files, only included by the corresponding Blake2 modules.

Can you say more about the issue about the filenames? It's changeable, but I'm curious what issues you ran into.

msprotz:
Let me merge this first then make a note of the filename issue -- this PR has been open for a while and I'm kind of eager to get this going to see if it would solve Python's issues.

msprotz:
I pushed one last simplification to the file structure that had become un-necessary with my latest changes to better support forward struct decls. Now the types that contains references to libintvector live in their respective Hacl_Hash_Blake2{b256,s128} files and are properly hidden behind the public header.

If you don't mind taking one last look and confirming that this is good for Python, I'll go ahead and merge.

chris-eibl:
I've just had a quick look - sorry, it's late here and I am busy the next days.

AFAICT, when doing some find-in-files, etc, and quickly rushing over the last commits - this looks good to me.

Sorry that I don't have more time right now. Don't feel blocked - but if you're in doubt about merging, maybe you can just start off a PR in Python based on your last commit and during working on it switch over?

msprotz:
I caught some incorrect cross-algorithm dependencies with e.g. Hacl_MAC_Poly1305 including Hacl_Blake2b_32.h, which I eventually fixed with https://github.com/FStarLang/karamel/pull/550

With that, everything is now green and I feel confident about merging. Thanks everyone

chris-eibl:
I did a draft PR based on https://github.com/hacl-star/hacl-star/commit/809c320227eecc61a744953f1ee574b4f24aabe3 and can confirm that for clang-cl on Windows the issue is fixed 👍 

Thank you so much for your efforts. @msprotz Maybe you can do a review there?

msprotz:
> for clang-cl on Windows the issue is fixed 👍

terrific news!

of course very happy to do a review there, cheers

Makefile
```diff
@@ -738,32 +738,40 @@ TARGET_H_INCLUDE = -add-early-include '"krml/internal/target.h"'
...
+  -add-include 'Vale.h:<inttypes.h>' \
+  -add-include 'EverCrypt_Hash.h:"libintvector.h"' \
+  -add-include 'Hacl_Streaming_HMAC.h:"libintvector-shim.h"'
 
 # Disabled for distributions that don't include code based on intrinsics.
 INTRINSIC_INT_FLAGS = \
   -add-include 'Hacl_P256:"lib_intrinsics.h"' \
   -add-include 'Hacl_Bignum:"lib_intrinsics.h"' \
-  -add-include 'Hacl_Bignum_Base:"lib_intrinsics.h"' \
+  -add-include 'Hacl_Bignum_Base.h:"lib_intrinsics.h"' \
```
mtzguido, line 774: This bit fixes the build (currently broken on main) after https://github.com/FStarLang/karamel/pull/542. I can't really comment on the rest, the high-level idea makes sense to me but I'm not too familiar with the structure of the build here.

mtzguido (approved):

R1kM (approved):

PR #1034, opened 2025-04-28
A few Makefile improvements

mtzguido:
## Proposed changes

Removing some of the substitutions needed for older F* versions. Now F* can
- Take a checked file as input for extracting it (so no need to find the original fst)
- Does not need `--cache_checked_modules` in order to *read* checked files, we always do it. We pass `-c` (an alias for the former) when we intend to *write* a checked file.
- When extracting an OCaml file, we default to only extracting the modules passed in the cmdline, so some `--extract_module` are gone.

## Types of changes

## Checklist


msprotz:
- Is the empty prims.krml no longer generated?
- Can we use the long option name instead of `-c`? `-c` is not super descriptive
thanks!!

mtzguido:
> * Is the empty prims.krml no longer generated?

It's not, since https://github.com/FStarLang/FStar/commit/3b670410a0032b74e4f9b899fcfb63039e92c98c F* never puts it in `ALL_KRML_FILES` (and others). So I also added the last patch, thanks for noticing this!

The previous change was just removing the substitution `prims` -> `Prims` since Prims is now uppercase as every other module.

> * Can we use the long option name instead of `-c`? `-c` is not super descriptive

Done

PR #1042, opened 2025-05-09
Lib_Memzero0.c: optimize APPLE_HAS_MEMSET_S macro

aeiouaeiouaeiouaeiouaeiouaeiou:
Let's use the same method that was added to https://github.com/FStarLang/karamel/pull/505.
It would be nice if someone tested this on non-desktop platforms.

## Types of changes

## Checklist

### DIFF
--- a/dist/gcc-compatible/Lib_Memzero0.c
+++ b/dist/gcc-compatible/Lib_Memzero0.c
@@ -11,18 +11,18 @@
 #if defined(__APPLE__) && defined(__MACH__)
 #include <AvailabilityMacros.h>
 // memset_s is available from macOS 10.9, iOS 7, watchOS 2, and on all tvOS and visionOS versions.
-#  if (defined(MAC_OS_X_VERSION_MIN_REQUIRED) && (MAC_OS_X_VERSION_MIN_REQUIRED >= __MAC_10_9))
-#    define APPLE_HAS_MEMSET_S 1
-#  elif (defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_7_0))
-#    define APPLE_HAS_MEMSET_S 1
+#  if (defined(MAC_OS_X_VERSION_MIN_REQUIRED) && defined(MAC_OS_X_VERSION_10_9) && (MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_9))
+#    define APPLE_HAS_MEMSET_S
+#  elif (defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && defined(__IPHONE_7_0) && (__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_7_0))
+#    define APPLE_HAS_MEMSET_S
 #  elif (defined(TARGET_OS_TV) && TARGET_OS_TV)
-#    define APPLE_HAS_MEMSET_S 1
-#  elif (defined(__WATCH_OS_VERSION_MIN_REQUIRED) && (__WATCH_OS_VERSION_MIN_REQUIRED >= __WATCHOS_2_0))
-#    define APPLE_HAS_MEMSET_S 1
+#    define APPLE_HAS_MEMSET_S
+#  elif (defined(__WATCH_OS_VERSION_MIN_REQUIRED) && defined(__WATCHOS_2_0) && (__WATCH_OS_VERSION_MIN_REQUIRED >= __WATCHOS_2_0))
+#    define APPLE_HAS_MEMSET_S
 #  elif (defined(TARGET_OS_VISION) && TARGET_OS_VISION)
-#    define APPLE_HAS_MEMSET_S 1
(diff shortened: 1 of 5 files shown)


msprotz:
@freakboy3742 can you take a quick look at this if you have time? thank you!

msprotz:
I'm confused -- I thought desktop ios was tested as part of the python regressions (we also run a build with macos here). Do you know what we're missing in terms of coverage?

aeiouaeiouaeiouaeiouaeiouaeiou:
@msprotz I know the latest version of Python doesn't support macOS older than 10.13 the MacPorts project tries to build packages for older systems as much as possible.
HACL* already includes code to run on unsupported systems, so I don't see anything bad with this fix.

freakboy3742:
Mechanically, those changes look fine. 

It's a matter of style preference whether `__MAC_10_9` is better than `1090`; personally, I prefer the constants, if only because they're more explicit (and less confusing when you get to macOS 10.10+, where the value becomes `101000`, but the constant is `__MAC_10_10`. The only advantage I can see to the constants is that on a pre-10.9 machine, the constant won't be defined... but then, the >= check should also fail, so the net result is the same.

The same goes for an explicit "define as false" vs "undefined" for the "not supported" case - both work; it's a matter of taste/preference/style.

As for missing coverage - I can't think of anywhere that tvOS, watchOS and visionOS coverage would be systematically tested at present. I have tested visionOS support as part of the testing BeeWare's Apple support packages, but not regularly as part of CI.

msprotz:
@freakboy3742 makes a good point -- on older versions, the defines do not exist, and such, the chain of ifdef-elif will simply end up determining that there is no memset_s support.

@aeiouaeiouaeiouaeiouaeiouaeiou can you clarify whether this is simply a cosmetic change, or if it fixes a build issue, and if so, what is the exact toolchain that used to build with this but now no longer builds?

msprotz:
Furthermore, there is a fallback codepath, so it's not like platforms that don't support memset_s won't build, they'll just get a preprocessor warning.

jmroot:
> @freakboy3742 makes a good point -- on older versions, the defines do not exist, and such, the chain of ifdef-elif will simply end up determining that there is no memset_s support.

Undefined macros evaluate to 0. `MAC_OS_X_VERSION_MIN_REQUIRED` will always be >= 0.

> can you clarify whether this is simply a cosmetic change, or if it fixes a build issue, and if so, what is the exact toolchain that used to build with this but now no longer builds?

Real world build failure: https://build.macports.org/builders/ports-10.8_x86_64-builder/builds/207154/steps/install-port/logs/stdio
It was fixed by this patch: https://github.com/macports/macports-ports/blob/master/lang/python314-devel/files/Lib_Memzero0.c.patch

Replacing `__MAC_10_9` with `1090` is essential for the check to work correctly with older macOS SDKs. IMO, the rest of this PR is a matter of style. (That includes replacing `__IPHONE_7_0` and `__WATCHOS_2_0` with their numeric equivalents, because not using the latest SDK to build for iOS and watchOS is virtually unheard of due to App Store rules.)

freakboy3742:
> Would my proposed solution (above) with the two `defined(...)` work? I'm loath to add magic numbers and I would much rather rely on named macros, mostly because of this comment:

I believe so, yes.

jmroot:
I don't have a Mac OS X 10.9 SDK handy to check, but if it defines `__MAC_10_9`, then yes checking if that is defined will also work.

jmroot:
It looks like `AvailabilityMacros.h` does not necessarily define `__MAC_10_9`, but does define `MAC_OS_X_VERSION_10_9 ` for this purpose: https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.9.sdk/usr/include/AvailabilityMacros.h#L101

msprotz:
What does this mean for this PR? That on 10.9 the code would use memset instead of memset_s?

jmroot:
> What does this mean for this PR? That on 10.9 the code would use memset instead of memset_s?

If `__MAC_10_9` continues to be the basis for the check, then possibly. But I don't think there's any downside to changing the code to:
```
#  if (defined(MAC_OS_X_VERSION_MIN_REQUIRED) && defined(MAC_OS_X_VERSION_10_9) && (MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_9))
#    define APPLE_HAS_MEMSET_S
…
```

msprotz:
Makes sense. Should other macros be also changed to be of the form `MAC_OS_X_VERSION_10_*`? The `__` in front of the existing macro makes me feel like this is an "internal" macro.

msprotz:
Great -- it looks like we've converged. @aeiouaeiouaeiouaeiouaeiouaeiou can I let you revise the PR to include the changes discussed?
1. two `defined(...)` tests and
2. use macros of the form `MAC_OS_X_VERSION_10_*` everywhere.

Thanks to all for your input!

aeiouaeiouaeiouaeiouaeiouaeiou:
@msprotz okay, I'll push the changes today.

msprotz (approved):
Looks good to me. Thanks everyone for your help!

PR #1050, opened 2025-07-01
Stabilize Hacl.Spec.Bignum.Lib by disabling nlarith

R1kM:
As described in the title. This file exhibited highly unstable behavior across Z3 versions, requiring more than 30 minutes with Z3 4.13.3.
Disabling non-linear arithmetic and fleshing out the proofs seems to help with proof stability.


msprotz:
Awesomesauce.

@mtzguido do you know where that leaves us for getting rid of z3 4.8.5 for HACL*?

Issue #64, opened 2017-10-08
Primitives and constructions should have test vectors in F*


beurdouche:
The test files handwritten in C should be re-extracted from F* (if possible)

beurdouche:
This is done for all new code.

PR #1052, opened 2025-07-14
Poly1305.Bitvectors: remove a flaky, unused lemma

mtzguido:
Bitvectors are primitive in F*, and this proof can only work via Z3 figuring it out. There is no point in a big proof, unless we expand FStar.BV to expose more facts about bitvectors. That said, Z3 4.8.5 could not figure it out, so this proof was spelled out (though I'm not sure how it did so, except by randomizing the solver--- there is AFAICT no lemma about `bv2list` that could help).

Howver, Z3 4.13.3 works pretty reliably. This snippet made the lemma pass reliably in this file.

	#push-options "--split_queries always"
	let int2bv_uext_64_128 (x : nat) :
	  Lemma  (requires FStar.UInt.size x 64)
		 (ensures  bv_uext #64 #64 (int2bv #64 x) == int2bv #128 x)
	=
	  assert (bv_uext #64 #64 (int2bv #64 x) == int2bv #128 x);
	  ()
	#pop-options

And also works in plain F* (I am adding a test for it FStarLang/FStar#3905).

(Do note however that the assert seems to help... but it does not make a difference in a clean F* module. I am investigating why separately.)

However, this lemma is unused and does not have an SMT pattern, so this patch just removes it. We could expose it from the F* library if needed, perhaps with the sizes being arguments.

## Proposed changes

## Types of changes

## Checklist

### DIFF
--- a/vale/code/crypto/poly1305/x64/Vale.Poly1305.Bitvectors.fst
+++ b/vale/code/crypto/poly1305/x64/Vale.Poly1305.Bitvectors.fst
@@ -109,23 +109,6 @@ let lemma_bv128_64_64_and x x0 x1 y y0 y1 z z0 z1 =
           rewrite_eqs_from_context ();
           norm [delta])
 
-#push-options "--smtencoding.elim_box true --z3refresh --z3rlimit 20 --max_ifuel 2 --max_fuel 2"
-let int2bv_uext_64_128 (x1 : nat) :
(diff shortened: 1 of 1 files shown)


msprotz (approved):
Fine by me, thanks! Again, let's wait for CI to be back so that we can make sure everything still verifies

PR #1053, opened 2025-07-15
All: simplify fuel settings

mtzguido:
Generated by these sed calls:

    sed 's/--initial_fuel \([0-9]*\) --max_fuel \1\>/--fuel \1/g'
    sed 's/--initial_ifuel \([0-9]*\) --max_ifuel \1\>/--ifuel \1/g'
    sed 's/--max_fuel 0\>/--fuel 0/g'
    sed 's/--max_ifuel 0\>/--ifuel 0/g'

I.e. use the `--fuel`/`--ifuel` options to set the initial and max fuels at once. Also, setting the max to zero is equivalent to using the shorter option.

No functional change. See similar F* PR: https://github.com/FStarLang/FStar/pull/3906.

## Proposed changes

## Types of changes

## Checklist


mtzguido:
Oh I forgot you had a Nix based CI. Then would this just mostly amount to removing then `runs-on: self-hosted` from nix.yml? Plus setting up the magic-nix-cache presumably. Nix is supposed to abstract away the rest of the system, so I don't think we need to do anything to the cloud runner VM.

What I had in mind was a simple, non-Nix, `make && make test` CI, as I'm not familiar with what else this CI is doing (eg updating hints and whatnot). But Nix would be better.

mtzguido:
It worked! It took 2h26m, but the first run is slower due to an empty cache (so it's building z3, F*, etc). I can make a PR with that patch if folks agree.

Ideally, the choice of self-hosted vs cloud could be flipped somewhere in the repo config, but AFAIK it requires an edit to the workflow file. So once the INRIA runner is back, we would revert the PR.

msprotz:
the run came back green, do you want to merge this temporarily on this branch to allow CI to run for this PR and your next few ones?

msprotz:
How do cloud runners reuse the nix cache? Is there a shared nix artifact repository somewhere shared across all runners?

mtzguido:
> the run came back green, do you want to merge this temporarily on this branch to allow CI to run for this PR and your next few ones?

Sure thing, updated.

> How do cloud runners reuse the nix cache? Is there a shared nix artifact repository somewhere shared across all runners?

Yes. There is a Github actions cache per repo, this is the one for my fork: https://github.com/mtzguido/hacl-star/actions/caches. I think all runners, cloud or not, have access to it. The magic-nix-cache action somehow (magically) hooks it up with Nix. Usually one has to use cache keys to not mix up different OSes, architectures, etc, but I think this is all handled by the magic action.

The self-hosted runner was not using this magic-cache, however. Maybe it can be enabled and then the INRIA machine would be stateless as far as actions go. But we should make sure it works.

msprotz:
Oh yes absolutely. I expected you would merge whatever comes back green if I've approved it.

PR #1059, opened 2025-07-18
Some more tidying up

mtzguido:
## Proposed changes

Adding interfaces, reducing some includes in Vale fstis, and using solver=2 in Vale.Poly1305.X64 which is the file currently failing in Everest. This does not make a big difference locally... nor using 4.8.5 apparently. Seems like this proof is just very very flaky.

## Types of changes

## Checklist

### DIFF
--- a/code/poly1305/Hacl.Impl.Poly1305.Lemmas.fst
+++ b/code/poly1305/Hacl.Impl.Poly1305.Lemmas.fst
@@ -9,11 +9,6 @@ open Lib.ByteSequence
 
 #set-options "--z3rlimit 50 --max_fuel 1"
 
-val uint_from_bytes_le_lemma: b:lseq uint8 16 -> Lemma
- (let lo = uint_from_bytes_le #U64 (sub b 0 8) in
-  let hi = uint_from_bytes_le #U64 (sub b 8 8) in
-  nat_from_bytes_le b == pow2 64 * uint_v hi + uint_v lo)
-
 let uint_from_bytes_le_lemma b =
   let r1 = nat_from_bytes_le b in
   let r2 = uints_from_bytes_le #U64 #SEC #2 b in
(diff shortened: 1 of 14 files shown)


msprotz (approved):

mtzguido:
I think I found the reason for my confusion. This was failing badly in the IDE for me, but not on batch. The Makefile has this:
```
$(addsuffix .checked,$(VALE_FSTS)): \
  FSTAR_FLAGS=$(VALE_FSTAR_FLAGS)
```
so it will pass all these options when checking the generated `Vale.Poly1305.X64.fst`:
```
VALE_FSTAR_FLAGS=--z3cliopt smt.arith.nl=false \
  --z3cliopt smt.QI.EAGER_THRESHOLD=100 --z3cliopt smt.CASE_SPLIT=3 \
  --max_fuel 1 --max_ifuel 1 --initial_ifuel 0 \
  --smtencoding.elim_box true --smtencoding.l_arith_repr native \
  --smtencoding.nl_arith_repr wrapped
```
But this file is in `obj/`, and the VS Code extension will use the top-level json config, which does not pass any of them. I did add them in `vale/Hacl.Vale.fst.config.json`, but then it only applies under `vale/`.

Emacs should be fine due to the fst-in targets.

F* really needs a single way of specifying flags.

Meanwhile, what about removing `VALE_FSTAR_FLAGS` and just adding a literal include with a proper set-options in the .vaf files?

msprotz:
That seems much better. I don't like encoding such semantic stuff in the Makefile where it typically gets overlooked.

FYI, I think this stems from the reconciliation of the separate build systems (vale used to be on scons for a while), and the vale project was setting these parameters wholesale in their scons file

Are these VALE_FLAGS only for .fst files that stem from .vaf files, or does it also cover Vale.* including the hand-written stuff? (I don't remember how VALE_FSTS is defined)

mtzguido:
I'll send a separate PR with the .vaf include. I'm gonna merge since I think the everest build may pass. I ran a 10x quake test of this file and it passed.

> Are these VALE_FLAGS only for .fst files that stem from .vaf files, or does it also cover Vale.* including the hand-written stuff? (I don't remember how VALE_FSTS is defined)

They are also used by other Vale files.
```
$(call only-for,$(HACL_HOME)/vale/%.checked): \
  FSTAR_FLAGS=$(VALE_FSTAR_FLAGS)

$(call only-for,$(HACL_HOME)/vale/specs/%.checked): \
  FSTAR_FLAGS=$(VALE_FSTAR_FLAGS)
$(call only-for,$(HACL_HOME)/vale/code/%.checked): \
  FSTAR_FLAGS=$(VALE_FSTAR_FLAGS)

$(addsuffix .checked,$(VALE_FSTS)): \
  FSTAR_FLAGS=$(VALE_FSTAR_FLAGS)
```
So the vaf include wouldn't solve that. I'll try to unify all of them.

msprotz:
Sounds exciting. I think this solves the problem of consistency between IDE and batch mode -- both invocations of F* would discover this file and parse options from it.

How about user discoverability? If someone is trying to debug a file, now they have to remember whether a fstar.config files exists at any point between the current directory and the root of the project... what do you think? Is that ok?

PR #1072, opened 2026-04-09
Cherry-pick Cygwin fix from python/cpython#148202

protz:


protz:
@mtzguido do you know what we need to fix here?

make: krml: No such file or directory
Makefile.basic:21: *** "Could not find a proper include directory for karamel. Either set KRML_INCLUDEDIR to it, or add krml to the PATH, or set KRML_EXE.".  Stop.

mtzguido:
Sorry about that, see #1073

R1kM (approved):

